PR #2, opened 2018-03-13
[TMP-39] Init node launcher structure

flyingleafe:
- [x] Define basic launcher structure (modules, brackets, etc.)
- [x] Basic CLI params parsing
- [x] Initiate logging
- [ ] Workers initiation

### DIFF
--- a/src/Disciplina/Launcher/Resource.hs
+++ b/src/Disciplina/Launcher/Resource.hs
@@ -1,0 +1,63 @@
+
+-- | Common resources used by Disciplina nodes
+
+module Disciplina.Launcher.Resource
+       ( BasicNodeResources (..)
+       , bracketBasicNodeResources
+       ) where
+
+import Universum
+
+import Mockable (Production)
+import System.Wlog (LoggerConfig (..), LoggerName, WithLogger, maybeLogsDirB, parseLoggerConfig,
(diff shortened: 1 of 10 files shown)


src/Disciplina/Launcher/Resource.hs
```diff
@@ -0,0 +1,63 @@
...
+-- to start working.
+data BasicNodeResources = BasicNodeResources
+    { bnrLoggerName :: !LoggerName
+    }
+
+----------------------------------------------------------------------------
+-- Logging
+----------------------------------------------------------------------------
+
+readLoggerConfig :: MonadIO m => Maybe FilePath -> m LoggerConfig
```
lambda76: Why use mtl-style MonadIO instead of concrete Production type with these logging functions, readLoggerConfig, getRealLoggerConfig and setupLoggers?

flyingleafe: Just a habit, I guess. Also, it might be so that in the future logging initialization isn't done in `Production` monad (e. g. in Cardano it's done directly in IO monad for reasons of some advanced initial logging, which we don't want yet)

lambda76: Ok, makes sense.

PR #7, opened 2018-04-04
[TMP-42] Basic interface for database

flyingleafe:
Simple database abstraction and RocksDB interface. PR is done only because of the work went there. RocksDB interface should be rewritten using Lootbox after DB layer in Lootbox is merged: https://github.com/serokell/lootbox/pull/2


src/Disciplina/DB/Class.hs
```diff
@@ -0,0 +1,12 @@
...
+import Universum
+
+-- | Read-only interface to key-value DB storing serialized data.
+class Monad m => MonadDBRead m where
+    dbGet :: ByteString -> m (Maybe ByteString)
+
+-- | Full interface to key-value DB.
+class MonadDBRead m => MonadDB m where
+    dbPut :: ByteString -> ByteString -> m ()
+    dbDelete :: ByteString -> m ()
```
Heimdell: What will be stored there?

I have `auth-data-structures:Data.Tree.AVL.KVStoreMonad..KVStoreMonad` class.

How will it interact with your class?

src/Disciplina/DB.hs
```diff
@@ -0,0 +1 @@
+{-# OPTIONS_GHC -F -pgmF autoexporter #-}
```
Heimdell: I'm against things like that, becase I think, reexporting module will define intentions more clear.

Heimdell (approved):
Aside from pragma magic, looks fine.

src/Disciplina/DB.hs
flyingleafe: Point taken, though we extensively used that in Cardano. I personally don't think that this is bad in cases where module subtree is simple and there's no need to not import some submodules in parent module - it saves time and simplifies adding new submodules. We should discuss it with the rest of the team and come up with some kind of best practice, written down somewhere.

src/Disciplina/DB/Class.hs
flyingleafe: >How will it interact with your class?
Obviously:
```
retrieve = fmap decodeStrict . dbGet . encodeStrict
store k = dbPut k . encodeStrict
```

PR #9, opened 2018-04-09
[DSCP-59] Define private blockchain types using Snowdrop

flyingleafe:
- [x] Add Snowdrop as a dependency and migrate to LTS-10.8
- [x] Define private block datatype
- [x] Define private transaction datatype
  - [x] Define datatypes and interfaces for hashing and signing

### DIFF
--- a/shell.nix
+++ b/shell.nix
@@ -1,0 +1,32 @@
+with import <nixpkgs> {
+  config = {
+    packageOverrides = pkgs: {
+      jemalloc = pkgs.jemalloc.overrideAttrs (oldAttrs: rec {
+        name = "jemalloc-${version}";
+        version = "4.5.0";
+        src = pkgs.fetchurl {
+          url = "https://github.com/jemalloc/jemalloc/releases/download/${version}/${name}.tar.bz2";
+          sha256 = "9409d85664b4f135b77518b0b118c549009dc10f6cba14557d170476611f6780";
+        };
+      });
+    };
+  };
+};
+
+let
+  hsPkgs = haskell.packages.ghc822;
+in
+  haskell.lib.buildStackProject {
+    name = "disciplina";
+    ghc = hsPkgs.ghc;
+    buildInputs = [
+      openssl
+      zlib
+      binutils
+      git
+      rocksdb
+      gmp
+      hsPkgs.cpphs
+      hsPkgs.happy
+    ];
+  }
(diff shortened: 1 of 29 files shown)


Heimdell:
Wow. I doesn't see Adga in my package indices on Ubuntu. I'll set it up manually, but we need to find what dependency references it.

flyingleafe:
@Heimdell ha-ha, welcome.
For me, installing Agda system-wide solved the problem.

Heimdell:
For me, it didn't, since terminals in linux honor `--pure` option. I had to kill `~/.stack/indices/Hackage` and now it is installing.

Heimdell:
It builds on archlinux (after injecting jemalloc-4.5 instead of 5.0 into `nix`).

Heimdell:
In builds on ubuntu with the same jemalloc injection.

Heimdell (approved):
It builds, tests are succeeding.

yorickvP:
We're working on supporting different ghc versions and supporting ssh dependencies. `cardano-sl-networking` fails to build with the following log: https://gist.github.com/yorickvP/da06129f5ec3083fe9527af76e342d18

flyingleafe:
I managed to pin exact `jemalloc` version (4.5.0) in `shell.nix`. Unfortunately, it means that at first build `jemalloc-4.5.0` itself should be built, which takes plenty of time.

flyingleafe:
During thinking about transaction types, I understood that these types should necessarily include hashes and signatures, and couldn't help but implement hashes and signatures in a way which can be later abstracted to Lootbox.

flyingleafe:
Seems enough. @Heimdell @lambda76 please take a look.

shell.nix
```diff
@@ -0,0 +1,32 @@
+with import <nixpkgs> {
+  config = {
+    packageOverrides = pkgs: {
+      jemalloc = pkgs.jemalloc.overrideAttrs (oldAttrs: rec {
```
lambda76: Might be better to pin exact version of nixpkgs instead of pinning per package.

lambda76 (approved):

shell.nix
flyingleafe: Unfortunately, in the older `nixpkgs` release which contained `jemalloc-4.5.0` there was no `ghc-8.2.2`, used in `lts-10.8`, which caused numerous other problems.

PR #10, opened 2018-04-18
[DSCP-72] Datatypes and functions for working with Activity Type Graph

flyingleafe:
- [x] Datatype for ATG
- [x] Final datatype for Subject ID
- [x] ATG path queries and traversals


flyingleafe:
I didn't find any reason why would we need anything more complicated than `Word32` for SubjectId (probably only a newtype without `Num` instances over `Word32` might be better).

Heimdell (approved):

lambda76 (approved):
Looks good to me.

PR #11, opened 2018-04-18
Attempt to fix Nix CI

flyingleafe:
Let's see what happens with renewed `default.nix`!

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -2,7 +2,6 @@ let
   nixpkgs = import "${overlay}/nixpkgs.nix";
   overlay = builtins.fetchGit {
     url = "ssh://git@github.com:/serokell/serokell-ops.git";
-    rev = "e6e22f100da2834685246127e064dfa5a321c9df";
   };
 in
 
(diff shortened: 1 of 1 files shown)


yegortimoshenko:
See: https://gist.github.com/serokell-bot/9e2504c0cd72d8b39f0fb97a680e8e3f#file-nix-ci-log-L336

flyingleafe:
@yegortimoshenko 
To avoid this, in `stack.yaml` we have the following:
```
ghc-options:
  # Needed to buid
  cardano-sl-networking: -fno-warn-redundant-constraints
```
I might be wrong, but seems like the problem is that `stack2nix` cannot into this config option. Is there a way fix it? Might be a valuable contribution to `stack2nix`, if that's the case.

flyingleafe:
@yegortimoshenko

I researched the internals of `stack2nix` a little, and then `cabal2nix`, which is used by the former, and then general structure of Haskell derivations in Nixpkgs, and it seems like there simply is no way of specifying additional compiler arguments (other than specified in package’s `.cabal` file) in Nix derivation for the package. Therefore, in order to support `ghc-options` from `stack.yaml` we need to patch virtually the whole Nix-Haskell ecosystem. Seems like too much for fixing CI for one project =\

I’ll be happy to be proven wrong, though.

yorickvP:
```nix
{
    cardano-sl-networking = appendConfigureFlag previous.cardano-sl-networking "--ghc-option=-fno-warn-redundant-constraints";
}
```

flyingleafe:
wow, thanks, @yorickvP , but why CI status isn't showing in PRs anymore? :thinking:

PR #12, opened 2018-04-19
[DSCP-66] [DSCP-48] Datatype for queries DSL

lambda76:
- [x] Datatype for DSL queries
- [x] Stub for DSL interpreter

### DIFF
--- a/src/Disciplina/DB/DSL/Interpret.hs
+++ b/src/Disciplina/DB/DSL/Interpret.hs
@@ -1,10 +1,10 @@
 module Disciplina.DB.DSL.Interpret
               ( runQuery
+              , RunQuery (..)
               )
               where
 
 import Universum
-import Disciplina.DB.Class (MonadDBRead)
 import Disciplina.DB.DSL.Types (QueryTx(..), QueryTxs(..), WHERE (..)
                                ,TxIdEq (..), TxGrade(..), QueryObj(..)
                                ,TxsFilterExpr(..), ObjHashEq(..))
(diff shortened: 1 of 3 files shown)


src/Disciplina/DB/DSL/Types.hs
```diff
@@ -0,0 +1,73 @@
...
+import Universum
+import qualified Disciplina.Core as Core (Grade(..))
+
+
+data Entity = Tx
+            | Txs
+            | Obj
+
+data FilterEntity = TxHash
+                  | ObjHash
```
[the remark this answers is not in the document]
lambda76: I agree with this. I'm not sure however how using type family will affect implementation of interpreter.

src/Disciplina/DB/DSL/Types.hs
```diff
@@ -0,0 +1,100 @@
...
+
+module Disciplina.DB.DSL.Types
+       ( Query (..)
+        ,WHERE(..)
+        ,Tx(..)
+        ,Txs(..)
+        ,Obj(..)
+        ,FilterEntity(..)
+        ,FilterExpr(..)
+        ,NodeWithProp(..)
```
Heimdell: Can we have spaces after `,`?

lambda76: yes sure, I will add space. @Heimdell only purpose of `WHERE` is to make queries easier to read, as in `SELECE Tx WHERE ...`. Do you think we should remove `WHERE`?

src/Disciplina/DB/DSL/Types.hs
```diff
@@ -0,0 +1,100 @@
...
+data NodeWithProp = SubjectNode Core.SubjectId
+                  | TxSubjectNode
+
+data WHERE = WHERE
+
+
+data FilterExpr a b where
+   (:==) :: FilterLHS a b -> FilterRHS a b -> FilterExpr a b
+   (:>=) :: FilterLHS a b -> FilterRHS a b -> FilterExpr a b
+   (:&)  :: FilterExpr a b -> FilterExpr c d -> FilterExpr e f
```
flyingleafe: If `a` and `c` are different, what's the meaning of the expression? Can you meaningfully combine two expressions, one of which selects `Txs`, and another selects `Obj`? And what meaning we can possibly attribute to `f` then?
With this definition of `FilterExpr` and `Query`, which type `q4 txParams subjId` has? (It has type `Query Txs f`, where `f` is ambiguous and there's no defined type which we can put in place of `f`).

I think that slightly different approach should be introduced here - `FilterExpr` (and, consequently, `Query`) should have only one type parameter, like this:
```
data FilterExpr a where
    (:==) :: forall b. FilterLHS a b -> FilterRHS a b -> FilterExpr a
    (:>=) :: forall b. FilterLHS a b -> FilterRHS a b -> FilterExpr a
    (:&)  :: FilterExpr a -> FilterExpr a -> FilterExpr a
    ...
```

lambda76: Thanks, I will try this approach and see if it compiles. The problem I had when I first tried this (without the `forall b` was that the parameters to `(:&)` both had to be of the same type `FilterExpr a`, but they were not.

src/Disciplina/DB/DSL/Types.hs
```diff
@@ -0,0 +1,54 @@
...
+data NodeWithProp = SubjectNode Core.SubjectId
+                  | TxSubjectNode
+
+-- | TODO, what is the correct type for TxSubjectIdEq?
+data TxsFilterExpr = TxSubjectIdEq Int
+                   | TxGradeEq Core.Grade
+                   | (:>=) TxGrade Core.Grade
+                   | (:&) TxsFilterExpr TxsFilterExpr
+                   | (:||) TxsFilterExpr TxsFilterExpr
+                   | (:~>) NodeWithProp NodeWithProp
```
flyingleafe: It actually makes more sense to have something like `TxSubjectIsDescendantOf Core.SubjectId` instead here. With this constructor it's possible to have expression like `SubjectNode 3 :~> SubjectNode 5`, which doesn't represent any predicate on a transaction.

src/Disciplina/DB/DSL/Interpret.hs
```diff
@@ -0,0 +1,99 @@
...
+  runTxsQuery :: QueryTxs -> m [PrivateTx]
+  runObjQuery :: QueryObj -> m (Maybe Int)
+
+instance MonadSearchTxObj IO where
+  runTxQuery = runIOTxQuery
+  runTxsQuery = runIOTxsQuery
+  runObjQuery = runIOObjQuery
+
+-- | Query type determine query return type
+class RunQuery a b | a -> b where
```
lambda76: Thinking about it, wouldn't it be better to have `class  (Monad m) => RunQuery m a b | a -> b` and skip `MonadSearchTxObj` type class?

flyingleafe: No, I don't think so. This way you'll have to declare three separate `RunQuery` instances for each separate monad, instead of three `RunQuery` instances in total.

src/Disciplina/DB/DSL/Interpret.hs
```diff
@@ -0,0 +1,99 @@
+module Disciplina.DB.DSL.Interpret
+              ( runQuery
+              , RunQuery (..)
```
flyingleafe: You should import `MonadSearchTxObj` class here. Could be easily added later on demand, though.

PR #15, opened 2018-04-29
[DSCP-76] Use `tasty` + `hspec` for test specifications.

flyingleafe:
Reasons for that are in the issue description.


volhovm (approved):

volhovm:
I'm not sure the CI works, so I've built in on my machine just in case.

PR #19, opened 2018-05-22
[WIP] [DSCP-92] Implement sized merkel tree

lambda76:
- [x] Datatype SizedMerkleTree a  is defined
- [x] There is a way to efficiently build SizedMerkleTree a  from [a] , if a  satisfies HasHash a . 
- [x] Datatype SizedMerkleProof a  is defined, which represents a subtree containing paths from one or several leaves to the root.
- [x] Given an index of an element (or range/set of indices) and a SizedMerkleTree a , one can get SizedMerkleProof a  corresponding to the index (indices)
- [x] One can check that given SizedMerkleProof  is valid (both in terms of hashes and size)
- [x] One can retrieve a SizedMerkleRoot a  from SizedMerkleTree a , which is basically a pair (root_hash, size) .
- [x] One can validate a SizedMerkleProof  against SizedMerkleRoot 
- [x] For all these actions there are tests.


src/Disciplina/Educator/SizedMerkleTree.hs
```diff
@@ -0,0 +1,237 @@
...
+       , mkBranch
+       , mkLeaf
+       ) where
+
+import Universum hiding (foldMap, toList)
+
+import Disciplina.Crypto (Hash, HasHash, hash, unsafeHash)
+import Disciplina.Crypto.Hash.Class (HashFunc, AbstractHash (..))
+
+import Codec.Serialise (Serialise(..), serialise)
```
Heimdell: Why do we have concrete dependencies on hash algorithms and serialisation?

It would be a lot simpler to extract this into separate package, if you abstracted them as interfaces.

src/Disciplina/Educator/SizedMerkleTree.hs
```diff
@@ -0,0 +1,237 @@
...
+import Data.Foldable (Foldable (..))
+import Prelude (show)
+
+import qualified Data.ByteString.Builder.Extra as Builder
+import qualified Data.ByteString.Lazy as LBS
+
+-- | Data type for root of sized merkle tree.
+data SizedMerkleRoot a = SizedMerkleRoot
+    { smrHash :: Hash LBS.ByteString  -- ^ returns root 'Hash' of Merkle Tree
+    , smrSize :: Word8 -- ^ size of root node,
```
Heimdell: This doesn't look like a "root". This is more like info of the tree.

src/Disciplina/Educator/SizedMerkleTree.hs
```diff
@@ -0,0 +1,237 @@
...
+
+instance Foldable SizedMerkleTree where
+    foldMap _ SizedMerkleEmpty      = mempty
+    foldMap f (SizedMerkleTree n) = foldMap f n
+
+    null SizedMerkleEmpty = True
+    null _                = False
+
+    length SizedMerkleEmpty      = 0
+    length (SizedMerkleTree n) = fromInteger (toInteger (smrSize (mRoot n)))
```
Heimdell: > `fromInteger . toInteger`
Why not `fromIntegral`?

src/Disciplina/Educator/SizedMerkleTree.hs
```diff
@@ -0,0 +1,237 @@
...
+  --show tree = "Merkle tree: "  <> Prelude.show (toList tree)
+
+type LeafIndex = Int
+
+data SizedMerkleNode a
+    = SizedMerkleBranch { mRoot  :: !(SizedMerkleRoot a)
+                        , mLeft  :: !(SizedMerkleNode a)
+                        , mRight :: !(SizedMerkleNode a)}
+    | SizedMerkleLeaf { mRoot  :: !(SizedMerkleRoot a)
+                      , mIndex :: !LeafIndex
```
Heimdell: What meaning does the `index` have?

src/Disciplina/Educator/SizedMerkleTree.hs
```diff
@@ -0,0 +1,237 @@
...
+     (hash $ toLazyByteString $ mconcat
+        [ byteString (one sl)
+        , byteString (one sr)
+        , byteString (convert hl)
+        , byteString (convert hr) ])
+     (sl + sr)
+
+-- | Smart constructor for 'SizedMerkleTree'.
+mkSizedMerkleTree :: (HasHash a, Foldable t) => t a -> SizedMerkleTree a
+mkSizedMerkleTree x = mkSizedMerkleTree' (zip [0,1..] (toList x))
```
Heimdell: Why not `fromFoldable`?

src/Disciplina/Educator/SizedMerkleTree.hs
```diff
@@ -0,0 +1,237 @@
...
+
+mkSizedMerkleTree' :: HasHash a => [(LeafIndex, a)] -> SizedMerkleTree a
```
Heimdell: Why not `fromList`?

src/Disciplina/Educator/SizedMerkleTree.hs
```diff
@@ -0,0 +1,237 @@
...
+    go w = if w .&. (w - 1) == 0 then w else go (w .&. (w - 1))
+
+data MerkleProof a =
+  MerkleProof { getMerkleProof :: [ProofElem a]  -- ^ list of proof elements
+              , getMerkleProofRoot :: Maybe (SizedMerkleRoot a) -- ^ leaf root proof is
+                                                                -- constructed for
+              }
+  deriving (Show, Eq, Ord, Generic, Serialise)
+
+data ProofElem a = ProofElem
```
Heimdell: Why do you have separate structure to represent proof, instead of reusing the tree you described above (in that case, you should add node carrying only hash)? Do you plan to write `insert`/`delete` implementation for this new proof datatype?

src/Disciplina/Educator/SizedMerkleTree.hs
```diff
@@ -0,0 +1,237 @@
...
+              ((xs, _), (ys, Just r))        -> (xs ++ ys, Just r)
+      where
+        lProofElem = ProofElem (mRoot ln) (mRoot rn) L
+        rProofElem = ProofElem (mRoot rn) (mRoot ln) R
+
+        lPath = constructPath (lProofElem:pElems) ln
+        rPath = constructPath (rProofElem:pElems) rn
+
+-- | Validate a merkle tree proof
+validateMerkleProof :: forall a. MerkleProof a -> SizedMerkleTree a -> Bool
```
Heimdell: Why do we need whole tree to validate a proof? It should be validatable using only so called "root". In fact, the single thing you do with the tree is taking that root.

Validation is done on the client side, and the client doesn't have the tree at all; however she should be able to perform validation nonetheless.

src/Disciplina/Educator/SizedMerkleTree.hs
```diff
@@ -0,0 +1,237 @@
...
+    hashProofElem :: ProofElem a -> SizedMerkleRoot a
+    hashProofElem (ProofElem pRoot sibRoot side) =
+      case side of
+        L -> mkBranchRootHash pRoot sibRoot
+        R -> mkBranchRootHash sibRoot pRoot
+
+validateMerkleProof (MerkleProof _ Nothing) _ = False
+
+
+-- | TODO, create tests and put in test module
```
Heimdell: I'd recommend putting the tests in separate `./test/` folder and using quickcheck/smallcheck/hedgehog and some test frameworks (you'd better ask @flyingleafe what test framework to use).

I also recommend having a REPL.hs file so you can rearrange situation from the test to play with it, since the test engine will generate random tree each time.

src/Disciplina/Educator/SizedMerkleTree.hs
```diff
@@ -0,0 +1,237 @@
...
+mkSizedMerkleTree' [] = SizedMerkleEmpty
+mkSizedMerkleTree' ls = SizedMerkleTree (go lsLen ls)
+  where
+    lsLen = Universum.length ls
+    go :: HasHash a => Int -> [(LeafIndex, a)] -> SizedMerkleNode a
+    go _  [x] = mkLeaf x
+    go len xs = mkBranch (go i l) (go (len - i) r)
+      where
+        i = powerOfTwo len
+        (l, r) = splitAt i xs
```
Heimdell: Is this method public?

If it is, are there any guarantees that `LeafIndices` will be
1) in order?
2) consequtive?

Is this tree created at once and never changed again?
How do you push new value?

`splitAt n` has `O(n)` time complexity. You actually traverse the list multitude of times, which is not good.

Heimdell:
After looking on the task, I removed some of my claims.

src/Disciplina/Educator/SizedMerkleTree.hs
lambda76: You are right, I will change this to `SizedMerkleRoot a`

src/Disciplina/Educator/SizedMerkleTree.hs
lambda76: I'm not aware of `fromFoldable`, will look it up.

src/Disciplina/Educator/SizedMerkleTree.hs
lambda76: Updated.

src/Disciplina/Educator/SizedMerkleTree.hs
lambda76: I'm happy to move this to separate package if @flyingleafe think we should do this.

volhovm: This question was raised several times. We have already agreed to use `Serialise` everywhere as a standard serialisation abstraction.

src/Disciplina/Educator/SizedMerkleTree.hs
volhovm: `SizedMerkleRoot` has nothing to do with the merkle tree root, so it should be renamed, I agree. Also, I would kindly ask you to arrange definitions in a readable inductive way -- this is the first datatype introduced and it's only used in the `SizedMerkleNode` which is defined after it's used in `SizedMerkleTree`.

src/Disciplina/Educator/SizedMerkleTree.hs
```diff
@@ -0,0 +1,237 @@
...
+    foldMap f (SizedMerkleTree n) = foldMap f n
+
+    null SizedMerkleEmpty = True
+    null _                = False
+
+    length SizedMerkleEmpty      = 0
+    length (SizedMerkleTree n) = fromInteger (toInteger (smrSize (mRoot n)))
+
+instance Show a => Show (SizedMerkleTree a) where
+  show SizedMerkleEmpty = "Empty Sized Merkle tree "
```
volhovm: Nit: 2 spaces instead of 4. This is not the only place btw.

src/Disciplina/Educator/SizedMerkleTree.hs
```diff
@@ -0,0 +1,237 @@
...
+  show (SizedMerkleTree n) = "Sized Merkle tree:  " <> Prelude.show n
+  --show tree = "Merkle tree: "  <> Prelude.show (toList tree)
+
+type LeafIndex = Int
+
+data SizedMerkleNode a
```
volhovm: Instead, you could have `Node = (Info, Node, Node)` and `Leaf a = (a, hash)`, which would make it possible to inline `SizedMerkleRoot` into `Node` constructor directly. `smrSize` is always 1, so it's a redundant info.

src/Disciplina/Educator/SizedMerkleTree.hs
```diff
@@ -0,0 +1,237 @@
...
+    }
+
+mkBranchRootHash :: SizedMerkleRoot a -- ^ left merkle root
+                 -> SizedMerkleRoot a -- ^ right merkle root
+                 -> SizedMerkleRoot a
+mkBranchRootHash (SizedMerkleRoot (AbstractHash hl) sl)
+                 (SizedMerkleRoot (AbstractHash hr) sr)
+  = SizedMerkleRoot
+     (hash $ toLazyByteString $ mconcat
+        [ byteString (one sl)
```
volhovm: Is it `byteString :: ByteString -> Builder`? Why not `mconcat $ map byteString $ sl : sr : (convert hl ++ convert hr)` then?

Also why do you even use a builder here?

src/Disciplina/Educator/SizedMerkleTree.hs
```diff
@@ -0,0 +1,237 @@
...
+
+import Disciplina.Crypto (Hash, HasHash, hash, unsafeHash)
+import Disciplina.Crypto.Hash.Class (HashFunc, AbstractHash (..))
+
+import Codec.Serialise (Serialise(..), serialise)
+import Data.Bits (Bits (..))
+import Data.ByteArray (ByteArrayAccess, convert)
+import Data.ByteString.Builder (Builder, byteString, lazyByteString)
+import Data.Foldable (Foldable (..))
+import Prelude (show)
```
volhovm: What's wrong with the universum's `show`? If you need to define your own `Show` instance, just `import qualified Prelude`.

src/Disciplina/Educator/SizedMerkleTree.hs
lambda76: Fixed.

src/Disciplina/Educator/SizedMerkleTree.hs
lambda76: yes, it's `ByteString -> Builder`. This is how it is done in cardano-sl, I guess it's for performance reasons.

volhovm: I see, ok.

src/Disciplina/Educator/SizedMerkleTree.hs
volhovm: Is `size` the only example? Then implement a separate function for it :)
I believe that's an easy tradeoff to figure out.

src/Disciplina/Educator/SizedMerkleTree.hs
lambda76: Do you mean I should use some other function to implement this? Or is it just a comment on changing the function name to fromFoldable?

volhovm: I believe this is about naming: https://www.stackage.org/lts-11.10/hoogle?q=fromFoldable

src/Disciplina/Educator/SizedMerkleTree.hs
Heimdell: The term "root" is usually connected to node itself. How about "signature"?

disciplina.cabal
```diff
@@ -62,6 +62,7 @@ library
                      , Disciplina.Educator.Block
                      , Disciplina.Educator.Txs
                      , Disciplina.Educator.State
+                     , Disciplina.Educator.SizedMerkleTree
```
flyingleafe: I think that `SizedMerkleTree` module fits better in `Disciplina.Crypto` package. It isn't very specific to `Educator` logic, it's a generic structure, even though right now it's only used by `Educator` node.

src/Disciplina/Educator/SizedMerkleTree.hs
```diff
@@ -0,0 +1,237 @@
...
+       01100111     x - 1
+       --------
+       01100000     x .&. (x - 1)
+     -}
+    go w = if w .&. (w - 1) == 0 then w else go (w .&. (w - 1))
+
+data MerkleProof a =
+  MerkleProof { getMerkleProof :: [ProofElem a]  -- ^ list of proof elements
+              , getMerkleProofRoot :: Maybe (SizedMerkleRoot a) -- ^ leaf root proof is
+                                                                -- constructed for
```
flyingleafe: I think that formatting like
```
data MerkleProof a = MerkleProof
    { ...
    , ...
    }
```
is way better.

src/Disciplina/Educator/SizedMerkleTree.hs
flyingleafe: Why do you need a `REPL.hs` when you have `stack ghci`, @Heimdell?

flyingleafe: Otherwise I agree, @lambda76 you better create a test module (similar to existing ones in `test` subdir) using Tasty + hspec from the very beginning.

src/Disciplina/Educator/SizedMerkleTree.hs
flyingleafe: @volhovm I think we should have separate `SizedMerkleRoot` datatype still, because these roots are meant to be published and used by clients without any branches.

src/Disciplina/Educator/SizedMerkleTree.hs
lambda76: ok

Heimdell:
@flyingleafe I used it to recreate tests in controlled manner. The `stack ghci` doesn't allow any prefabs to survive after `:r`.

disciplina.cabal
lambda76: Ok. I will move it to Disciplina.Crypto

src/Disciplina/Crypto/MerkleTree.hs
```diff
@@ -0,0 +1,232 @@
...
+    toLazyByteString :: Builder -> LBS.ByteString
+    toLazyByteString = Builder.toLazyByteStringWith (Builder.safeStrategy 1024 4096) mempty
+
+-- | Smart constructor for MerkleTree.
+fromFoldable :: (HasHash a, Foldable t) => t a -> MerkleTree a
+fromFoldable = fromList . F.toList
+
+-- | Smart constructor for MerkleTree.
+fromContainer :: (HasHash (Element t), Container t) => t -> MerkleTree (Element t)
+fromContainer = fromFoldable . toList
```
flyingleafe: Why do you use `fromFoldable`, but not `fromList` here? This way you basically do `toList` twice, it seems.

src/Disciplina/Crypto/MerkleTree.hs
```diff
@@ -0,0 +1,232 @@
...
+data Side = L | R
+  deriving (Show, Eq, Ord, Generic, Serialise)
+
+-- | Construct merkle proof by recusive walking the tree and collecting ProofElem until we hit
+-- matching leafRoot.
+mkMerkleProof :: forall a. MerkleTree a -- ^ merkle tree we want to construct a proof from
+                        -> LeafIndex    -- ^ leaf index used for proof
+                        -> MerkleProof a
+mkMerkleProof MerkleEmpty _ = MerkleProof [] Nothing
+mkMerkleProof (MerkleTree rootNode) n =
```
flyingleafe: In the task requirements there was also options of providing a set or range of indices and getting the _batch proof_ (i. e. a single subgraph of a Merkle tree including all required leaves). This feature is important, because it's required for preparing a data response for data disclosure.
So I propose the following:
- Change the type for proof (make it not a list of nodes with hanging siblings, but an actual tree with some internal nodes allowed to have no children)
- Make a generic method for preparing a proof for _set_ of indices (let's omit ranges for now), and implement method for a single index as a special case.

Heimdell:
Again, its may be worth it to use AVL+ as a base of implementation. It has all the things done, including proofs.

PR #21, opened 2018-06-07
[WIP] [DSCP-61] [DSCP-86] Define Educator state using Snowdrop

flyingleafe:
- [x] Define `Block`, `BlockHeader`, `BlockPayload` and `BlockUndo` in Snowdrop-compatible way
- [x] Define `BlkConfiguration` for private chain
- [x] Define `BlkStateConfiguration` for private chain
- [ ] Implement `Expander` to transform raw private block into Snowdrop-compatible one
- [ ] Implement `Validator`s to validate transactions.

### DIFF
--- a/disciplina.cabal
+++ b/disciplina.cabal
@@ -51,8 +51,9 @@ library
                      , Disciplina.DB.DSL.Interpret.SimpleTxDB
 
                      -- Core datatypes and functions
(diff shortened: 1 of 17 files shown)


flyingleafe:
As described in https://issues.serokell.io/issue/DSCP-98, we decided not to use Snowdrop for private chain. Closing then.

PR #22, opened 2018-06-09
[DSCP-86] [DSCP-97] Multi-subject courses

flyingleafe:
Version of #21 with commits regarding Snowdrop implementation removed.


src/Disciplina/Crypto/Serialise.hs
```diff
@@ -53,3 +55,10 @@ instance {-# OVERLAPPABLE #-}
     HasAbstractSignature ss a where
     unsafeAbstractSign sk = unsafeAbstractSign sk . serialise
     unsafeAbstractVerify pk = unsafeAbstractVerify pk . serialise
+
+-- | Type alias for denoting raw bytes. Indended to be used with hashes
+-- and signatures, like in type `Hash Raw`, and not type-safe hashing and
+-- signing.
+-- TODO: probably it makes sense to make it a newtype, like in Cardano?
+-- Also, is it the most appropriate place for it?
+type Raw = LBS.ByteString
```
lambda76: Should we use `Raw` type in `mrHash` field in `MerkleSignature`?

flyingleafe: Yes, that would be nice

src/Disciplina/DB/DSL/Interpret/SimpleTxDB.hs
```diff
@@ -1,92 +1,147 @@
...
-               . filtered ((==sId).ciSubject._ptxCourseId)
+evalSimpleTxsQuery (SELECTTxs _ (TxHasSubjectId sId)) = do
+    db <- asks _getSimpleObj
+    subjToCourseMap <- asks _courseIdToSubjectId
+    let txs = db ^.. traversed
+                  . _SSTx
+                  . filtered (subjectIdHasCourseId subjToCourseMap . _ptxCourseId)
+    return txs
+  where subjectIdHasCourseId subjToCourseMap courseId =
+          case subjToCourseMap ^? at courseId . _Just . to (any (==sId)) of
```
flyingleafe: It's simpler without lenses:
```
any (== sId) <$> M.lookup courseId subjToCourseMap
```

src/Disciplina/DB/DSL/Interpret/SimpleTxDB.hs
```diff
@@ -1,92 +1,147 @@
...
-  where isDescendantOf x y = hasPathFromTo activityTypeGraphIndexed x y
+evalSimpleTxsQuery (SELECTTxs _ (TxHasDescendantOfSubjectId sId)) = do
+    db <- asks _getSimpleObj
+    subjToCourseMap <- asks _courseIdToSubjectId
+    let txs = db ^.. traversed
+                  . _SSTx
+                  . filtered (hasDescendantOf subjToCourseMap . _ptxCourseId)
+    return txs
+  where hasDescendantOf subjToCourseMap courseId =
+          case subjToCourseMap ^? at courseId . _Just . to (any (isDescendantOf sId)) of
```
flyingleafe: Same as in previous case. Also, could be done more efficiently by implementing a more general `hasPathFromTo` method which accepts a list of possible "to" vertices, but it's definitely not essential here.

PR #23, opened 2018-06-11
[DSCP-106] Switch to `ReaderT ctx m`-based monad stack

Martoon-00:
This includes:
* Introduction of node roles - educator, witness.
* NodeContext, which includes common part of context for every node + parts
specific for given specific role.
* WorkModes - basic, for educator and for witness, in code we abstract
from specific monad and require implementation of needed workmode.
* Removal of Mockables and Production.
** Using `async` instead of `fork`. Added alias for module with `async`
functions.
** Networking left disabled as soon as it heavily relies on implementation
of Mockables. To be fixed in DSCP-105.


Martoon-00:
Thanks, with your comments this forms into complete and reasonable picture, going to fix soon.

Martoon-00:
In recent commits I introduced two typeclasses which serve for `*Params -> *Resources` and `*Resources -> NodeContext *` conversions. Hope this makes sense (although I don't understand potential difference between resources pack and node context).

src/Disciplina/Launcher/Resource.hs
```diff
@@ -1,27 +1,70 @@
...
+        bracketResource bnpLoggingParams $ \bnrLoggerName ->
+        bracketResource bnpDBType bnpDBPath $ \bnrDB ->
+        action BasicNodeResources {..}
+    @
+
+    Make sure you do not drop continuation anywhere!
+    (this won't happen if you do not do any Cont stuff by hand).
+    -}
+    bracketResourceC :: param -> ContT r IO res
+    bracketResourceC = ContT . bracketResource
```
flyingleafe: K P A C U B O
(no irony, pretty indeed)

src/Disciplina/Launcher/Mode.hs
```diff
@@ -1,49 +1,69 @@
...
+    modifyLoggerName name = local $ ncLoggerName %~ name
+
+---------------------------------------------------------------------
+-- Misc
+---------------------------------------------------------------------
 
-type BasicRealMode = ReaderT BasicRealContext Production
+-- | Forms context using given pack of resources.
+class FormNodeContext res ctx | ctx -> res, res -> ctx where
+    formNodeContext :: res -> IO ctx
```
flyingleafe: Frankly, this typeclass seemingly doesn't add a lot of convenience - it basically has only two instances (and will almost always have), and for these two instances it would be quite straightforward to just have two functions `formEducatorNodeContext` and `formWitnessNodeContext`, which you'll be able to pass as a param to `runRealMode`.
(I don't mean to be like Vieth, but this class really seems to complicate the code more than simplify it.)

src/Disciplina/WorldState/Mode.hs
```diff
@@ -0,0 +1,59 @@
+{-# LANGUAGE TemplateHaskell #-}
+
+-- | Module contains the definition of Witness's WorkMode and its implementations.
+
+module Disciplina.WorldState.Mode
```
flyingleafe: Probably it's better to have `Disciplina.Witness.Launcher.Mode` for sake of uniformity. (Also, yes, you can additionally rename the whole `Disciplina.WorldState` to `Disciplina.Witness`, it would be nice).

src/Disciplina/Launcher/Mode.hs
Martoon-00: Okay, I was just too lazy to create new two modules, will do now

src/Disciplina/WorldState/Mode.hs
Martoon-00: :tada:

src/Disciplina/Launcher/Mode.hs
flyingleafe: @Martoon-00 didn't get it, which two new _modules_? :thinking:

Martoon-00: Oh not, I can just place those functions near to resources types, this typeclass really doesn't make sense

Martoon-00: Sorry, just saw your comment. I thought dedicated `runEducatorMode` should be places in separate `Disciplina.Educator.Launcher.Runner` module (and I did so eventually)

src/witness/Main.hs
```diff
@@ -5,48 +5,50 @@ module Main where
...
+                { bnpLoggingParams = wpLogParams
+                }
+            , wpDBParams = DBParams{ dbpPath = wpDbPath }
             }
-    runProduction . bracketBasicNodeResources basicParams $
-        \nr -> runBasicRealMode nr $
-          bracketTransportTCP (15000 {-- connection timeout ms--})
-                              (TCP.defaultTCPAddr "127.0.0.1" "10128") $ \transport -> do
+    launchWitnessRealMode witnessParams $ do
+          -- bracketTransportTCP (15000 {-- connection timeout ms--})
```
volhovm: To be removed.

src/Disciplina/Launcher/Resource.hs
```diff
@@ -1,27 +1,70 @@
...
+----------------------------------------------------------------------------
+
+{- No top-level @acquire@s and @release@s for intension, having bracket built
+from brackets of smaller resources automatially forces proper coupling of
+acquire and release operations and so it's easier to use in various types of
+node which require different resources.
+-}
+
+-- | Resources construction.
+class BracketResource param res | param -> res, res -> param where
```
volhovm: 1. This will not work for the same reasons you can't use global `ResourceT` if you intend to fork later (and we definitely will).
2. In any case I would recommend to use standard resource allocation frameworks, as already mentioned `resourcet` or `componentm`.

src/Disciplina/Educator/Launcher/Resource.hs
```diff
@@ -0,0 +1,23 @@
...
+
+import Universum
+
+import Disciplina.Educator.Launcher.Params (EducatorParams (..))
+import Disciplina.Launcher.Resource (BasicNodeResources (..), BracketResource (..))
+
+-- | Datatype which contains resources required by all Disciplina nodes
+-- to start working.
+data EducatorResources = EducatorResources
+    { erBasicResources :: !BasicNodeResources
```
volhovm: Each educator is also a witness.

src/Disciplina/Launcher/Mode.hs
```diff
@@ -1,49 +1,58 @@
...
     )
 
 ---------------------------------------------------------------------
 -- WorkMode implementations
 ---------------------------------------------------------------------
 
-data BasicRealContext = BasicRealContext
-    { _brcLoggerName :: LoggerName
+-- | For node role returns related context.
+type family CustomContext r = cc | cc -> r
```
volhovm: This is an absolutely useless abstraction. You could just define `EducatorCustomContext = EducatorCustomContext { ecContext :: NodeContext, whatever :: WhateverContext }` instead, and I can't see any reason not to do that.

Martoon-00: Then one will need to define basic typeclasses like `MonadLogging` for both `EducatorMode` and `WitnessMode` in the same way but separately.

src/Disciplina/Launcher/Resource.hs
Martoon-00: Could you elaborate on "global `ResourceT`"? Some scenario when we fork threads and don't wait for them to complete?

src/Disciplina/Launcher/Mode.hs
flyingleafe: @Martoon-00 
We can still do this and use approach with `Has*` lenses to not define such typeclasses. I personally do not like this approach too much, but I admit that it's more general than current one. We'll discuss it with @volhovm in more detail tomorrow.

src/Disciplina/Launcher/Resource.hs
flyingleafe: @Martoon-00 
Misha is right, I missed the fact that `StM (ContT m a)` isn't equal to `m a`, which means that forking inside it is unsafe. `ResourceT` is basically very similar to your construction with `ContT`, and we already had problems with `ResourseT` and concurrency in Cardano because of that - every forked process loses the exception handling continuation. 
We have two options: don't try to use pretty abstractions for resource handling at all and do `release` and `acquire` manually, or think very carefully about how can we use `ResourceT` (or similar) with `Control.Concurrent.Async.Lifted` (without `Safe`).

src/Disciplina/Educator/Mode.hs
```diff
@@ -0,0 +1,55 @@
...
+    ( Witness.WitnessWorkMode m
+    )
+
+---------------------------------------------------------------------
+-- WorkMode implementation
+---------------------------------------------------------------------
+
+data EducatorContext = EducatorContext
+    { _ecWitnessCtx :: Witness.WitnessContext
+
```
flyingleafe: Unnecessary blank line.

src/Disciplina/Educator/Mode.hs
```diff
@@ -0,0 +1,55 @@
...
+import Ether.Internal (HasLens (..))
+import System.Wlog (HasLoggerName (..))
+
+---------------------------------------------------------------------
+-- WorkMode class
+---------------------------------------------------------------------
+
+-- | Set of typeclasses which define capabilities of Educator node.
+type EducatorWorkMode m =
+    ( Witness.WitnessWorkMode m
```
flyingleafe: Nope. `EducatorWorkMode` should include only `BasicWorkMode` plus Educator-specific stuff, like SQL interface. This mode will be used for all Educator-specific logic. Also, we'll have a separate mode (like `CombinedWorkMode` or stuff) for pieces of logic which needs both.

src/Disciplina/Witness/Mode.hs
```diff
@@ -0,0 +1,60 @@
...
+---------------------------------------------------------------------
+-- Instances
+---------------------------------------------------------------------
+
+instance HasLens NodeDB WitnessContext NodeDB where
+    lensOf = wcDB
+
+instance {-# OVERLAPPING #-} HasLoggerName WitnessRealMode where
+    askLoggerName = view wcLoggerName
+    modifyLoggerName name = local $ wcLoggerName %~ name
```
flyingleafe: Do `HasLens` and `HasLoggerName` imply `WithLogger` and `MonadDB` constraints? I didn't see any default instances for these (maybe I'm blind, dunno)

src/Disciplina/Async.hs
```diff
@@ -0,0 +1,19 @@
+{- | This module is an alias for lifted-async stuff.
+
+Among the project we use monad stack consisting of @ReaderT ctx m@,
+and thus tend to use 'Control.Concurrent.Async.Lifted.Safe' module for
+threading because it perfectly fits for IO-based monad stacks which have no
+state embeded into them, i.e. @StM m a ~ a@.
+(Among alternatives there are 'Control.Concurrent.Async' for 'IO' and
+'Control.Concurrent.Async.Lifted' for monad stack with embeded state which puts
+the burden of carrying state around on its user)
```
gromakovsky: Another alternative is `UnliftIO.Async`.

src/witness/Main.hs
```diff
@@ -5,48 +5,42 @@ module Main where
...
             logInfo "Starting node"
-            node (simpleNodeEndPoint transport) (const noReceiveDelay) (const noReceiveDelay)
-                 prng1 serialisePacking (B8.pack "I am node 1") defaultNodeEnvironment $ \node1 ->
-                        NodeAction (witnessListeners . nodeId $ node1) $ \converse -> do
-                            mapM_ (\w -> fork $ w (nodeId node1) [] converse) witnessWorkers
-                            logInfo "Hit return to stop"
-                            _ <- liftIO getChar
-                            logInfo "Stopping node"
+            -- TODO: This networking can't live without Production and Mockables
+            --       so leaving it commented for now
```
gromakovsky: Btw, if you are using `cardano-sl`'s `Node` and stuff, note that in `develop` it uses `IO` nowadays instead of `Mockable`s.

flyingleafe: @gromakovsky We don't use it anymore.

src/Disciplina/Witness/Mode.hs
Martoon-00: I seems so, it's written by you: https://github.com/DisciplinaOU/disciplina/blob/martoon%2Fdscp106-true-modes/src/Disciplina/DB/Real/Functions.hs#L70

flyingleafe: :thinking: :thinking: :thinking:

src/Disciplina/Async.hs
flyingleafe: @Martoon-00 
I've read documentation on `UnliftIO` (https://www.stackage.org/package/unliftio) and I think we should give it a shot. Async interface for it (https://www.stackage.org/haddock/lts-11.13/unliftio-0.2.7.0/UnliftIO-Async.html) is much simpler and doesn't use all these `Forall (Pure m)` constraints, which makes code simple and quicker to build.

src/Disciplina/Launcher/Resource.hs
```diff
@@ -1,26 +1,29 @@
...
-data BasicNodeResources = BasicNodeResources
-    { bnrLoggerName :: !LoggerName
-    , bnrDB         :: !NodeDB
-    }
+----------------------------------------------------------------------------
+-- Resources
+----------------------------------------------------------------------------
+
+-- | Resources safe allocation.
+class AllocResource param resource | param -> resource, resource -> param where
```
volhovm: I doubt this typeclass has any value.

PR #25, opened 2018-06-13
[WIP] [DSCP-100] Pure private blocks validation logic

lambda76:
- [x] Update private transaction types
- [x] Create private block validation function 
- [x] Make project compile
- [x] Make SimpleTxDB tests pass
- [x] Write validator tests

### DIFF
--- a/src/Dscp/Educator/Validate.hs
+++ b/src/Dscp/Educator/Validate.hs
@@ -17,9 +17,10 @@ import Dscp.Educator.Serialise ()
 import Dscp.Educator.Txs (PrivateTx (..), PrivateTxAux (..), PrivateTxWitness (..))
 
 -- | Validate private block.
--- Block is valid if header matches body
--- and all transactions have valid signatures and signature public key
--- correspond to student public key
+-- Block is valid if
+-- 1. Header matches body
+-- 2. All transactions have valid signatures
+-- 3. Transaction signature public key correspond to student public key
 validate :: (MonadError [Text] m) => PrivateBlock -> m ()
 validate pb =
     let txs = pb ^. pbBody . pbbTxs
--- a/src/Dscp/Educator/Validate.hs
+++ b/src/Dscp/Educator/Validate.hs
@@ -1,0 +1,32 @@
+
+-- | Validation of private block
+
+module Dscp.Educator.Validate
+       ( validate
+       ) where
+
+import Universum
+
+import Dscp.Crypto (hash, verify)
+import Dscp.Crypto.MerkleTree (fromFoldable, getMerkleRoot)
+import Dscp.Educator.Block (PrivateBlock, pbBody, pbHeader, pbbTxs, pbhBodyProof)
+import Dscp.Educator.Serialise ()
+import Dscp.Educator.Txs (PrivateTxAux (..), PrivateTxWitness (..))
+
+-- | Validate private block.
+-- Block is valid if header matches body
+-- and all transactions have valid signatures.
+validate :: PrivateBlock -> Bool
+validate pb =
+    let txs = pb ^. pbBody . pbbTxs
+        merkleRoot = getMerkleRoot (fromFoldable txs)
+        headerProof = pb ^. pbHeader . pbhBodyProof
+        headerProofValid = headerProof == merkleRoot
+        txsSigProofValid = all (==True) (map validateTx txs)
+    in headerProofValid && txsSigProofValid
+  where validateTx :: PrivateTxAux -> Bool
+        validateTx (PrivateTxAux {..}) =
+          let txProofSig = _ptwSig _ptaWitness
+              txProofKey = _ptwKey _ptaWitness
+              tx = hash _ptaTx
+          in verify txProofKey tx txProofSig
--- a/src/Dscp/Educator/Validate.hs
+++ b/src/Dscp/Educator/Validate.hs
@@ -2,27 +2,27 @@
 -- | Validation of private block
 
 module Dscp.Educator.Validate
-       ( validate
+       ( validatePrivateBlk
        ) where
 
 import Universum
 
-import Control.Monad.Except (MonadError)
 import Serokell.Util.Verify (verResToMonadError, verifyGeneric)
 
-import Dscp.Core (Address (..), SignedSubmission (..), Submission (..))
+import Dscp.Core (Address (..), SignedSubmission (..), Submission (..),
+                  SubmissionWitness (..))
 import Dscp.Crypto (fromFoldable, getMerkleRoot, hash, verify)
 import Dscp.Educator.Block (PrivateBlock (..), PrivateBlockHeader (..), pbBody, pbbTxs)
 import Dscp.Educator.Serialise ()
-import Dscp.Educator.Txs (PrivateTx (..), PrivateTxAux (..), PrivateTxWitness (..))
+import Dscp.Educator.Txs (PrivateTx (..))
 
 -- | Validate private block.
 -- Block is valid if
 -- 1. Header matches body
 -- 2. All transactions have valid signatures
 -- 3. Transaction signature public key correspond to student public key
-validate :: (MonadError [Text] m) => PrivateBlock -> m ()
-validate pb =
+validatePrivateBlk :: PrivateBlock -> Either [Text] ()
+validatePrivateBlk pb =
     let txs = pb ^. pbBody . pbbTxs
         merkleRoot = getMerkleRoot (fromFoldable txs)
         headerVer = validateTxHeader (_pbHeader pb) merkleRoot
@@ -31,11 +31,12 @@ validate pb =
   where validateTxHeader (PrivateBlockHeader {..}) merkleRoot =
           [(_pbhBodyProof == merkleRoot, "Header signature missmatch")]
 
-        validateTx (PrivateTxAux {..}) =
-          let txProofSig = _ptwSig _ptaWitness
-              txProofKey = _ptwKey _ptaWitness
-              tx = hash _ptaTx
-              txAddrHash = addrHash (sStudentId (ssSubmission (_ptxSignedSubmission _ptaTx)))
-          in [(verify txProofKey tx txProofSig, "Tx signature not valid")
-             ,(hash txProofKey == txAddrHash, "Tx public key missmatch")
-             ]
+        validateTx (PrivateTx {..}) =
+          let submission = ssSubmission _ptxSignedSubmission
+              witnessKey = _swKey (ssWitness _ptxSignedSubmission)
+              witnessSign = _swSig (ssWitness _ptxSignedSubmission)
+              txAddrHash = addrHash (sStudentId submission)
+              witnessSigValid = verify witnessKey (hash submission) witnessSign
+              witnessKeyValid = hash witnessKey == txAddrHash
+          in [(witnessKeyValid, "Tx public key missmatch")
+             ,(witnessSigValid, "Tx signature not valid")]
--- a/src/Dscp/Educator/Validate.hs
+++ b/src/Dscp/Educator/Validate.hs
@@ -7,26 +7,34 @@ module Dscp.Educator.Validate
 
 import Universum
 
-import Dscp.Crypto (hash, verify)
-import Dscp.Crypto.MerkleTree (fromFoldable, getMerkleRoot)
-import Dscp.Educator.Block (PrivateBlock, pbBody, pbHeader, pbbTxs, pbhBodyProof)
+import Control.Monad.Except (MonadError)
+import Serokell.Util.Verify (verResToMonadError, verifyGeneric)
+
+import Dscp.Core (Address (..), SignedSubmission (..), Submission (..))
+import Dscp.Crypto (fromFoldable, getMerkleRoot, hash, verify)
+import Dscp.Educator.Block (PrivateBlock (..), PrivateBlockHeader (..), pbBody, pbbTxs)
 import Dscp.Educator.Serialise ()
-import Dscp.Educator.Txs (PrivateTxAux (..), PrivateTxWitness (..))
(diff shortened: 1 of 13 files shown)


src/Disciplina/Core/Types.hs
```diff
@@ -62,15 +68,65 @@ newtype CourseId = CourseId
...
+    { ssSubmission :: !Submission
+    -- ^ Student submission
+    , ssWitness    :: !SubmissionWitness
+    -- ^ Submission witness
+    } deriving (Eq, Show, Generic)
+
+-- | Witness contains data required to verify transaction.
+-- Included 'PublicKey' belongs either to Student or Educator.
+-- TODO: maybe we can say that Educator's key is already known
+-- to everybody, and not include it into Educator's witness?
```
flyingleafe: Hey, you've just copied the old comment, but it isn't relevant anymore :)

src/Disciplina/Core/Types.hs
```diff
@@ -62,15 +68,65 @@ newtype CourseId = CourseId
...
+data SubmissionWitness = SubmissionWitness
+    { _swKey :: !PublicKey
+    , _swSig :: !SubmissionSig
+    } deriving (Show, Eq, Generic)
+
+-- | Datatype for verifiable transaction (transaction with a witness)
+data SubmissionWitnessAux = SubmissionWitnessAux
```
flyingleafe: What, but this is absolutely the same as `SignedSubmission` :thinking:

src/Disciplina/Core/Types.hs
```diff
@@ -62,15 +68,65 @@ newtype CourseId = CourseId
...
+    { sStudentId  :: !StudentId
+    -- ^ Student who created this submission
+    , sType       :: !SubmissionType
+    -- ^ Submission type
+    , sAssignment :: !Assignment
+    -- ^ Assignment of this submission
+    } deriving (Eq, Show, Generic)
+
+-- | Type alias for Submission signature.
+type SubmissionSig = Signature Submission
```
flyingleafe: Better to have `Signature (Hash Submission)` (and `Hash Submission` is basically `SubmissionId`)

src/Disciplina/Educator/Txs.hs
```diff
@@ -4,70 +4,27 @@
...
-    , _ptxEducatorId :: !EducatorId
-    -- ^ Included to simplify educator signature checking
-    -- on student's side.
-    , _ptxPayload    :: !PrivateTxPayload
-    -- ^ Actual contents of transaction.
+    { _ptxSignedSubmission  :: !SignedSubmission
+    -- ^ Every transaction contains one signed student submission
+    , _ptxGrade             :: !Grade
+    -- ^ Grade for this submission
+    , _ptxTime              :: !Integer
```
flyingleafe: Why not `UTCTime`, for example? If it's stored in DB as an integer timestamp, it doesn't mean we cannot have more semantically valid type in Haskell.

src/Disciplina/Educator/Validate.hs
```diff
@@ -0,0 +1,32 @@
...
+        headerProof = pb ^. pbHeader . pbhBodyProof
+        headerProofValid = headerProof == merkleRoot
+        txsSigProofValid = all (==True) (map validateTx txs)
+    in headerProofValid && txsSigProofValid
+  where validateTx :: PrivateTxAux -> Bool
+        validateTx (PrivateTxAux {..}) =
+          let txProofSig = _ptwSig _ptaWitness
+              txProofKey = _ptwKey _ptaWitness
+              tx = hash _ptaTx
+          in verify txProofKey tx txProofSig
```
flyingleafe: You forgot that `PublicKey` from witness should correspond to `StudentId` (i. e. `StudentId` should be an `Address` derived from `PublicKey`.

src/Disciplina/Educator/Validate.hs
```diff
@@ -0,0 +1,32 @@
...
+-- Block is valid if header matches body
+-- and all transactions have valid signatures.
+validate :: PrivateBlock -> Bool
+validate pb =
+    let txs = pb ^. pbBody . pbbTxs
+        merkleRoot = getMerkleRoot (fromFoldable txs)
+        headerProof = pb ^. pbHeader . pbhBodyProof
+        headerProofValid = headerProof == merkleRoot
+        txsSigProofValid = all (==True) (map validateTx txs)
+    in headerProofValid && txsSigProofValid
```
flyingleafe: This is all good, but in real life you'll probably want to at least log some error messages if some checks failed. In order to provide some info about failure if validation has failed, you may use `VerificationRes` datatype from `serokell-util` instead of `Bool`: https://github.com/serokell/serokell-util/blob/master/src/Serokell/Util/Verify.hs

src/Disciplina/Core/Types.hs
```diff
@@ -62,15 +68,65 @@ newtype CourseId = CourseId
...
+    deriving (Eq, Show, Generic)
+
+-- | Assignment doesn't contain actual assignment contents - only hash of them.
+data Assignment = Assignment
+    { aCourseId   :: !CourseId
+    -- ^ Course this assignement belongs to
+    , aType       :: !AssignmentType
+    -- ^ Assignment type
+    , aAssignment :: !Text
+    -- ^ Hash of assignment content
```
flyingleafe: No, this `Text` field is actually supposed to be a short description of an assignment, and a _hash_ of an assignment is a separate field (so the assignment ID is a hash of its contents, not a hash of this datatype).
However, I'm thinking about maybe leaving assignment ID a hash of a datatype, and providing a hash of assignment contents separately - this way more data will be validated by student's submission signature. I'll think about it and tell my decision a little bit later.

src/Disciplina/Educator/Validate.hs
```diff
@@ -0,0 +1,41 @@
...
+import Disciplina.Educator.Serialise ()
+import Disciplina.Core (Submission (..), SignedSubmission (..), Address (..))
+import Disciplina.Crypto (fromFoldable, getMerkleRoot, hash, verify, )
+
+-- | Validate private block.
+-- Block is valid if
+-- 1. Header matches body
+-- 2. All transactions have valid signatures
+-- 3. Transaction signature public key correspond to student public key
+validate :: (MonadError [Text] m) => PrivateBlock -> m ()
```
flyingleafe: This is fully pure code which isn't meant to be run in any IO-based monad, so it's better to simply use `Either [Text] ()` as a return type here. See `exceptions` guide: https://github.com/input-output-hk/cardano-sl/blob/develop/docs/exceptions.md#pure-code-regular-errors

tests/Test/Common.hs
```diff
@@ -164,3 +177,61 @@ worldTProperty
...
+mkPubKey seed = fst (mkKeyPair seed)
+
+-- | Create private key from seed
+mkPrivKey :: Char -> SecretKey
+mkPrivKey seed = snd (mkKeyPair seed)
+
+-- | Create key pair from seed
+mkKeyPair :: Char -> (PublicKey, SecretKey)
+mkKeyPair seed =
+  let (CryptoPassed x) = Ed25519.secretKey (C.replicate 32 seed)
```
flyingleafe: Why do you use `Char` as a key and replicate it to get 32-bit bytestring, why not use a `ByteString` as a seed directly instead?

tests/Test/Disciplina/Educator/Validate.hs
```diff
@@ -0,0 +1,95 @@
...
+
+courseCompScience1 :: CourseId
+courseCompScience1 = CourseId 3
+
+studentAPubKey, studentBPubKey,
+  educatorJPubKey, educatorKPubKey :: PublicKey
+studentAPubKey = mkPubKey 'a'
+studentBPubKey = mkPubKey 'b'
+educatorJPubKey = mkPubKey 'j'
+educatorKPubKey = mkPubKey 'k'
```
flyingleafe: You can just generate a bunch of keypairs randomly in advance, using e. g. `unsafePerformIO`.

src/Disciplina/Educator/Validate.hs
```diff
@@ -0,0 +1,41 @@
...
+  where validateTxHeader (PrivateBlockHeader {..}) merkleRoot =
+          [(_pbhBodyProof == merkleRoot, "Header signature missmatch")]
+
+        validateTx (PrivateTxAux {..}) =
+          let txProofSig = _ptwSig _ptaWitness
+              txProofKey = _ptwKey _ptaWitness
+              tx = hash _ptaTx
+              txAddrHash = addrHash (sStudentId (ssSubmission (_ptxSignedSubmission _ptaTx)))
+          in [(verify txProofKey tx txProofSig, "Tx signature not valid")
+             ,(hash txProofKey == txAddrHash, "Tx public key missmatch")
```
flyingleafe: Typo: should be "mismatch".

flyingleafe: Actually, it's way better to define a custom type for tx validation and block validation errors and define `Show`/`Buildable` instances for it to get proper error messages. This way you also wouldn't need to compare string literals in tests (which is ugly).

src/Disciplina/Core/Types.hs
```diff
@@ -62,15 +67,59 @@ newtype CourseId = CourseId
...
+
+-- | Assignment doesn't contain actual assignment contents - only hash of them.
+data Assignment = Assignment
+    { aCourseId   :: !CourseId
+    -- ^ Course this assignement belongs to
+    , aType       :: !AssignmentType
+    -- ^ Assignment type
+    , aAssignment :: !Text
+    -- ^ Description of assignment
+    } deriving (Eq, Show, Generic)
```
flyingleafe: As I already said, it probably makes sense to store contents' hash for assignment and submission separately. This way we also might get rid of `AssignmentType` inside `Assignment` (the same for `SubmissionType`), because it can be encoded into contents' hash. I'll think about it a little bit and write down the necessary spec in issue's comments.

flyingleafe:
@lambda76 See this comment about final structure of `Assignment` and `Submission`: https://github.com/DisciplinaOU/disciplina/pull/31#issuecomment-397888381

src/Disciplina/Educator/Validate.hs
flyingleafe: Also, naming the function simply `validate` isn't nice, be more specific, e.g. `validatePrivateBlock`.

src/Dscp/Educator/Validate.hs
```diff
@@ -0,0 +1,41 @@
...
+        merkleRoot = getMerkleRoot (fromFoldable txs)
+        headerVer = validateTxHeader (_pbHeader pb) merkleRoot
+        blockValid = verifyGeneric (headerVer <> concatMap validateTx txs)
+    in verResToMonadError toList blockValid
+  where validateTxHeader (PrivateBlockHeader {..}) merkleRoot =
+          [(_pbhBodyProof == merkleRoot, "Header signature missmatch")]
+
+        validateTx (PrivateTxAux {..}) =
+          let txProofSig = _ptwSig _ptaWitness
+              txProofKey = _ptwKey _ptaWitness
```
flyingleafe: Actually, we don't need `PrivateTxWitness` (and, consequently, `PrivateTxAux` type) at all. We decided that there will be no Educator's signatures on transactions, and student will only sign submissions. So here we should only check `SubmissionWitness` against Student's address.

tests/Test/Dscp/Educator/Validate.hs
```diff
@@ -0,0 +1,96 @@
...
+educatorJPrivKey = mkPrivKey 'j'
+educatorKPrivKey = mkPrivKey 'k'
+
+educatorKKeyPair, studentAKeyPair :: (PublicKey, SecretKey)
+educatorKKeyPair = mkKeyPair 'k'
+studentAKeyPair = mkKeyPair 'a'
+
+mkPrivateTxAux :: PublicKey -- ^ Public key used to create student addr
+               -> (PublicKey, SecretKey) -- ^ Key pair used to create submission witness
+               -> (PublicKey, SecretKey) -- ^ Key pair used to create transaction witness
```
flyingleafe: As said in previous comment, we only need submission witness (provided by student). Also, in tx definitions you misplace student and educator keypairs, causing test to fail.

tests/Test/Common.hs
lambda76: Argument to `secretKey` need to be 32 bytes long for pattern match to succeed (will get `CryptoFailed` otherwise). Or do you mean something else?

src/Disciplina/Educator/Validate.hs
lambda76: Sounds good. Could you elaborate on how to do this?

tests/Test/Disciplina/Educator/Validate.hs
lambda76: I added txs generated from unsafePerformIO some lines below. I rather keep these to have control over some of the test keys.

tests/Test/Common.hs
flyingleafe: I meant, why not just pass a `ByteString` insead of `Char` to `mkKeyPair` and check if it's 32 bytes long instead of accepting only `Chars` and constructing a bytestring of 32 equal bytes, which effectively reduces available seed entropy 2<sup>248</sup> times? 
This isn't by any means critical, as it's only used for tests, but I just didn't quite get the idea here.

lambda76: The idea was just to have a simple way to create deterministic key in pure code and to be sure creation always succeeds. It could of course been done in many other ways. One better way would probable be to take `ByteString` and pad it with zeros to length 32 (or shrink it if longer than 32), as you suggested. It doesn't make much difference IMO except for keyspace. I can update the code to use second approach, no problem.

lambda76:
Yes, I had not finished everything yet. Anyway, doing rebase now so I guess no more work on this ticket once rebase is finished and builds.

PR #27, opened 2018-06-16
[DSCP-106] Rename `*Params` types in executables' param modules

Martoon-00:
To avoid name conflict with types defined in `Disciplina/*/Launcher/Params.hs`


Martoon-00:
Executables' `Param.hs` are said to be merged with lib's ones.

PR #29, opened 2018-06-16
[DSCP-105] Integrate loot-network 

volhovm:
Now you can run:
```
stack exec disciplina-witness -- --bind 127.0.0.1:8000:8001 
stack exec disciplina-witness -- --bind 127.0.0.1:8002:8003 --peer 127.0.0.1:8000:8001 
```


src/educator/Main.hs
```diff
@@ -1,27 +1,23 @@
...
-            }
+    educatorParams <- getEducatorParams
     launchEducatorRealMode educatorParams $ do
         logInfo "This is the stub for Educator node executable"
         logWarning "Please don't forget to implement everything else!"
+
+getEducatorParams :: IO EducatorParams
+getEducatorParams =
+    execParser $ info (helper <*> versionOption <*> educatorParamsParser) $
+    fullDesc <> progDesc "Discplina educator node."
```
flyingleafe: Typo: `s/Discplina/Disciplina/`

src/witness/Main.hs
```diff
@@ -8,40 +7,30 @@ import Universum
...
-            --                 logInfo "Hit return to stop"
-            --                 _ <- liftIO getChar
-            --                 logInfo "Stopping node"
+            logInfo "Starting node."
             logInfo "All done."
 
             logInfo "Hey, here log-warper works!"
             logWarning "Don't forget to implement everything else though!"
+
+
```
flyingleafe: Why two blank lines?

src/Dscp/Witness/Launcher/Runner.hs
```diff
@@ -7,18 +6,17 @@ import Universum
...
 import Control.Monad.Component (runComponentM)
 
-import Dscp.Launcher.Resource (AllocResource (..))
+import Dscp.Resource.Class (AllocResource (..))
 import Dscp.Witness.Launcher.Mode (WitnessContext (..), WitnessRealMode)
 import Dscp.Witness.Launcher.Params (WitnessParams (..))
 import Dscp.Witness.Launcher.Resource (WitnessResources (..))
 
+-- TODO Maybe this function should be "-> IO WitnessContext" and other
+-- non-resource context parts can be allocated here.
```
flyingleafe: What are non-resource context parts which require IO to be allocated? Can you provide an example?

src/Dscp/Resource/Network.hs
```diff
@@ -0,0 +1,94 @@
...
+
+-- TODO Should I also release sockets?
+
+----------------------------------------------------------------------------
+-- Client
+----------------------------------------------------------------------------
+
+-- | Client networking params.
+data NetCliParams = NetCliParams
+    { ncPeers :: Set ZTNodeId
```
flyingleafe: Why don't you make your datatypes strict? (Not only here, but in many other places)

src/Dscp/CLI/Common.hs
```diff
@@ -43,3 +53,27 @@ versionOption :: Parser (a -> a)
...
+-- ZMQ TCP
+----------------------------------------------------------------------------
+
+-- | Parse peers to connect to.
+peersParser :: Parser (Set ZTNodeId)
+peersParser =
+    fmap Set.fromList $
+    many $
+    option (eitherReader parseZTNodeId)
+           (long "peer" <> metavar "HOST:PORT1:PORT2" <> help "Peer(s) we should connect to")
```
flyingleafe: Does ZMQ require that exactly two ports are provided? If not (if any number of ports equal to one or more is allowed), then this `metavar` should be like `HOST:PORT1[:PORTx]*`

src/Dscp/Witness/Launcher/Mode.hs
```diff
@@ -32,29 +36,53 @@ import qualified Dscp.Launcher.Mode as Basic
...
+
+----------------------------------------------------------------------------
+-- Sanity check
+----------------------------------------------------------------------------
+
+_sanity :: WitnessRealMode ()
+_sanity = _sanityCallee
+  where
+    _sanityCallee :: WitnessWorkMode m => m ()
+    _sanityCallee = pass
```
Martoon-00: Nice check

src/Dscp/Witness/Launcher/Runner.hs
Martoon-00: Maybe, but it can be changed later easily if needed, we would hardly use this function outside of the module :woman_shrugging:

src/Dscp/CLI/Common.hs
volhovm: Exactly two. We might discuss this later, but I obviously anticipated the confusion here. This model is much better and reliable from the ZMQ standpoint -- one socket is ROUTER (for p2p communication), other is PUB (subscriptions).

src/educator/Main.hs
volhovm: Addressed.

src/Dscp/Resource/Network.hs
flyingleafe: @volhovm cho po fixam

volhovm:
@flyingleafe addressed all of your comments.

src/Dscp/Network/Wrapped.hs
```diff
@@ -0,0 +1,283 @@
...
+
+    withTimeout :: (STM () -> m x) -> m x
+    withTimeout action
+        | timeout < 0 = action retry
+        | otherwise = do
+              timeoutVar <- liftIO newEmptyTMVarIO
+              let timer = do
+                      liftIO $ threadDelay timeout
+                      atomically $ putTMVar timeoutVar ()
+              withAsync timer $ \_async -> action (readTMVar timeoutVar)
```
Martoon-00: How does it work? IIUC If call of `recvBtq` above sees empty queue, `timeoutVar` is performed and you block until `timeout` passes and try to throw `CRETimeout`, and immediatelly after that kill the action without waiting for its complection, why so?

src/Dscp/Network/Wrapped.hs
```diff
@@ -0,0 +1,283 @@
...
+    => ClientEnv t
+    -> Int
+    -> m (NodeId t, d)
+cliRecvUpdate = cliRecvOne @SubK @t @d
+
+----------------------------------------------------------------------------
+-- Launching
+----------------------------------------------------------------------------
+
+-- | Launch server on the background.
```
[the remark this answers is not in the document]
volhovm: You mean "in"? It's not a "typo", but thanks.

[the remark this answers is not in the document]
Martoon-00: Actually I mean `s/server/client` (bad line to comment I've chosen)

src/Dscp/Network/Wrapped.hs
volhovm: I doubt it works like this. In my view, STM's `orElse` takes two "blocking" actions -- `recvBtq` and `readTMVar` -- and then will give me the result which comes earlier. It'd be the other way if one of these actions was not blocking, of course.

I'll test this (should have done it anyway).

I'll test it anyway (I should have).

src/Dscp/Network/Wrapped.hs
```diff
@@ -0,0 +1,283 @@
...
+    , lAction   :: !(ListenerEnv t -> m ())
+    }
+
+runListener ::
+       forall t m n. (NetworkingServ t n, MonadLogging n, MonadMask n)
+    => (forall x. m x -> n x)
+    -> Listener t m
+    -> n ()
+runListener nat Listener{..} = do
+    logDebug $ fromString $ "Launching listner " <> show lId
```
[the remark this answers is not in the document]
volhovm: I still didn't make up my mind regarding whether it's a good lib or not. I'll switch to it if I decide it's alright, but raw strings for now are not that terrible imo  :)

src/Dscp/CLI/Common.hs
volhovm: I've tested timeouts, they work as expected.

PR #30, opened 2018-06-16
[DSCP-104] Loot log

Martoon-00:
This switches logging to one based on lootbox.

There were several technical decisions taken:
* Using `RIO` newtype instead of `ReaderT ctx IO`. Simply speaking, it is done because `MonadLogging` typeclass provided by lootbox is already instantiated with `instance (r ~ Capabilities) => MonadLogging (ReaderT r IO)`, and I wanted to avoid defining overlapping instances.

* Logs of educators and witnesses are assumed to go into different folders (`logs/educator` and `logs/witness` accordingly) via passing `--log-dif` cmd line parameter, so educators and witnesses can share the same config without printing to the same file. (perhaps the time has come to write a simple launcher script).

Also see https://github.com/serokell/lootbox/pull/15.

Can not be merged until:
- [x] Depends on lootbox'es commit not in master
- [ ] Wait for DSCP-105 to be merged (with some neat technical decisions)

### DIFF
--- a/src/Dscp/Witness/Launcher/Mode.hs
+++ b/src/Dscp/Witness/Launcher/Mode.hs
@@ -11,14 +11,12 @@ module Dscp.Witness.Launcher.Mode
     , WitnessContext (..)
     , WitnessRealMode
     , wcDB
-    , wcLoggerName
+    , wcLogging
     ) where
-
-import Universum
 
 import Control.Lens (makeLenses)
 import Ether.Internal (HasLens (..))
-import System.Wlog (HasLoggerName (..), LoggerName)
+import Loot.Log.Rio (LoggingIO)
 
 import Dscp.DB.Class (MonadDB)
 import Dscp.DB.Real.Types (NodeDB)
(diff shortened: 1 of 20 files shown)


src/Dscp/Witness/Launcher/Mode.hs
```diff
@@ -40,12 +38,12 @@ type WitnessWorkMode m =
 
 data WitnessContext = WitnessContext
     { _wcDB         :: NodeDB
-    , _wcLoggerName :: LoggerName
+    , _wcLoggerName :: LoggingIO
```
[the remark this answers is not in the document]
Martoon-00: :man_facepalming:

stack.yaml
```diff
@@ -1,4 +1,5 @@
-resolver: lts-10.8
+resolver:
+    https://raw.githubusercontent.com/serokell/lootbox/e445a9729e6d203234e9a3135d9c1fc99949da39/snapshot.yaml
```
Martoon-00: Still to be removed in favour of copy-paste

src/educator/Main.hs
```diff
@@ -22,6 +22,7 @@ main = do
                 , wpDBParams = DBParams{ dbpPath = epDbPath }
                 }
             }
-    launchEducatorRealMode educatorParams $ do
+    launchEducatorRealMode educatorParams $
+      modifyLogName (<> "node") $ do
```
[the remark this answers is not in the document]
Martoon-00: I hope having educator's and witness'es logs in different folders will be enough

stack.yaml
```diff
@@ -1,4 +1,4 @@
-resolver: lts-10.8
+resolver: lts-10.7
```
flyingleafe: :(

Martoon-00: Maybe that was an extra, rly

yegortimoshenko:
This is going to take a while, our build server is at 100% of its capacity. Working around issue with `-hide-package` (via @kirelagin's patch) caused a mass rebuild of all Haskell packages.

flyingleafe:
@Martoon-00 I'd like it to be merged first, and DSCP-105 later, so I'll merge it now.

PR #31, opened 2018-06-17
[DSCP-109] [WIP] Private chain database schema

Heimdell:


database/schema.sql
```diff
@@ -0,0 +1,181 @@
...
+
+    primary key (addr)
+
+) without rowid;
+
+-- Creating 'StudentCourses' table.
+--
+-- [Appears to be] many-to-many ref between Students and Cources.
+--
+create table if not exists StudentCources (
```
flyingleafe: Typo: should be `Courses`.

database/schema.sql
```diff
@@ -0,0 +1,181 @@
...
+
+create index if not exists Blocks_prev_hash on Blocks (prev_hash);
+
+-- Creating 'Educator' table.
+--
+create table if not exists Educator (
+    signature  BLOB non null,
+
+    primary key (signature)
+) without rowid;
```
flyingleafe: What's this table for, which signatures are stored there?

src/Disciplina/DB/DSL/Interpret/Sqlite3.hs
```diff
@@ -0,0 +1,79 @@
...
+import Control.Monad.Catch
+import Control.Monad.Trans
+
+import qualified Data.Set as Set (Set, singleton, member)
+import Database.SQLite.Simple
+
+import Disciplina.DB.DSL.Class
+
+newtype Sqlite3T m a = Sqlite3T { getSqlite3T :: ReaderT Connection m a }
+    deriving (MonadTrans)
```
flyingleafe: You don't need this particular newtype because we'll have only one `ReaderT` over IO in our work monad. What you should have is:
- some typeclass `MonadSql` which can `query` and `execute` queries or something
- an `instance (HasLens' Connection ctx, MonadIO m) => ReaderT ctx m`, which performs these queries using a `Connection` from a context.
This way you'll automatically provide SQL capabilities to any working monad which will have opened Sqlite connection in its context (which is what we need).

src/Disciplina/DB/DSL/Interpret/Sqlite3.hs
```diff
@@ -0,0 +1,79 @@
...
+
+runSqlite3T :: (MonadIO m, MonadCatch m) => Sqlite3T m a -> String -> m a
+runSqlite3T connectionString action = do
+    connection <- liftIO $ open connectionString
+    res <- getSqlite3T action
+        `runReaderT` connection
+        `finally`    close connection
+    return res
```
flyingleafe: Also, yes, you don't need to open and close the database anywhere but in `*Resource` modules (where you do that using `ComponentM`).

database/schema.sql
```diff
@@ -0,0 +1,181 @@
...
+
+    primary key (hash),
+    foreign key (student_addr)    references Students    (addr),
+    foreign key (assignment_hash) references Assignments (hash),
+
+    -- Composite foreign key.
+    -- We may not need it, actually, but it may help us.
+    --
+    foreign key (student_addr, assignment_hash)
+        references StudentAssignments(student_addr, assignment_hash)
```
flyingleafe: I think we should either use only this composite foreign key (and get student and assignment for submission through join with `StudentAssignments`, or use simple foreign keys. I think we might go the first way - it's more semantically valid (I just didn't think of it while I was doing PDM).

database/schema.sql
```diff
@@ -0,0 +1,181 @@
...
+
+create index if not exists StudentCources_student_addr on StudentCources (student_addr);
+create index if not exists StudentCources_course_id    on StudentCources (course_id);
+
+-- Creating 'Assignments' table.
+--
+create table if not exists Assignments (
+    hash       BLOB     not null,
+    course_id  INTEGER  not null,
+    type       INTEGER  not null,
```
flyingleafe: We'll probably rethink structure of `Assignment`s and `Submission`s a bit (namely, we might have contents hash instead of type tag). I'll add a new comment later.

database/schema.sql
```diff
@@ -0,0 +1,181 @@
...
+    primary key (id asc)
+);
+
+-- Creating 'Subjects' table.
+--
+create table if not exists Subjects (
+    id         INTEGER,
+    course_id  INTEGER  not null,
+
+    primary key (id asc),
```
flyingleafe: 1) Not `asc`, subject IDs are part of ATG and should be the same for the whole network.
2) I made a mistake in PDM, in `Subjects` table there should be composite primary key `(subject_id, course_id)`, because one subject can be related to many courses. I'll fix PDM and add update to it to this PR.

flyingleafe:
@Heimdell Fixed mistakes in PDM and applied some fixes to SQL schema definition already.
Major points:
- Integer indices for blocks are necessary in order to be able to get N consequent blocks efficiently (iterating hash by hash from the top really sucks).
- In `Transactions` table, `idx` field is added to represent transactions ordering inside the block.
- In `Assignments` and `Submissions` tables `type` fields are replaced by `content_hash` field. `content_hash` determines the type of assignments/submissions in the following way:
  -  For submissions, submission is considered `Offline` if `content_hash == <offline_hash>`, where `<offline_hash>` is some constant like `hash "pva228"`
  -  For assignments, assignment is considered `CourseFinal` if `content_hash == <course_final_hash>`, where `<course_final_hash>` is another constant like `hash "skovoroda"`.

database/schema.sql
Heimdell: Its a table to store current `Educator` in use. I thought about storing its signature in the database; however now I see that some keyring storage is better for that.

database/schema.sql
Heimdell: Will fix.

database/schema.sql
flyingleafe: Exactly, we should store keys in a separate keyfile.

database/schema.sql
```diff
@@ -0,0 +1,167 @@
...
+
+) without rowid;
+
+create index if not exists StudentAssigments_student_addr    on StudentAssignments (student_addr);
+create index if not exists StudentAssigments_assignment_hash on StudentAssignments (assignment_hash);
+
+-- Creating 'Submissions' table.
+--
+create table if not exists Submissions (
+    hash             BLOB     not null,
```
lambda76: This hash is the hash of the other fields right?

database/schema.sql
```diff
@@ -0,0 +1,167 @@
...
+    -- Composite primary.
+    --
+    primary key (student_addr, course_id),
+
+    foreign key (student_addr) references Students (addr),
+    foreign key (course_id)    references Courses  (id)
+
+) without rowid;
+
+create index if not exists StudentCourses_student_addr on StudentCourses (student_addr);
```
lambda76: I don't know squlite that well. What kind of index do you get here as default? Is the default index strategy good for BLOB data?

database/schema.sql
flyingleafe: @lambda76 right

database/schema.sql
flyingleafe: @lambda76 SQLite only uses btree indices, so we have no choice here. It would be better to have hashtable indices, but we'd need something like PostgreSQL for that. We don't expect high load on the Educator database, so we resolve this tradeoff in favor of portability and simplicity by using SQLite.

src/Dscp/DB/DSL/Interpret/Sqlite3.hs
```diff
@@ -0,0 +1,188 @@
+
+{-# LANGUAGE QuasiQuotes #-}
+
+module Dscp.DB.DSL.Interpret.Sqlite3 where
```
flyingleafe: Explicit exports, please. https://github.com/serokell/serokell-util/blob/master/serokell-style.md#export-lists

src/Dscp/DB/DSL/Interpret/Sqlite3.hs
```diff
@@ -0,0 +1,188 @@
...
+import Control.Monad.Reader
+import Control.Monad.Trans
+
+import qualified Data.Set as Set (Set, singleton, member, empty)
+
+import Database.SQLite.Simple hiding (query)
+import Database.SQLite.Simple.ToField (ToField(..))
+import Database.SQLite.Simple.FromField (FromField(..))
+
+import Text.InterpolatedString.Perl6
```
flyingleafe: Explicit imports, please. https://github.com/serokell/serokell-util/blob/master/serokell-style.md#imports
The same later.

src/Dscp/DB/DSL/Interpret/Sqlite3.hs
```diff
@@ -0,0 +1,188 @@
+
+{-# LANGUAGE QuasiQuotes #-}
+
+module Dscp.DB.DSL.Interpret.Sqlite3 where
+
+import Universum
+
+import Control.Monad.Reader
+import Control.Monad.Trans
```
flyingleafe: I am quite sure that these are re-exported in Universum already.

src/Dscp/DB/DSL/Interpret/Sqlite3.hs
```diff
@@ -0,0 +1,188 @@
...
+import Dscp.Core.Types ( SignedSubmission(..)
+                       , Submission(..)
+                       , SubmissionType(..)
+                       , SubmissionWitness(..)
+                       , SubmissionSig
+                       , Assignment(..)
+                       , AssignmentType(..)
+                       , Address(..)
+                       , CourseId(..)
+                       , Grade(..)
```
flyingleafe: This seems nice, but better let `stylish-haskell` deal with that (at least for sake of uniformity).

src/Dscp/DB/DSL/Interpret/Sqlite3.hs
```diff
@@ -0,0 +1,188 @@
...
+            left join Assignments on Submissions.assignment_hash = Assignments.hash
+            {
+                additionalJoins
+            }
+            where {buildWhereStatement filterExpr}
+            ;
+        |]
+        ()
+
+buildWhereStatement = go
```
flyingleafe: Explicit type signatures for top-level bindings, please (actually don't know how this isn't a warning :thinking:).

src/Dscp/DB/DSL/Interpret/Sqlite3/Schema.hs
```diff
@@ -0,0 +1,186 @@
...
+
+module Dscp.DB.DSL.Interpret.Sqlite3.Schema (schema) where
+
+import Universum
+
+import Text.InterpolatedString.Perl6 (q)
+
+schema :: IsString s => s
+schema = [q|
+
```
flyingleafe: Wait, what kind of copypaste is that? You already have `database/schema.sql`, just use `quoteFile` here: http://hackage.haskell.org/package/template-haskell-2.7.0.0/docs/Language-Haskell-TH-Quote.html

src/Dscp/DB/DSL/Interpret/Sqlite3.hs
```diff
@@ -0,0 +1,192 @@
...
+import Dscp.Crypto (PublicKey)
+import Dscp.DB.DSL.Class
+import Dscp.DB.DSL.Interpret.Sqlite3.Schema (schema)
+import Dscp.DB.DSL.Interpret.Sqlite3.Instances ()
+import Dscp.DB.SQLite
+import Dscp.Educator.Txs (PrivateTx (..), PrivateTxId)
+
+ensureSchemaIsSetUp :: Connection -> IO ()
+ensureSchemaIsSetUp (Connection db) = do
+    exec db schema
```
flyingleafe: In my world, `Dscp.DB.DSL.Interpret.*` modules are for logic directly related to processing of external search queries, encoded in the search query DSL. This function (together with database schema init query should go somewhere to `Dscp.DB.SQLite.*`. This is also true for `FromRow`/`ToRow` instances.

src/educator/Main.hs
```diff
@@ -13,7 +13,7 @@ import Dscp.Educator (EducatorParams (..), educatorParamsParser, launchEducatorR
 main :: IO ()
 main = do
     educatorParams <- getEducatorParams
-    launchEducatorRealMode educatorParams $
+    launchEducatorRealMode educatorParams $ do
```
flyingleafe: Kind of unnecessary change :thinking:

src/Dscp/DB/DSL/Interpret/Sqlite3.hs
Heimdell: Will move.

src/educator/Main.hs
Heimdell: I looked to place to initialize schema, and my first intent was to place it here; however, I found a proper place in short time, but left this change out. Should I `git rebase -i` it out?

src/Dscp/DB/DSL/Interpret/Sqlite3/Instances.hs
```diff
@@ -0,0 +1,34 @@
...
+
+-- TODO(kir): split into separate .Instances module.
+instance FromField Address where
+    fromField f = Codec.deserialise <$> fromField f
+
+instance FromField CourseId where
+    fromField f = Codec.deserialise <$> fromField f
+
+instance FromField Grade where
+    fromField f = Codec.deserialise <$> fromField f
```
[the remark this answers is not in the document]
Heimdell: Currently it is an `Enum`; we had a thread with Kirill Kuvshinov there we looked on other encodings of grade, which aren't `Enum` at all. So, I rather leave it as it is, for sake of uniformity.

src/Dscp/DB/DSL/Interpret/Sqlite3/Schema.hs
Heimdell: They left spaces inside quasi-quoter in that version of documentation. It took me time to figure out, that spaces do matter there, and they fixed it in latest version (`2.13.0.0`) of docs. My bad, I didn't checked if docs are fresh enough.

src/Dscp/DB/SQLite/SchemaInit.hs
```diff
@@ -0,0 +1,16 @@
+
+{-# LANGUAGE QuasiQuotes #-}
+
+module Dscp.DB.SQLite.SchemaInit
+    ( ensureSchemaIsSetUp
```
flyingleafe: I don't see a reason why `Schema` and `SchemaInit` are two different modules.

Heimdell: I will fix this now. After second thought, only the `Schema` module should remain, consuming `SchemaInit`.

PR #33, opened 2018-06-17
[DSCP-81] Integrate hpack

volhovm:
Saved up some space :)
Also I wanted to turn on `-Werror`, but it appears that there are lots of places with warnings, so I suggest to turn it on later, after we get rid of csl-related dependencies at least.


tools/package.yaml
```diff
@@ -0,0 +1,12 @@
...
+name:        disciplina-tools
+description: This package contains tools for development of Disciplina
+             platform, such as simulators of cryptographic protocols.
+
+executables:
+  disciplina-simulacrum:
+    <<: *exec-common
+    source-dirs: src/simulacrum
+    dependencies:
+      - mtl
```
Martoon-00: Lost newline? (and +1 in module above)

volhovm: Vim does strange things.

tools/disciplina-tools.cabal
```diff
@@ -1,46 +0,0 @@
...
-  main-is:             Main.hs
-  build-depends:       universum
-                     , base >= 4.9
-                     , mtl
-
-  default-language:    Haskell2010
-  ghc-options:         -threaded -rtsopts
-                       -Wall
-                       -fno-warn-orphans
-                       -O2
```
Martoon-00: Does `hpack` add `-O2` automatically?

volhovm: It doesn't, but I doubt it was ever actually needed. Will return it though.

PR #34, opened 2018-06-18
[DSCP-102] Provide SQLite context

Martoon-00:
Within this PR we do:
* add access to SQL DB operations into context to educator
* along with adding dedicated typeclasses into workmode
* allowing both real and in-memory work in theory (thus no separated `SQLite/Real/*` modules)
* rename `DB` -> `RocksDB` for existing types/functions, move modules in `DB` into `Rocks` subfolder

### DIFF
--- a/src/Dscp/CLI/Common.hs
+++ b/src/Dscp/CLI/Common.hs
@@ -5,6 +5,7 @@
 module Dscp.CLI.Common
        ( logParamsParser
        , dbPathParser
+       , sqliteDbPathParser
        , versionOption
        ) where
 
@@ -38,6 +39,13 @@ dbPathParser = strOption $
     metavar "FILEPATH" <>
     help "Path to database directory for witness node." <>
     value "witness-db"
+
+sqliteDbPathParser :: Parser FilePath
+sqliteDbPathParser = strOption $
+    long "db-path" <>
+    metavar "FILEPATH" <>
+    help "Path to database directory for educator's private data." <>
+    value "educator-db"
 
 versionOption :: Parser (a -> a)
 versionOption = infoOption ("disciplina-" <> (showVersion version)) $
--- a/src/Dscp/DB/SQLite/Class.hs
+++ b/src/Dscp/DB/SQLite/Class.hs
@@ -1,0 +1,22 @@
+module Dscp.DB.SQLite.Class where
+
+import Universum
+
+import Database.SQLite.Simple (FromRow, Query, ToRow)
+
+-- There are more functions in that library, feel free
+-- to add them to typeclass below if there are needed.
+
+-- | Full interface to SQLite DB.
+class Monad m => MonadSQLiteDB m where
+    -- | Make a simple @INSERT@ or other SQL query.
+    query :: (FromRow row, ToRow params) => Query -> params -> m [row]
+
+    -- | Make an @INSERT@ SQL query, feeding result to given function on per-row
+    -- basis.
+    queryStreamed
+        :: (FromRow row, ToRow params)
+        => Query -> params -> a -> (a -> row -> m a) -> m a
+
+    -- | Perform a simple SQL query which does not return any result.
+    execute :: ToRow q => Query -> q -> m ()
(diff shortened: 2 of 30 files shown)


Heimdell:
However, don't merge before @flyingleafe checks.

Martoon-00:
Glad to hear that from the one who writes one of the most readable and simple codes I've ever seen in haskell :smiley_cat:

src/Dscp/DB/SQLite/Functions.hs
```diff
@@ -0,0 +1,51 @@
...
+
+import Dscp.DB.SQLite.Class (MonadSQLiteDB (..), MonadSQLiteDBRead (..))
+import Dscp.DB.SQLite.Types (MonadSQLiteCtx, SQLiteDB (..), SQLiteDBLocation (..),
+                             SQLiteParams (..))
+
+-----------------------------------------------------------
+-- Opening/closing
+-----------------------------------------------------------
+
+newtype InvalidSQLitePathException = InvalidSQLitePathException FilePath
```
volhovm: I guess exceptions structure in our sql structure should be more elaborate -- a newtype per exception is a doubtful approach.

src/Dscp/DB/SQLite/Types.hs
```diff
@@ -0,0 +1,30 @@
...
+
+import Universum
+
+import Database.SQLite.Simple (Connection)
+import Ether.Internal (HasLens)
+
+-- | Set of constraints necessary to operate on SQLite database.
+type MonadSQLiteCtx ctx m =
+    ( MonadReader ctx m
+    , HasLens SQLiteDB ctx SQLiteDB
```
volhovm: `HasLens'`

src/Dscp/DB/SQLite/Types.hs
```diff
@@ -0,0 +1,30 @@
...
+       , SQLiteParams (..)
+       ) where
+
+import Universum
+
+import Database.SQLite.Simple (Connection)
+import Ether.Internal (HasLens)
+
+-- | Set of constraints necessary to operate on SQLite database.
+type MonadSQLiteCtx ctx m =
```
volhovm: `Ctx` postfix is unusual.

src/Dscp/DB/SQLite/Types.hs
```diff
@@ -0,0 +1,30 @@
...
+    ( MonadReader ctx m
+    , HasLens SQLiteDB ctx SQLiteDB
+    , MonadIO m
+    , Monad m
```
volhovm: `MonadIO` implies `Monad`, as well as `MonadReader`.

src/Dscp/DB/SQLite/Class.hs
```diff
@@ -0,0 +1,16 @@
...
+
+-- | Read-only interface to SQLite DB.
+class Monad m => MonadSQLiteDBRead m where
+    readMemes :: m ()
+    readMemes = pass
+
+-- | Full interface to SQLite DB.
+class MonadSQLiteDBRead m => MonadSQLiteDB m where
+    addProveOfCatsAwesomeness :: () -> m ()
+    addProveOfCatsAwesomeness () = return ()
```
flyingleafe: Well, why not already place here methods like `query` and `update`, with signatures like in similar methods in `sqlite-simple`? It's possible that we'll only need only them anyway.

src/Dscp/DB/SQLite/Types.hs
Martoon-00: Memno

src/Dscp/DB/SQLite/Types.hs
Martoon-00: I'll replace all occurrences later, currently it seems to be accessible only from your custom lootbox commit

Martoon-00:
Updates:
* Merged `MonadSQLRead` and `MonadSQL` typeclasses - SQLite does not provide easy way to distinguish read-only operations.
* Provided appropriate methods in `MonadSQL`
* Moved `RIO` monad to separate module - so that we can use it in e.g. DB modules freely
* Some other small fixes

src/Dscp/CLI/Common.hs
```diff
@@ -39,6 +40,13 @@ dbPathParser = strOption $
     help "Path to database directory for witness node." <>
     value "witness-db"
 
+sqliteDbPathParser :: Parser FilePath
+sqliteDbPathParser = strOption $
+    long "db-path" <>
```
volhovm: The key is the same as for `dbPathParser` by the way, please don't copypaste that hard.

PR #36, opened 2018-06-20
[DSCP-112] Swagger spec for Student API

flyingleafe:
### How to review
1. Copy raw link to file `educator.yaml` or `student.yaml` from this PR
2. Go to [editor.swagger.io](https://editor.swagger.io/)
3. `File > Import URL`
4. Paste the copied link
5. Enjoy the API spec represented with a nice Swagger GUI
6. Stop enjoying and diss the spec instead

### General comments about this spec

1. Student and Educator APIs are meant to be separate and be served on two different ports, so there are two separate Swagger spec files. This is done because there are completely different authentication requirements for endpoints used by Educator and Student, and implementing them as one monolithic API really seems to be very cumbersome and error-prone. Separating the API into two completely distinct parts, where each part uses single authentication systems across its endpoints, seems to be a much cleaner solution.
2. You might find URLs too long and repetitive: e. g. why have `GET /students/{studentAddr}/grades` and `GET /grades` as two different methods? Why not instead have only `GET /grades` and a bunch of filter query parameters for it? I had the following reasoning:
    - If we'll have combined API for Educator and students, I'd like to make sure that in order to determine if a user has rights to make a request it's enough to look at method and URL path. If we'll have combined API and fully filter-based structure, a student `X` will have access to `GET /grades?student=X`, but not to `GET /grades`, which seems super counter-intuitive and ugly.
    - There are some possible query parameters which fit query string much more, such as filtering by timestamp/grade boundaries, sorting parameters, pagination, etc. I'd like to separate such parameters from parameters like `courseId`.

### What does this spec cover

- [ ] Student API
  - [x] Methods structure
  - [x] JSON schemas for API datatypes
  - [ ] Authentication
  - [ ] Response sorting
  - [x] Response filtering
  - [ ] Response pagination
  - [ ] Versioning
    - [x] (at least partially, minor versions in headers are not yet defined)

- [ ] Educator API
  - [ ] Methods structure
    - [x] (at least partially)
  - [ ] JSON schemas for API datatypes
    - [x] (at least partially)
  - [x] Authentication (no auth, serves only locally)
  - [ ] Response sorting
  - [ ] Response filtering
  - [ ] Response pagination
  - [ ] Versioning
    - [x] (at least partially, minor versions in headers are not yet defined)


flyingleafe:
I tend to think that we really should have two separate APIs for Educator integration and for Student, after thinking a bit more about authentication methods. It's more or less clear how to make a proper™ authentication in case of separate APIs:
- Educator API simply serves only localhost.
- Student API serves openly, and uses OAuth2 (or custom public-key based authorization mechanism) for student's authorization.

Given that, I think that we should deem the API which is currently described here a part of Educator API, because it has these `/student/{studentAddr}` URIs, whereas separate Student API would simply have URIs like `/courses`, `/grades`, `/submissions`, etc.

lambda76:
Some thoughts, 
1. How is API versioning handled? Should we have prefix all endpoints with /api/v1/ ?
2. I prefer singular in naming, i.e /student/{studentAddr}/course/{courseId} instead of /students/{studentAddr}/courses/{courseId}
3. How should hashes and addresses be encoded? Should we encode it with BaseX something? 
4. What is the GET /grades API endpoint for? Isn't everything related to grades so this basically exports all data?
5. Some endpoints under submission section start with /students/, why are they not under student section?

specs/disciplina/educator/api/swagger.yaml
```diff
@@ -0,0 +1,650 @@
...
+    properties:
+      hash:
+        $ref: "#/definitions/Hash"
+      submission_hash:
+        $ref: "#/definitions/Hash"
+      grade:
+        type: "integer"
+        format: "int32"
+      timestamp:
+        $ref: "#/definitions/Time"
```
Martoon-00: Perhaps you wished to mark it as `readOnly`.

specs/disciplina/educator/api/swagger.yaml
```diff
@@ -0,0 +1,650 @@
...
+            $ref: "#/definitions/Grade"
+      
+  
+securityDefinitions:
+  petstore_auth:
+    type: "oauth2"
+    authorizationUrl: "http://petstore.swagger.io/oauth/dialog"
+    flow: "implicit"
+    scopes:
+      write:pets: "modify pets in your account"
```
Martoon-00: The truth about our relation to users has been revealed 😏

specs/disciplina/educator/api/swagger.yaml
```diff
@@ -0,0 +1,650 @@
...
+      - "application/json"
+      parameters:
+      - $ref: "#/parameters/SubmissionHash"
+      responses:
+        200:
+          description: "successful operation"
+        404:
+          description: "Submission with given hash not found"
+  
+  /grades:
```
Heimdell: Do we need such an endpoint? It is practically a database dump.

flyingleafe: Not really, it only lists _grades_, and there's lots of other stuff in our database. This endpoint basically provides a way for looking at the whole private chain, like a block explorer for it.

flyingleafe:
@lambda76 , sorry for taking too long to reply.
In most my answers to your question, I'll refer to this article: https://www.vinaysahni.com/best-practices-for-a-pragmatic-restful-api , which I find quite reasonable.
1. We'll use API prefix for major version and custom headers for minor/patch version (see https://www.vinaysahni.com/best-practices-for-a-pragmatic-restful-api#versioning). This way it's easiest to introduce a new API version with breaking changes while still keeping old one functional.
2. Using plurals with object's IDs is more consistent, because you usually use plural for describing a whole resource and getting lists of objects (like `GET /students`), so it's more consistent if in order to work with a particular student we just need to append their address to resources' name (`GET /students/<addr>`, not `GET /student/<addr>`)
3. Addresses will probably use `base58` encoding (similarly to Bitcoin, Cardano and tons of others), because it's more readable and doesn't contain easy-to-mismatch pairs of characters; simple hashes and other binary data will use `base64`.
4. Not _all_ the data, there are lots of other data (courses descriptions, new assignments, ungraded submissions) which isn't yet connected with grades.
5. Look closely, they are under both `students` and `submissions` sections (because these sections is how Swagger renders _tags_. One endpoint can have many tags.)

flyingleafe:
Added a verbal spec for interaction between Student and Educator, which describes different options of workflow: https://github.com/DisciplinaOU/disciplina/blob/2314b0371cd18dd9072633b514eff1a321dfeb84/specs/disciplina/educator/api/spec.org

Please, read, diss and point out vague and ill-defined moments.

Heimdell (approved):
The `spec.org` part looks fine.

specs/disciplina/educator/api/spec.org
```diff
@@ -0,0 +1,120 @@
...
+  These two modes can be _combined_: there can be a set of default course
+  assignments, and particular students also can receive additional ones during
+  the course.
+** Submissions
+   There are two modes of operations of submitting _submissions_ to assignments:
+   
+   1) *Offline submissions*. If a course is meant to be offline (e. g. it's a
+      classic university course with course assignments done on paper) and there
+      are no actual digital _content_ of submissions, then submissions in
+      Disciplina system become effectively fictional. This means that these
```
volhovm: Not exactly. They serve the role of protecting the student from the unhonest educator. Since every assignment must be signed by the student before a grade for the submission can be assigned, student can effectively stop the course at any point. Otherwise, educator can at any moment control grades for the future assignments which were not really made by the student.

volhovm (commented):
Corresponds to what we've discussed IRL.

specs/disciplina/educator/api/spec.org
```diff
@@ -0,0 +1,120 @@
...
+     assigned to a student at some moment during the course. Different students
+     can possibly have different final assignments, so we cannot automatically
+     assign these at the beginning of the course as well.
+
+  Also, there are two ways of completing the final assignment:
+  1) *Fictional/offline final submission*. Similarly to _offline course
+     submissions_, they are submitted automatically as soon as final assignment
+     is assigned.
+  2) *Online final submission*. Similarly to _online course submissions_, these
+     ones should refer to some actual submission contents via _contents hash_
```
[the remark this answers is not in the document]
lambda76: `s/submission contents via _contents hash_/submission content via _content hash_'`

lambda76 (changes requested):
These are my suggestions

specs/disciplina/educator/api/spec.org
```diff
@@ -0,0 +1,120 @@
...
+     is defined outside Disciplina system (e. g. in the integrated CRM).
+     Therefore, as soon as *Student* enrolls to a course, they can submit
+     submissions for these default assignments.
+  2) *On-demand assignments*. *Educator* assigns assignments to students as the
+     course goes. In the beginning of the course, there might be no course
+     assignments assigned to a *Student*. A *Student* should poll Student API to
+     check for updates regarding course assignments.
+     
+  These two modes can be _combined_: there can be a set of default course
+  assignments, and particular students also can receive additional ones during
```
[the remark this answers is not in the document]
flyingleafe: I wrote "particular" just to emphasize that not _every_ student might receive additional assignments. I agree though, that it can be dropped.

specs/disciplina/educator/api/spec.org
```diff
@@ -0,0 +1,120 @@
...
+   2) *Online submissions*. If a course is meant to be online (e. g. it's a
+      MOOC) and every submission is digital, then submissions in Disciplina
+      system should refer to actual contents of submissions via _contents hash_.
+      These _online_ submissions are submitted as the *Student* prepares them. 
+
+   These two modes can be _combined_: a course can have both _online_ and
+   _offline_ assignments, and submissions for these assignments are sent,
+   accordingly, manually or automatically.
+
+* Finishing the course
```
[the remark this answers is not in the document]
flyingleafe: Why "studying _the_ course", but "completing _a_ course"?

lambda76 (approved):

flyingleafe:
Student API is sort of ready to diss, @lambda76 @Heimdell @Martoon-00 @volhovm  
https://github.com/DisciplinaOU/disciplina/blob/62cc4bcb8f16e6d6da3c9b1a460a40d12574977e/specs/disciplina/educator/api/student.yaml
It turned out to be quite small, but it seems that it's all we need to implement the demo app. I think we can live without a pagination for now. The only thing which left which is _necessary_ is authentication mechanism.

Also, this PR seems to be too long and large. I propose to merge it as soon as basic Student API spec is good enough, and do other things (Educator API, authentication, pagination, sorting, etc.) in separate PRs/issues.

specs/disciplina/educator/api/student.yaml
```diff
@@ -0,0 +1,414 @@
...
+      tags:
+        - courses
+      summary: Get Educator\'s courses
+      description: Gets a list of Educator\'s courses, both enrolled and available.
+      operationId: getCourses
+      parameters:
+        - in: query
+          name: enrolled
+          required: false
+          description: Is a student already enrolled in a course?
```
volhovm: Sorry, what does it mean? I don't understand a semantics.

specs/disciplina/educator/api/student.yaml
```diff
@@ -0,0 +1,414 @@
...
+        - application/json
+      consumes:
+        - application/json
+      parameters:
+        - in: body
+          name: submission
+          description: A submission
+          required: true
+          schema:
+            $ref: '#/definitions/Submission'
```
volhovm: `Submission` has "grade" field. I doubt that the student must send any grades to the educator. He should send the "ungraded submission" instead.

EDIT: "grade" field of the submission is not required, I see. The automatic example in swagger is just bad.

specs/disciplina/educator/api/student.yaml
```diff
@@ -0,0 +1,414 @@
...
+      operationId: deleteSubmission
+      produces:
+        - application/json
+      parameters:
+        - $ref: '#/parameters/SubmissionHashPath'
+      responses:
+        '200':
+          description: successful operation
+        '403':
+          description: Cannot delete a graded submission
```
volhovm: Maybe 403 should return the string with the error code so that we can process all 403 responses in a similar way?

specs/disciplina/educator/api/student.yaml
```diff
@@ -0,0 +1,414 @@
...
+    properties:
+      id:
+        $ref: '#/definitions/CourseId'
+      desc:
+        type: string
+      subjects:
+        type: array
+        items:
+          $ref: '#/definitions/Subject'
+      is_enrolled:
```
volhovm: Why this field is not required?

specs/disciplina/educator/api/student.yaml
```diff
@@ -0,0 +1,414 @@
...
+      - grade
+      - timestamp
+      - has_proof
+    properties:
+      grade:
+        type: integer
+        format: int32
+      timestamp:
+        $ref: '#/definitions/Time'
+      has_proof:
```
volhovm: What does this field mean?

specs/disciplina/educator/api/student.yaml
```diff
@@ -0,0 +1,414 @@
...
+      assignment:
+        $ref: '#/definitions/Assignment'
+      witness:
+        type: string
+        format: binary
+        
+  Transaction:
+    type: object
+    required:
+      - hash
```
volhovm: Is it a hash of block or a hash of transaction?

volhovm (changes requested):

specs/disciplina/educator/api/student.yaml
volhovm: Makes sense, it should be written like this.

specs/disciplina/educator/api/student.yaml
volhovm: Ok, so technically it's `was_published`.

volhovm (approved):

specs/disciplina/educator/api/student.yaml
flyingleafe: @volhovm That would be incorrect, because the _grade_ isn't _published_. `has_proof` is way less confusing, I believe.

PR #37, opened 2018-06-24
[DSCP-117] validation exceptions structure

lambda76:
- [x] Validation datatype for private block validation failures

### DIFF
--- a/src/Dscp/Educator/Validate.hs
+++ b/src/Dscp/Educator/Validate.hs
@@ -53,10 +53,13 @@ instance Buildable BlockValidationFailure where
     build (MerkleSignatureMismatch {..}) =
       "Merkle tree root signature mismatch. Expected " `mappend` show bvfExpectedSig
       `mappend` " got " `mappend` show bvfActualMerkleSig
-    build (SubmissionInvalid (SubmissionPublicKeyMismatch {..})) =
+    build (SubmissionInvalid x) = build x
+
+instance Buildable SubmissionValidationFailure where
+    build (SubmissionPublicKeyMismatch {..}) =
       "Submission public key address mismatch. Expected " `mappend` show svfExpectedPubKey `mappend`
       " got " `mappend` show svfActualPubKey
-    build (SubmissionInvalid (SubmissionSignatureMismatch {..})) =
+    build (SubmissionSignatureMismatch {..}) =
       "Submission signature mismatch. Submission data " `mappend` show svfSubmissionHash
       `mappend` " with signature " `mappend` show svfSubmissionSig
       `mappend` " does not correspond to public key " `mappend` show svfSubmissionSigKey
--- a/src/Dscp/Educator/Validate.hs
+++ b/src/Dscp/Educator/Validate.hs
@@ -2,7 +2,8 @@
 -- | Validation of private block
 
 module Dscp.Educator.Validate
-       ( ValidationFailure (..)
+       ( BlockValidationFailure (..)
+       , SubmissionValidationFailure (..)
        , validatePrivateBlk
        ) where
 
@@ -13,7 +14,7 @@ import Data.Text.Buildable (build)
 import Dscp.Core (Address (..), SignedSubmission (..), Submission (..),
                   SubmissionSig, SubmissionWitness (..))
 import Dscp.Crypto (Hash, PublicKey, MerkleSignature, fromFoldable, getMerkleRoot, hash, verify)
-import Dscp.Educator.Block (PrivateBlock (..), PrivateBlockHeader (..), pbBody, pbbTxs)
+import Dscp.Educator.Block (PrivateBlock (..), PrivateBlockHeader (..), _pbBody, _pbbTxs)
 import Dscp.Educator.Serialise ()
 import Dscp.Educator.Txs (PrivateTx (..))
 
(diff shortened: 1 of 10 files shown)


src/Dscp/Educator/Validate.hs
```diff
@@ -2,34 +2,67 @@
...
+import qualified Text.Show
+import qualified Data.Text.Buildable ()
+
+
+-- | Validation failures
+data ValidationFailure
+    = MerkleSignatureMismatch !(MerkleSignature PrivateTx) !(MerkleSignature PrivateTx)
+      -- ^ Header merkle tree root does not match
+      -- body merkle tree root
+    | TxPublicKeyMismatch !(Hash PublicKey) !(Hash PublicKey)
```
flyingleafe: It's better to semantically separate _block header_ validation errors and _transaction_ validation errors. Better to do like this:
```
data BlockValidationFailure
    = MerkleSignatureMismatch ....
    | TxInvalid !TxValidationFailure
    ...

data TxValidationFailure
    = TxPublicKeyMismatch ...
    | TxSignatureMismatch ...
```
Also, it's better to rename `TxValidationFailure` to `SubmissionValidationFailure`, because we're actually checking submission signatures here. These both will allow to re-use `SubmissionValidationFailure` in API logic of validation incoming submissions.

flyingleafe: Also, please, use record syntax with proper field names, like:
```
    ...
    = MerkleSignatureMismatch 
        { bvfExpectedMerkleSig :: !(MerkleSignature PrivateTx)
        , bvfActualMerkleSig   :: !(MerkleSignature PrivateTx) }
    ...
```

src/Dscp/Educator/Validate.hs
```diff
@@ -2,34 +2,67 @@
...
+      -- ^ Witness public key do not match transaction address
+    | TxSignatureMismatch !PublicKey !SubmissionSig !(Hash Submission)
+      -- ^ Witness signature do not match transaction submission signature
+    deriving (Eq)
+
+instance Show ValidationFailure where
+    show e = toString . pretty $ e
```
flyingleafe: Eta-reduction: `show = toString . pretty`.
(You can use `hlint` to see such hints automatically)

src/Dscp/Educator/Validate.hs
```diff
@@ -13,63 +14,89 @@ import Data.Text.Buildable (build)
...
-      "Tx public key address mismatch. Expected " `mappend` show expected `mappend`
-      " got " `mappend` show got
-    build (TxSignatureMismatch key sig h) =
-      "Tx signature mismatch. Submission data " `mappend` show h `mappend` " with signature "
-      `mappend` show sig `mappend` " does not correspond to public key " `mappend` show key
+instance Buildable BlockValidationFailure where
+    build (MerkleSignatureMismatch {..}) =
+      "Merkle tree root signature mismatch. Expected " `mappend` show bvfExpectedSig
+      `mappend` " got " `mappend` show bvfActualMerkleSig
+    build (SubmissionInvalid (SubmissionPublicKeyMismatch {..})) =
```
flyingleafe: There should be a separate `Buildable` instance for `SubmissionValidationFailure` instead, and case for `SubmissionInvalid` should use this instance.

src/Dscp/Educator/Validate.hs
```diff
@@ -2,41 +2,104 @@
...
+    build (SubmissionInvalid x) = build x
+
+instance Buildable SubmissionValidationFailure where
+    build (SubmissionPublicKeyMismatch {..}) =
+      "Submission public key address mismatch. Expected " `mappend` show svfExpectedPubKey `mappend`
+      " got " `mappend` show svfActualPubKey
+    build (SubmissionSignatureMismatch {..}) =
+      "Submission signature mismatch. Submission data " `mappend` show svfSubmissionHash
+      `mappend` " with signature " `mappend` show svfSubmissionSig
+      `mappend` " does not correspond to public key " `mappend` show svfSubmissionSigKey
```
Heimdell: You can `mconcat [..., ..., ...]` or use `[qc|foo {svfSubmissionHash} bar]` after importing `Text.InterpolatedString.Perl6` from `interpolatedstring-perl6` for brevity. Latter one does not require `show` here and there.

src/Dscp/Educator/Validate.hs
```diff
@@ -2,41 +2,104 @@
...
-              txAddrHash = addrHash (sStudentId submission)
+              subAddrHash = addrHash (sStudentId submission)
               witnessSigValid = verify witnessKey (hash submission) witnessSign
-              witnessKeyValid = hash witnessKey == txAddrHash
-          in [(witnessKeyValid, "Tx public key missmatch")
-             ,(witnessSigValid, "Tx signature not valid")]
+              witnessKeyValid = hash witnessKey == subAddrHash
+          in [(witnessKeyValid,
+               SubmissionInvalid { bvfSubmissionFailure =
+                                     SubmissionPublicKeyMismatch { svfExpectedPubKey = subAddrHash
```
Heimdell: Honestly, I recommend you the following indent schema:
```haskell
foo = bar
  where
    barbar = Bar
        { qux = 1
        , other = ...
        }
```
It is easier to read, especially in github diff form, and it is more compact and predictable.

src/Dscp/Educator/Validate.hs
```diff
@@ -2,41 +2,104 @@
...
+  where validateHeader (PrivateBlockHeader {..}) merkleRoot =
+          [(_pbhBodyProof == merkleRoot,
+            MerkleSignatureMismatch { bvfExpectedSig     = _pbhBodyProof
+                                    , bvfActualMerkleSig =  merkleRoot
+                                    })]
+
+        validateSub (PrivateTx {..}) =
           let submission = ssSubmission _ptxSignedSubmission
               witnessKey = _swKey (ssWitness _ptxSignedSubmission)
               witnessSign = _swSig (ssWitness _ptxSignedSubmission)
```
Heimdell: The `_swKey` looks like a field from datatype which has lens generated.
If it is, I recommend you using the lenses, not the field direclty, like that:
```haskell
ssWitness^.ptxSignedSubmission.swSig
```
That way you don't have to import things, which names start from underscope, and are able to call `use[s]`, `zoom` and other powerful combinators from `Control.Lens`.

tests/Test/Dscp/Educator/Validate.hs
```diff
@@ -62,10 +56,33 @@ spec_ValidateBlock = describe "Validate private block" $ do
...
+                 (SubmissionSignatureMismatch { svfSubmissionHash = hashTxSub tx2
+                                              , svfSubmissionSig = getTxSig tx2
+                                              , svfSubmissionSigKey = getTxKey tx2
+                                              })]
+        validatePrivateBlk (mkBlock [tx3]) `shouldBe`
+          Left [SubmissionInvalid
+                 (SubmissionPublicKeyMismatch { svfExpectedPubKey = hash studentAPubKey
+                                              , svfActualPubKey = hash (getTxKey tx3)
+                                              })
+               ,SubmissionInvalid
```
Heimdell: Not an issue, but, please, put spaces after `,`

Heimdell: You can write lists like that:
```haskell
[ 1
, 2
]
```

src/Dscp/Educator/Validate.hs
lambda76: Actually, these were not lenses. I took the liberty and made lenses, hope it was ok to do so

PR #43, opened 2018-06-30
[DSCP-119] Utility functions for AES encryption.

flyingleafe:
Simple utility functions and datatypes which greatly simplify the usage of `cryptonite` encryption primitives. Currently only for encrypting/decrypting a secret key with a passphrase.

### DIFF
--- a/src/Dscp/Crypto/Encrypt.hs
+++ b/src/Dscp/Crypto/Encrypt.hs
@@ -1,0 +1,116 @@
+
+-- | Utilities for encrypting/decrypting byte arrays
+
+module Dscp.Crypto.Encrypt
+       ( -- * Passphrases
+         PassPhrase
+       , getPassPhrase
+       , mkPassPhrase
+       , minPassPhraseLength
+
+         -- * Encrypted bytearray
+       , Encrypted
+       , eAuthTag
+       , eCiphertext
+
+         -- * Utility functions
+       , encrypt
+       , decrypt
+       ) where
+
+import Crypto.Cipher.AES (AES256)
+import Crypto.Cipher.Types (AEAD, AEADMode (AEAD_GCM), AuthTag, BlockCipher (aeadInit),
+                            Cipher (cipherInit), IV, aeadSimpleDecrypt, aeadSimpleEncrypt, nullIV)
+import Crypto.Error (onCryptoFailure)
+import Data.ByteArray (ByteArray, ByteArrayAccess)
+import qualified Data.ByteArray as BA
+import Fmt ((+|), (|+))
+
+import Dscp.Crypto.Impl (hash)
+
+-------------------------------------------------------------
+-- Passphrases
+-------------------------------------------------------------
+
+newtype PassPhrase = PassPhrase
+    { getPassPhrase :: ByteString
+    } deriving (Eq, Ord, Show, Monoid, ByteArray, ByteArrayAccess)
+
+-- | Minimal passphrase length. Should be enough.
+minPassPhraseLength :: Int
+minPassPhraseLength = 8
+
+mkPassPhrase :: ByteString -> Either Text PassPhrase
+mkPassPhrase bs
+    | length bs < minPassPhraseLength = Left shortPassErr
+    | otherwise = Right $ PassPhrase bs
+  where
+    shortPassErr = "Passphrase is too short, minimal length is "+|minPassPhraseLength|+" chars."
+
+-------------------------------------------------------------
+-- Encryption/decryption
+-------------------------------------------------------------
+
+-- | Datatype which combines ciphertext with cipher authentication tag.
+data Encrypted ba = Encrypted
+    { eAuthTag    :: !AuthTag
+    , eCiphertext :: !ba
+    } deriving (Eq)
+
+-- | Authentication tag length. Number 16 is the same as in the
+-- `Crypto.MAC.Poly1305.Auth` tag datatype.
+-- TODO: probably it can be made shorter without any problems?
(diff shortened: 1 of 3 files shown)


tests/Test/Dscp/Crypto/Encrypt.hs
```diff
@@ -0,0 +1,21 @@
...
+spec_encryption :: Spec
+spec_encryption = describe "AES encryption functions" $ do
+    it "should stay the same after encryption and decryption" $ property $
+        \(pp, bs :: ByteString) -> decrypt pp (encrypt pp bs) === Just bs
+    it "should yield different ciphertexts for different plaintexts" $ property $
+        \(pp, bs :: ByteString, bs' :: ByteString) -> bs /= bs' ==> encrypt pp bs /= encrypt pp bs'
+    it "should encrypt differently for different passphrases" $ property $
+        \(pp, pp', bs :: ByteString) -> pp /= pp' ==> encrypt pp bs /= encrypt pp' bs
+    it "should fail when trying to decrypt a ciphertext with wrong passphrase" $ property $
+        \(pp, pp', bs :: ByteString) -> decrypt pp (encrypt pp' bs) === Nothing
```
Heimdell: I would add `pp /= pp' ==>` here, or it could suddenly fail.

Heimdell (approved):
Fund only one minor thing.

Everything else looks very good.

src/Dscp/Crypto/Encrypt.hs
```diff
@@ -0,0 +1,116 @@
...
+-- | Encrypt given 'ByteArray' with AES.
+encrypt :: ByteArray ba => PassPhrase -> ba -> Encrypted ba
+encrypt pp plaintext =
+    let aead = prepareAEAD pp
+    in uncurry Encrypted $
+       aeadSimpleEncrypt aead authHeader plaintext authTagLength
+
+-- | Decrypt given 'Encrypted' datatype or fail, if passphrase
+-- doesn't match.
+decrypt :: ByteArray ba => PassPhrase -> Encrypted ba -> Maybe ba
```
Martoon-00: I suppose it's better to have `Either Text ba` here.
In cardano I saw several times how people were lazy to check why such function may fail (even when reason was only one - password mismatch) and wrote some general non-speaking error text at call site.

src/Dscp/Crypto/Encrypt.hs
```diff
@@ -0,0 +1,116 @@
...
+
+import Dscp.Crypto.Impl (hash)
+
+-------------------------------------------------------------
+-- Passphrases
+-------------------------------------------------------------
+
+newtype PassPhrase = PassPhrase
+    { getPassPhrase :: ByteString
+    } deriving (Eq, Ord, Show, Monoid, ByteArray, ByteArrayAccess)
```
Martoon-00: Don't we want blind `Show` instance?

Martoon-00 (changes requested):
LGTM, although I do not osilau much into cryptostuff.
Please address some minor comments

src/Dscp/Crypto/Encrypt.hs
flyingleafe: Fair enough. I'll actually make blind `Buildable` instance too.

src/Dscp/Crypto/Encrypt.hs
flyingleafe: In cases when there's indeed only one possible reason for failure, `Maybe` is sort of semantically more valid (e. g. the number of possible representatives of return type are the same as the number of possible outcomes of evaluating the function).
However, we can use the best of both worlds and define a custom exception type here (with one constructor and corresponding `Buildable` instance).

Martoon-00 (approved):

src/Dscp/Crypto/Encrypt.hs
```diff
@@ -0,0 +1,170 @@
...
+-- | Chosen cipher. We'll stick with AES256 for now.
+-- TODO: are there better alternatives?
+type CipherType = AES256
+
+-- | Authentication header used for our encryption (empty one).
+authHeader :: ByteString
+authHeader = ""
+
+-- | AEAD mode used for our encryption scheme.
+-- We use GCM mode, because it's the most modern and performant one
```
volhovm: And also because it provides authentication, otherwise we'd had to use some other MAC.

src/Dscp/Crypto/Encrypt.hs
```diff
@@ -0,0 +1,170 @@
...
+--      AEAD_CCM: https://en.wikipedia.org/wiki/CCM_mode
+--      AEAD_CWC: https://en.wikipedia.org/wiki/CWC_mode
+--      AEAD_EAX: https://en.wikipedia.org/wiki/EAX_mode
+--      AEAD_GCM: https://en.wikipedia.org/wiki/Galois/Counter_Mode
+aeadMode :: AEADMode
+aeadMode = AEAD_GCM
+
+-- | Initial vector for chosen cipher.
+initIV :: IV CipherType
+initIV = nullIV
```
volhovm: Using fixed initial vector is insecure, it should be generated randomly and appended to the beginning of the message. Google for more details.

volhovm: @flyingleafe this is ~important, please consider my comment. The issue may not be critical, but it's not how one should deal with crypto. Let's append IV to the message.

flyingleafe: @volhovm I hear you. See DSCP-130 (going to do it tomorrow).

PR #44, opened 2018-06-30
[DSCP-120, DSCP-121, DSCP-122] Add educator secret key storage 

Martoon-00:
Added educator secret key storage.
Among features mentioned in https://issues.serokell.io/issue/DSCP-118, following stuff is handled:
* Keyfile is created with mode 600, without CPP this time
* Secure randomness when generating secret key (copy-pasted from you known which project)
* Aeson roundtrip tests (do we need some other tests?)

What remains:
- [x] Ask everybody to install `libssl-dev` as soon as OpenSSL is used for secure randomness
- [x] Discuss where do we want to declare `Arbitrary` instances. For now I place them next to corresponding types


lambda76:
Sorry, I missed your ping martoon-00. The work I've done on dscp-120 is probably obsolete now.

Martoon-00:
Ready for diss

src/Dscp/Crypto/Signing/Class.hs
```diff
@@ -56,6 +91,36 @@ deriving instance ByteArrayAccess (SK ss) =>
...
+-- They are used for desirisalisation.
+deriving instance FromByteArray (PK ss) =>
+    FromByteArray (AbstractPK ss)
+deriving instance FromByteArray (SK ss) =>
+    FromByteArray (AbstractSK ss)
+deriving instance FromByteArray (Sig ss) =>
+    FromByteArray (AbstractSig ss a)
+
+-- | Generate a secret key.
+keyGen
```
lambda76: For ed25519 which we use as Signature scheme I believe, there is already a `generateSecretKey` function, http://hackage.haskell.org/package/cryptonite-0.25/docs/Crypto-PubKey-Ed25519.html. Can we make use of that one instead of creating our own here? For example having a `genSecretKey` function in our `SignatureScheme` class.

Martoon-00: Nice point, it seem to do exactly the same thing.

src/Dscp/Crypto/Signing/Class.hs
```diff
@@ -35,18 +54,34 @@ newtype AbstractPK ss = AbstractPK (PK ss)
...
+
+instance (SignatureScheme ss, FromByteArray (AbstractSK ss)) =>
+         Arbitrary (AbstractSK ss) where
+    arbitrary =
+        leftToPanicWith "arbitrary abstractSK" . fromByteArray . BS.pack <$>
+        vector (secretKeyLength $ schemeLengths @ss)
 
--- | Wrapper for a secret key. 'Eq', 'Ord' and 'Show' instances
--- are not derived for security reasons.
+-- | Wrapper for a secret key. 'Show' instance is not derived for
```
lambda76: For `SignatureScheme CryptoEd25519` `Show` is safe, will show as `<scrubbed-bytes>`, http://hackage.haskell.org/package/memory-0.14.14/docs/src/Data-ByteArray-ScrubbedBytes.html#line-44

Martoon-00: I guess I'll remove comment then, having specific "secret" here seems more pleasant though.

src/Dscp/Crypto/Aeson.hs
```diff
@@ -0,0 +1,20 @@
...
+
+import Data.Aeson (FromJSON (..), ToJSON (..), object, withObject, (.:), (.=))
+
+import Dscp.Crypto.Encrypt (Encrypted (..))
+import Dscp.Crypto.Instances ()
+import Dscp.Util.Aeson (AsByteString (..), Base64)
+
+instance ToJSON a => ToJSON (Encrypted a) where
+    toJSON Encrypted{..} = object
+        [ "encrypted" .= eCiphertext
```
flyingleafe: Ciphertext is also going to be an arbitrary sequence of bytes, it also should be represented as `base64`, I assume? :thinking: 
Actually, I'd better provide a `Serialise` instance for `Encrypted` and store it in JSON in serialized way as `base64`. This way we'll be able to hide `Encrypted` constructor from exports, preventing it to be used unsafely.

src/Dscp/Crypto/ByteArray.hs
```diff
@@ -1,3 +1,6 @@
+{-# LANGUAGE DataKinds                 #-}
+{-# LANGUAGE ExistentialQuantification #-}
+{-# LANGUAGE TypeInType                #-}
```
flyingleafe: Why did you need these extensions in this file if there's no related changes in the code? :thinking:

src/Dscp/Crypto/Encrypt.hs
```diff
@@ -22,16 +21,21 @@ module Dscp.Crypto.Encrypt
...
+import Crypto.Cipher.Types (AEAD, AEADMode (AEAD_GCM), AuthTag (..), BlockCipher (aeadInit),
                             Cipher (cipherInit), IV, aeadSimpleDecrypt, aeadSimpleEncrypt, nullIV)
 import Crypto.Error (onCryptoFailure)
 import Data.ByteArray (ByteArray, ByteArrayAccess)
 import qualified Data.ByteArray as BA
+import qualified Data.ByteString as BS
 import Data.Text.Buildable (build)
 import Fmt ((+|), (|+))
+import Test.QuickCheck (Arbitrary (..), suchThatMap, vector)
+import Test.QuickCheck.Instances ()
```
flyingleafe: No, please, let's keep all `Arbitrary` instances in `tests` (or at least in separate `*.Arbitrary` modules).
I'd really prefer to keep them only in tests unless really necessary otherwise (like in case with using these instances in benchmarks).

src/Dscp/Crypto/Encrypt.hs
```diff
@@ -96,7 +103,13 @@ mkPassPhrase bs
...
     { eAuthTag    :: !AuthTag
     , eCiphertext :: !ba
-    } deriving (Eq)
+    } deriving (Eq, Functor)
+
+instance Buildable ba => Show (Encrypted ba) where
+    show = toString . pretty
+
+instance Buildable ba => Buildable (Encrypted ba) where
+    build Encrypted{..} = "encrypted "+|eCiphertext|+""
```
flyingleafe: Why not simply use `base64` representation of serialised `Encrypted` (as I suggested to do also with Aeson instance?)

src/Dscp/Crypto/Encrypt.hs
```diff
@@ -143,19 +156,37 @@ prepareAEAD (PassPhrase pp) =
...
-       aeadSimpleEncrypt aead authHeader plaintext authTagLength
+        -- TODO: Subject to optimization. For now using converion to bytestring
+        -- since @ba@ is not instance of 'ByteArray'.
+        plaintextBS = BA.convert plaintext
+        (authtag, ciphertextBS :: ByteString) =
+            aeadSimpleEncrypt aead authHeader plaintextBS authTagLength
+        ciphertext =
+            leftToPanicWith "encrypt: got malformed item: " $
+            fromByteArray ciphertextBS
+    in  Encrypted authtag ciphertext
```
flyingleafe: What is the motivation for these changes? :thinking: If it's only to allow for using `encrypt` directly on `SecretKey` datatype, then it doesn't seem like the most elegant way to perform it (if you don't want to think of fancy ways to avoid conversions, then you might've renamed existing functions to `encryptBA/decryptBA` and make `encrypt/decrypt` function a simple wrappers which perform back and forth conversion.

src/Dscp/Crypto/Signing/Class.hs
```diff
@@ -28,25 +35,42 @@ class SignatureScheme ss where
...
 
 -- | 'GeneralizedNewtypeDeriving' cannot into type families.
 deriving instance Eq (PK ss) => Eq (AbstractPK ss)
 deriving instance Ord (PK ss) => Ord (AbstractPK ss)
 deriving instance Show (PK ss) => Show (AbstractPK ss)
+deriving instance Monoid (PK ss) => Monoid (AbstractPK ss)
+
+instance (SignatureScheme ss, FromByteArray (AbstractSK ss)) =>
+         Arbitrary (AbstractSK ss) where
+    arbitrary = generator genSecretKey
```
flyingleafe: Again, I'd really like to keep `Arbitrary` instances in `tests` while possible.

src/Dscp/Crypto/Random.hs
```diff
@@ -0,0 +1,75 @@
+-- | Secure generation of random numbers and bytestrings.
```
flyingleafe: fresh smell of Cardano copypaste :peka:

src/Dscp/Educator/Secret/Real/Functions.hs
```diff
@@ -0,0 +1,104 @@
...
+---------------------------------------------------------------------
+-- Storage operations
+---------------------------------------------------------------------
+
+-- | Where keyfile would lie.
+storePath :: EducatorSecretParams -> AppDirectory -> FilePath
+storePath EducatorSecretParams{..} (AppDirectory appDir) =
+    fromMaybe defPath espPath
+  where
+    defPath = appDir </> "secret.key"
```
flyingleafe: Hey, it should be `educator.key` (because in app directory there also would be `witness.key`, which is probably going to be different, and probably something else).
Also, there should be an option to set up a _completely_ custom absolute  path to the key, and from what I see here I conclude that it's only possible to choose app _directory_, but not the file's _basename_.

src/Dscp/Educator/Secret/Real/Types.hs
```diff
@@ -0,0 +1,52 @@
...
+      -- ^ Password from encrypted secret key stored on disk.
+    }
+
+---------------------------------------------------------------------
+-- Instances
+---------------------------------------------------------------------
+
+-- | Intermediate form of 'EducatorSecret' for JSON serialization.
+data EducatorSecretJson = EducatorSecretJson
+    { esjEncSecretKey :: Encrypted SecretKey
```
flyingleafe: Missed bang-pattern. Also, why do you need this intermediate form anyway?

src/Dscp/Util/Aeson.hs
```diff
@@ -0,0 +1,72 @@
...
+    parseJSON = withObject "Versioned content" $ \o -> do
+        ver <- o .: "version"
+        unless (ver == theVersion) $
+            fail $ "Only "+||theVersion||+" version is supported"
+
+        content <- o .: "content"
+        return $ Versioned content
+
+instance Arbitrary a => Arbitrary (Versioned a) where
+    arbitrary = Versioned <$> arbitrary
```
flyingleafe: Same as in previous comments.

src/Dscp/Crypto/Aeson.hs
Martoon-00: Makes sense.

src/Dscp/Crypto/Encrypt.hs
Martoon-00: Yep, will do

src/Dscp/Crypto/Encrypt.hs
Martoon-00: I tried to do conversion in smarter way like adding a wrapper and `FromByteArray ba => ByteArray (PotentiallyMalformed ba)` for it, but it seemed too painful because `ByteArray` requires `Monad` and `ByteArrayAccess`. Would be happy to hear another way to workaround that.

src/Dscp/Educator/Secret/Real/Types.hs
Martoon-00: Secret key is kept encrypted in file, while in context we have unencrypted one, and it seemed improper to involve encryption into JSON serialization.

src/Dscp/Crypto/ByteArray.hs
Martoon-00: :man_facepalming:

src/Dscp/Crypto/Encrypt.hs
Martoon-00: Really, it's safe to show encrypted stuff, will do

src/Dscp/Educator/Secret/Real/Functions.hs
Martoon-00: Btw doesn't it mean that having `--keyfile-path` CLI option would be ambiguous?

src/Dscp/Crypto/Encrypt.hs
```diff
@@ -156,31 +157,47 @@ prepareAEAD (PassPhrase pp) =
...
-        fromByteArray plaintextBS
+    in maybeToRight PassPhraseInvalid $
+       aeadSimpleDecrypt aead authHeader ciphertext tag
+
+-- | 'FromByteArray' versions of 'encryptBA' and 'decryptBA'.
+-- Default implementation is correct if the following holds:
+--
+-- * If some @ba@ is valid, then any @ba'@ of the same length
+-- has to be also valid.
+class EncryptSafe ba where
```
flyingleafe: Frankly, I think that this class is really unnecessary. I struggle to think about some special datatypes, which don't have `ByteArray` instance and cannot have `FromByteArray` one, and still can (and should) be encrypted in some fancy optimized way which doesn't involve serialisation.

tests/Test/Dscp/Util/Serialise.hs
```diff
@@ -0,0 +1,17 @@
+module Test.Dscp.Util.Serialise
+    ( spec_Serialisation
+    ) where
+
+import Test.Common
+
+import Dscp.Util.Aeson (AsByteString, Base64Encoded, Versioned)
+import Test.Dscp.Serialise (aesonRoundtripProp)
+import Test.Dscp.Util.Instances ()
```
Martoon-00: @flyingleafe I did it like in your previous PR, but I'd prefer to have `Test.Dscp.Instances.*` modules instead - otherwise one has to remember each time where the required instance lies and whether is it present at all.

src/Dscp/Crypto/Encrypt.hs
Martoon-00: Yes, but point of this typeclass is to restrict usage of `encrypt` and `decrypt`, since we use this `FromByteArray` stuff and thus they are actually claimed unsafe. I can rename them somehow to show that they work with instances of `ByteArray`/`FromByteArray` to avoid confusion.

src/Dscp/Crypto/Encrypt.hs
```diff
@@ -220,31 +223,41 @@ prepareAEAD pp iv =
...
 encrypt pp plaintext =
     let eIV = fromMaybe (error "encrypt: impossible: random IV with invalid size") .
-              makeIV @ByteString . unsafePerformIO $
+              makeIV @ByteString . unsafePerformIO . runSecureRandom $
               getRandomBytes cipherBlkSize
         aead = prepareAEAD pp eIV
+        plaintextBS = BA.convert plaintext
         (eAuthTag, eCiphertext) =
-            aeadSimpleEncrypt aead authHeader plaintext authTagLength
+            aeadSimpleEncrypt aead authHeader plaintextBS authTagLength
```
Martoon-00: `aeadSimpleEncrypt` returns exactly the same thing as which it accepts :unamused:

PR #46, opened 2018-07-03
[DSCP-130] Use non-static IV for symmetric encryption scheme

flyingleafe:
Using fixed initial vectors for authenticated encryption schemes is unsafe (see https://en.wikipedia.org/wiki/Initialization_vector).
This PR introduces random generation of IV on each encryption. Yes, this means that encrypting the same plaintext with the same password will yield different ciphertexts, but that's actually the point.
Also, now encryption key is derived from a passphrase using PBKDF2 and not simply a hash, which is more secure.

### DIFF
--- a/src/Dscp/Crypto/Encrypt.hs
+++ b/src/Dscp/Crypto/Encrypt.hs
@@ -23,16 +23,18 @@ import Codec.Serialise (Serialise (..))
 import Codec.Serialise.Decoding (decodeBytes)
 import Codec.Serialise.Encoding (encodeBytes)
 import Crypto.Cipher.AES (AES256)
-import Crypto.Cipher.Types (AEAD, AEADMode (AEAD_GCM), AuthTag (..), BlockCipher (aeadInit),
-                            Cipher (cipherInit, cipherKeySize), IV, KeySizeSpecifier (..),
-                            aeadSimpleDecrypt, aeadSimpleEncrypt, nullIV)
+import Crypto.Cipher.Types (AEAD, AEADMode (AEAD_GCM), AuthTag (..),
+                            BlockCipher (aeadInit, blockSize), Cipher (cipherInit, cipherKeySize),
+                            IV, KeySizeSpecifier (..), aeadSimpleDecrypt, aeadSimpleEncrypt, makeIV)
 import Crypto.Error (onCryptoFailure)
 import Crypto.Hash.Algorithms (SHA512 (..))
 import qualified Crypto.KDF.PBKDF2 as PBKDF2
+import Crypto.Random (getRandomBytes)
 import Data.ByteArray (ByteArray, ByteArrayAccess)
 import qualified Data.ByteArray as BA
 import Data.Text.Buildable (build)
 import Fmt ((+|), (|+))
+import System.IO.Unsafe (unsafePerformIO)
 import Text.Show (show)
 
 import Dscp.Crypto.ByteArray (FromByteArray (..))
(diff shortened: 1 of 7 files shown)


src/Dscp/Crypto/Encrypt.hs
```diff
@@ -112,6 +112,10 @@ type CipherType = AES256
 authHeader :: ByteString
 authHeader = ""
 
+-- | Salt used for key derivation from password.
+passGenSalt :: ByteString
+passGenSalt = "dscp-educator-pass-gen"
```
Martoon-00: If you use fixed salt then adversary can still generate rainbow table and guess your password.
I may be wrong though. If I understand correctly, we used password hash as salt in similar code.

flyingleafe: No, in Cardano we've also used fixed salt for symmetric encryption key derivation (see `Pos.Crypto.HD`).
Generating a rainbow table for given key derivation function is kind of tiresome. The purpose of the _salt_ is to make your own key derivation function unique and different from other ones, thus preventing a rainbow table for your function from being _already available_.

src/Dscp/Crypto/Encrypt.hs
```diff
@@ -125,37 +131,107 @@ authHeader = ""
...
+            keyFromPassPhrase pp
         cipher :: CipherType =
             onCryptoFailure impossible identity $
             cipherInit ppHashKey
     in onCryptoFailure impossible identity $
-       aeadInit aeadMode cipher initIV
+       aeadInit aeadMode cipher iv
 
 -- | Encrypt given 'ByteArray' with AES.
+-- TODO: use Secure Random from OpenSSL for IV calculation instead.
```
volhovm: Why do you need it?

src/Dscp/Util.hs
```diff
@@ -32,3 +42,23 @@ wrapRethrowIO
...
+
+toBase :: ByteArrayAccess ba => Base -> ba -> Text
+toBase base = decodeUtf8 @Text @ByteString . convertToBase base
+
+fromBase :: forall ba. FromByteArray ba => Base -> Text -> Either String ba
+fromBase base =
+    convertFromBase base . encodeUtf8 @Text @ByteString >=>
+    fromByteArray @ba @ByteString
+
+toBase64, toHex :: ByteArrayAccess ba => ba -> Text
```
volhovm: We also have it in `serokell-util`. Not saying you necessarily should use it, but...

tests/Test/Dscp/Crypto/Serialise.hs
```diff
@@ -0,0 +1,12 @@
+module Test.Dscp.Crypto.Serialise where
```
volhovm: Please also add positive/negative tests for the encryption itself (not for the binary instances only).

volhovm (changes requested):
Also, can you please specify that `Dscp.Crypto.Encrypt` is a symmetric encryption module (in the module's haddock)? In general the PR is ok.

tests/Test/Dscp/Crypto/Serialise.hs
flyingleafe: They are already added in previous PR, look at the `Test.Dscp.Crypto.Encrypt` module.

volhovm: Aha! Alright then.

src/Dscp/Util.hs
flyingleafe: Fair enough, but these ones work for every `ByteArrayAccess`, so I'll stick with them.

src/Dscp/Crypto/Encrypt.hs
flyingleafe: For the same reason why we might need it at all - to make random more _random_ where _randomness_ matter. Do you claim that there's no need for this and using Secure Random(tm) from OpenSSL is simply irrelevant?

volhovm (approved):

src/Dscp/Crypto/Encrypt.hs
volhovm: I doubt that we actually need to use openssl. It's not like it's the only way to generate random data properly.

PR #48, opened 2018-07-03
[DSCP-131] Properly terminate zmq context

volhovm:
@Martoon-00 pls test, it should be alright now. Actually, before that, zmq context was terminating also, but in 5 seconds (I have no idea where does this number come from).

Relevant change in lootbox:
https://github.com/serokell/lootbox/commit/9b803ba99ce0970766bf5496a855289408daad23


Martoon-00:
@volhovm Btw do you understand which pending messages did it wait for?

volhovm:
@Martoon-00 
>As far as I understand, that delay should only fire if there are yet unsent messages.

Not really. Indeed, the effect you're describing takes place as well, but the particular problem was only caused by not closing sockets. I've set LINGER option of client's socket to 0 as well because we won't receive the response anyway.

PR #51, opened 2018-07-05
[DSCP-123] Simple launcher script

Martoon-00:
Added launcher scripts for both educator and witness nodes.
Without that, launching educator is now nontrivial (I want to avoid questions like which password to use or where educator's secret lies). So we need at least an always-actual example of node CLI parameters.  

Layout of files used by nodes is also changed to keep everything in single place. Read-only files are put separately from files which change after every launch, both in dedicated directories (not to litter project root), although I don't think I managed to end with good names for those directories.


scripts/launch/node.sh
```diff
@@ -0,0 +1,60 @@
+#!/bin/bash
```
lambda76: Please use `/usr/bin/env bash`, `/bin/bash` dose not exist on NixOS.

scripts/launch/node.sh
```diff
@@ -0,0 +1,60 @@
...
+--db-path $tmp_files/witness.db
+--log-dir $tmp_files/logs
+"
+
+if [[ "$mode" == "educator" ]]; then
+    stack exec "disciplina-educator" -- $witness_params $educator_params
+elif [[ "$mode" == "witness" ]]; then
+    stack exec "disciplina-witness" -- $witness_params
+elif [[ "$mode" == "" ]]; then
+    echo "Please specify which node to run (e.g. \"educator\" or just \"e\")"
```
lambda76: `'s/node/mode/'`

scripts/launch/node.sh
```diff
@@ -0,0 +1,60 @@
...
+##################
+
+for var in "$@"
+do
+    if [[ "$var" == "witness" ]] || [ "$var" == "w" ]; then
+        mode="witness"
+    elif [[ "$var" == "witness" ]] || [ "$var" == "e" ]; then
+        mode="educator"
+    else
+        echo "Unknown parameter \"$var\", ignoring"
```
volhovm: Better fail.

scripts/launch/node.sh
Martoon-00: That is so for purpose, I see it like we have educator node and witness node.

Martoon-00: Right, set it `node` everywhere

Martoon-00:
Thanks for comments, fixed.

PR #53, opened 2018-07-08
[DSCP-124] Provide logging context into resource allocation

Martoon-00:
Split resource allocation into two phases, first prepares logging so that in second phase we can use it for other resources allocation and reporting possible errors.

### DIFF
--- a/src/Dscp/Educator/Secret/Real/Functions.hs
+++ b/src/Dscp/Educator/Secret/Real/Functions.hs
@@ -6,15 +6,17 @@ module Dscp.Educator.Secret.Real.Functions
 
 import Data.Aeson (eitherDecode', encode)
 import qualified Data.ByteString.Lazy as LBS
+import Fmt ((+|), (|+))
+import Loot.Log (MonadLogging, logInfo)
 import System.Directory (doesFileExist)
 import System.FilePath ((</>))
 
 import Dscp.Crypto (PassPhrase, decrypt, encrypt, genSecretKey, runSecureRandom)
-import Dscp.Educator.Secret.Real.Error (EducatorSecretError (..), rewrapSecretIOError)
+import Dscp.Educator.Secret.Real.Error (EducatorSecretError (..), rewrapSecretIOErrors)
 import Dscp.Educator.Secret.Real.Types (EducatorSecret (..), EducatorSecretJson (..),
                                         EducatorSecretParams (..), KeyfileContent)
 import Dscp.Resource.AppDir (AppDirectory (..))
-import Dscp.System (ensureModeIs, mode600, whenPosix)
+import Dscp.System (ensureModeIs, mode600, setMode, whenPosix)
 import Dscp.Util (leftToThrow)
 import Dscp.Util.Aeson (Versioned (..))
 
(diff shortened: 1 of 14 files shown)


src/Dscp/Educator/Secret/Real/Functions.hs
```diff
@@ -51,11 +53,13 @@ genStore = do
     return EducatorSecret{..}
 
 -- | Read store under given path.
-readStore :: (MonadIO m, MonadCatch m) => FilePath -> PassPhrase -> m EducatorSecret
+readStore
+    :: (MonadIO m, MonadCatch m, MonadLogging m)
```
flyingleafe: Why did you add `MonadLogging` here, if you didn't add any actual logging? :thinking:

src/Dscp/Resource/Functions.hs
```diff
@@ -0,0 +1,67 @@
...
+    eres <- try $ runComponentM initDesc (allocInitResources params) $
+      \initCtx -> do
+         eres <- try $ runComponentR desc initCtx component main
+         whenLeft eres $ \(err :: ComponentError) -> do
+             runRIO initCtx $ logError (""+||Doc.pretty err||+"")
+         either throwM pure eres
+    whenLeft eres $ \case
+        ComponentRuntimeFailed{ componentErrorOriginalException = cause }
+            | isJust (fromException @ComponentError cause) ->
+            -- already printed error above
```
flyingleafe: Is it possible to have something _different_ but `ComponentError` inside `ComponentRuntimeFailed`? If it is, then we'll silently pass an error in this case.

src/Dscp/Resource/Functions.hs
```diff
@@ -0,0 +1,67 @@
...
+
+import Dscp.Launcher.Rio (runRIO)
+import Dscp.Resource.Class (InitContext (..), InitParams (..))
+import Dscp.Resource.Logging (allocLogging)
+
+-- | Catches and prints possible synchronous exceptions thrown
+-- in 'runComponentM'. This includes both resource allocation errors and
+-- application errors.
+handleComponentErrors :: IO a -> IO (Either ComponentError a)
+handleComponentErrors action = do
```
flyingleafe: This function seems to be kinda useless, as `runResourceAllocation` does it manually (for reasons).

src/Dscp/Educator/Secret/Real/Functions.hs
Martoon-00: `ensureModeIs` can report upon fixing file mode

src/Dscp/Resource/Functions.hs
Martoon-00: Right :man_facepalming:

src/Dscp/Educator/Secret/Real/Functions.hs
flyingleafe: Ok, sorry, missed it.

src/Dscp/Resource/Functions.hs
Martoon-00: Won't control pass to the second `case` branch is situation you describe?

flyingleafe: @Martoon-00 yep sorry, that was stupid :man_facepalming:

PR #56, opened 2018-07-11
[DSCP-146] Servant API for Student

flyingleafe:
- [x] JSON instances for all necessary types
- [x] Type-level definition of API
- [x] Stub handlers
- [x] Integration with executor and CLI interface
- [ ] Bindings with actual DB logic

### DIFF
--- a/src/Dscp/CLI/Common.hs
+++ b/src/Dscp/CLI/Common.hs
@@ -12,6 +12,8 @@ module Dscp.CLI.Common
        , ourZTNodeIdParser
        , netCliParamsParser
        , netServParamsParser
+
+       , networkAddressParser
        ) where
 
 import qualified Data.Set as Set
@@ -25,6 +27,7 @@ import Dscp.DB.Rocks.Real.Types (RocksDBParams (..))
 import Dscp.DB.SQLite.Types (SQLiteDBLocation (..), SQLiteParams (..))
 import Dscp.Resource.Logging (LoggingParams (..))
 import Dscp.Resource.Network (NetCliParams (..), NetServParams (..))
+import Dscp.Web (NetworkAddress (..))
 import Paths_disciplina (version)
 
 logParamsParser :: Log.Name -> Parser LoggingParams
@@ -88,3 +91,22 @@ netCliParamsParser = NetCliParams <$> peersParser
 
 netServParamsParser :: Parser NetServParams
 netServParamsParser = NetServParams <$> peersParser <*> ourZTNodeIdParser
+
+---------------------------------------------------------------------------
+-- Utils
+---------------------------------------------------------------------------
+
+parseNetAddr :: String -> Either String NetworkAddress
+parseNetAddr str =
+    let host = toText $ takeWhile (/= ':') str
+        portS = drop (length host + 1) str
+    in case readMaybe portS of
+        Just port -> Right $ NetworkAddress host port
+        Nothing   -> Left "Invalid network address"
+
+networkAddressParser :: String -> String -> Parser NetworkAddress
+networkAddressParser pName helpTxt =
+    option (eitherReader parseNetAddr) $
+    long pName <>
+    metavar "HOST:PORT" <>
+    help helpTxt
--- a/src/Dscp/Core/Grade.hs
+++ b/src/Dscp/Core/Grade.hs
@@ -1,0 +1,21 @@
+
+-- | Utils for working with grades
+
+module Dscp.Core.Grade
+       ( gA
+       , gB
+       , gC
+       , gD
+       , gE
+       , gF
+       ) where
+
+import Dscp.Core.Types (Grade (..))
+
+gA, gB, gC, gD, gE, gF :: Grade
+gA = UnsafeGrade 100
+gB = UnsafeGrade 80
+gC = UnsafeGrade 60
+gD = UnsafeGrade 40
+gE = UnsafeGrade 20
+gF = UnsafeGrade 0
--- a/src/Dscp/Core/Types.hs
+++ b/src/Dscp/Core/Types.hs
@@ -9,6 +9,7 @@ module Dscp.Core.Types
        , Subject (..)
        , Student
        , Grade (..)
+       , mkGrade
        , EducatorId
        , Assignment (..)
        , AssignmentType (..)
@@ -68,9 +69,20 @@ newtype Subject = Subject
 instance HasId Subject
 
 -- | Assignment/course grade.
--- TODO: decide on final format of the grade.
-data Grade = F | D | C | B | A
-    deriving (Eq, Ord, Enum, Bounded, Show, Generic)
+-- An integer from 0 to 100. Constructor is unsafe, because it's possible
+-- to make a grade outside these bounds.
+newtype Grade = UnsafeGrade
+    { getGrade :: Word8
+    } deriving (Eq, Ord, Num, Show, Generic)
+
+instance Bounded Grade where
+    minBound = 0
+    maxBound = 100
+
+mkGrade :: Word8 -> Maybe Grade
+mkGrade a
+    | a >= minBound && a <= maxBound = Just $ UnsafeGrade a
+    | otherwise                      = Nothing
 
 -- | Student is identified by their public address.
 type Student = Address
--- a/src/Dscp/Core/Types.hs
+++ b/src/Dscp/Core/Types.hs
@@ -73,11 +73,11 @@ instance HasId Subject
 -- to make a grade outside these bounds.
 newtype Grade = UnsafeGrade
     { getGrade :: Word8
-    } deriving (Eq, Ord, Num, Show, Generic)
+    } deriving (Eq, Ord, Show, Generic)
 
 instance Bounded Grade where
-    minBound = 0
-    maxBound = 100
+    minBound = UnsafeGrade 0
+    maxBound = UnsafeGrade 100
 
 mkGrade :: Word8 -> Maybe Grade
 mkGrade a
--- a/src/Dscp/Core/Types.hs
+++ b/src/Dscp/Core/Types.hs
@@ -144,7 +144,7 @@ offlineHash = unsafeHash ("offline" :: ByteString)
 -- | Datatype to represent the notion of "offline"- and "online"-ness
 -- of assignments and submissions.
 data DocumentType = Online | Offline
-    deriving (Eq, Show, Generic)
+    deriving (Eq, Ord, Show, Enum, Generic)
 
 documentType :: Hash Raw -> DocumentType
 documentType h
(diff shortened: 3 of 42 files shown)


src/Dscp/Educator/Web/Student/API.hs
```diff
@@ -0,0 +1,111 @@
...
+    :> Verb 'GET 200 '[JSON] Submission
+
+type MakeSubmission
+    = "submissions"
+    -- :> Summary "Make a new submission"
+    -- :> Description "Posts a new submission with a given body. Request body should \
+                   -- \contain valid student's signature of submission contents, \
+                   -- \otherwise an error will be raised."
+    :> ReqBody '[JSON] Core.SignedSubmission
+    :> Verb 'POST 200 '[JSON] Submission
```
Martoon-00: Btw `POST` usually returns 201.

src/Dscp/Educator/Web/Student/Types.hs
```diff
@@ -0,0 +1,73 @@
...
+    , aDesc           :: !Text
+    , aLastSubmission :: !(Maybe Submission)
+    } deriving (Show, Eq, Generic)
+
+data Submission = Submission
+    { sHash           :: !(Hash Core.Submission)
+    , sContentsHash   :: !(Hash Raw)
+    , sAssignmentHash :: !(Hash Core.Assignment)
+    , sGrade          :: !(Maybe Grade)
+    } deriving (Show, Eq, Generic)
```
Martoon-00: I guess `sGrade` will be Nothing for request and Just for response. Won't we wish separate datatypes?

src/Dscp/Educator/Web/Student/Types.hs
```diff
@@ -0,0 +1,73 @@
...
+---------------------------------------------------------------------------
+
+deriveJSON defaultOptions ''PrivateTx
+
+deriveJSON defaultOptions ''Course
+deriveJSON defaultOptions ''Assignment
+deriveJSON defaultOptions ''Submission
+deriveJSON defaultOptions ''Grade
+deriveJSON defaultOptions ''BlkProof
+deriveJSON defaultOptions ''ErrResponse
```
Martoon-00: `defaultOptions` is an evil, let's use at least 'Serokell.Util' one's

tests/Test/Dscp/DB/DSL/Interpret/SimpleTxDB.hs
```diff
@@ -48,27 +48,27 @@ mkLinAlgPrivateTx :: StudentKeySeed -> EducatorKeySeed -> PrivateTx
 mkLinAlgPrivateTx sKeySeed eKeySeed =
     let studentKey = mkPubKey sKeySeed
         educatorKeyPair = mkKeyPair eKeySeed
-    in mkPrivateTx courseLinearAlg C studentKey educatorKeyPair
+    in mkPrivateTx courseLinearAlg 60 studentKey educatorKeyPair
```
Martoon-00: Someone is going to suffer when grade representation changes again. How about extracting several constants?

src/Dscp/Educator/Web/Student/API.hs
```diff
@@ -0,0 +1,111 @@
...
+---------------------------------------------------------------------------
+
+type GetAssignments
+    = "assignments"
+    -- :> Summary "Get student's assignments"
+    -- :> Description "Gets a list of student's assignments. Filter parameters are \
+                   -- \used to specify specific course, type, etc."
+    :> QueryParam "course" Core.Course
+    :> QueryParam "type" Core.DocumentType
+    :> QueryParam "final" Bool
```
Martoon-00: I'm not sure whether my comment to swagger PR was published
It seems to me that having 'Bool' here is not good because you can not add new assignment types in a backward-compatible way.

Martoon-00 (changes requested):
Here's my diss.
I'd also like to see roundtrip tests, although I guess you're working on it right now.

src/Dscp/Educator/Web/Student/Types.hs
flyingleafe: You got it wrong. Student will send actual `Core.SignedSubmission` datatype to POST method. This adapter `Submission` type is only for responses. A submission may be already graded or not graded yet, hence `Maybe` here.

src/Dscp/Educator/Web/Student/Types.hs
flyingleafe: I do. `Data.Aeson.Options` is a module from `aeson-options` package, which has been extracted from `Serokell.Util` by @vrom911 some time ago.

tests/Test/Dscp/DB/DSL/Interpret/SimpleTxDB.hs
flyingleafe: Fair enough, I just needed to fix tests then.

src/Dscp/Educator/Web/Student/Types.hs
Martoon-00: Oh got it, nice.

Martoon-00 (approved):
LGTM

src/Dscp/Educator/Web/Student/Error.hs
```diff
@@ -0,0 +1,29 @@
...
+-- | API errors
+
+module Dscp.Educator.Web.Student.Error
+       ( APIError (..)
+       ) where
+
+import Data.Aeson (FromJSON (..), ToJSON (..), Value (..), withText)
+import Data.Data (Data (..), toConstr)
+
+data APIError
```
volhovm: Pls document these errors somewhere, at least in the API description.

src/Dscp/Educator/Web/Student/Types.hs
```diff
@@ -0,0 +1,95 @@
...
+import Dscp.Core.Address (addrFromText)
+import Dscp.Core.Aeson ()
+import qualified Dscp.Core.Types as Core
+import Dscp.Crypto (Hash, Raw)
+import Dscp.Educator.Txs (PrivateTx (..))
+import Dscp.Educator.Web.Student.Error (APIError)
+import Dscp.Util (fromBase64)
+import Dscp.Util.Aeson (AsByteString, Base64Encoded)
+
+data Course = Course
```
volhovm: Even though these datatypes are in different modules, a different naming would look better IMO. How about `W` prefix? `WCourse`/`WAssignment` etc?

src/Dscp/Web/Server.hs
```diff
@@ -0,0 +1,21 @@
+
+-- | Utils for serving HTTP APIs
+
+module Dscp.Web.Server
```
volhovm: `ServerUtils` maybe then?

stack.yaml
```diff
@@ -8,18 +8,22 @@ nix:
...
 - pretty-tree-0.1.0.0
 - time-units-1.0.0
 - componentm-0.0.0.2
 - rio-0.1.2.0
 - serokell-util-0.8.0
+- formatting-6.2.5
 - typed-process-0.2.2.0
 - teardown-0.5.0.0
 - unliftio-0.2.7.0
+- yaml-0.8.32
```
volhovm: `yaml-0.8.31.1` from the lts is not good enough?

src/Dscp/CLI/Common.hs
```diff
@@ -88,3 +91,22 @@ netCliParamsParser = NetCliParams <$> peersParser
 
 netServParamsParser :: Parser NetServParams
 netServParamsParser = NetServParams <$> peersParser <*> ourZTNodeIdParser
+
+---------------------------------------------------------------------------
+-- Utils
+---------------------------------------------------------------------------
+
+parseNetAddr :: String -> Either String NetworkAddress
```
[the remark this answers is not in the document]
flyingleafe: Yes, it is supposed to parse `host:port`. It might be done much better for sure, I just wanted something which will work right now (kind of)

src/Dscp/Core/Grade.hs
```diff
@@ -0,0 +1,21 @@
+
+-- | Utils for working with grades
+
+module Dscp.Core.Grade
```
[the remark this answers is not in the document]
flyingleafe: When exactly did we agree to do that?... I must've missed that argument.
Also, if you read wikipedia's survey on grading systems (https://en.wikipedia.org/wiki/Grading_systems_by_country), you'll see that they all are compared with percentage. That was my reasoning behind integral 0-100 scale, and I still think that it's more than enough.

src/Dscp/Core/Types.hs
```diff
@@ -68,9 +69,20 @@ newtype Subject = Subject
 instance HasId Subject
 
 -- | Assignment/course grade.
--- TODO: decide on final format of the grade.
-data Grade = F | D | C | B | A
-    deriving (Eq, Ord, Enum, Bounded, Show, Generic)
+-- An integer from 0 to 100. Constructor is unsafe, because it's possible
+-- to make a grade outside these bounds.
+newtype Grade = UnsafeGrade
+    { getGrade :: Word8
```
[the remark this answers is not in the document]
flyingleafe: But why exactly? We will most certainly never need more a more fine portion of a grade than 1/100 :man_shrugging: 
Let's then put `Double` from 0 to 1 here, if you really want to cover _all_ the cases.

src/Dscp/Educator/Web/Student/Types.hs
flyingleafe: Since `purescript-bridge`, I have a subconscious disgust for adapter types with letter prefixes :|
We can do it still, but I'd like to collect more votes for or against.

src/Dscp/Web/Server.hs
flyingleafe: If I wrote "Methods for serving HTTP APIs", you'd suggest to rename it to `ServerMethods`?
I'd better leave it like that to be more consistent with naming of actual API serving modules (e. g. `Dscp.Educator.Web.Student.Server`, later there will be `Dscp.Witness.Web.Explorer.Server`, etc.)

stack.yaml
flyingleafe: Yes, it's not: https://serokell.slack.com/archives/C742ZKYF5/p1531389945000182

src/Dscp/Core/Grade.hs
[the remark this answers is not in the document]
volhovm: 1. https://serokell.slack.com/archives/G7S2TS9BK/p1530029141000299?thread_ts=1530025876.000777&cid=G7S2TS9BK
The main argument as I understand it is that we want to define  `0/1` and `1/1` grades.

2. Anyway, why `gA = 100` and not `91` or `95.5`? Maybe it does make sense to introduce a grade system datatype for conversion?

PR #57, opened 2018-07-11
[DSCP-144] [DSCP145] Fetching private transactions with proofs / Creating blocks

Heimdell:


src/Dscp/DB/SQLite/Queries.hs
```diff
@@ -2,6 +2,10 @@
 
 module Dscp.DB.SQLite.Queries where
```
volhovm: By the way, explicit exports.

src/Dscp/DB/SQLite/Queries.hs
```diff
@@ -140,6 +149,134 @@ getStudentTransactions student = do
...
+
+            return (pruned, bodies)
+
+  where
+    groupToAssocWith :: Eq k => Ord k => (a -> k, a -> v) -> [a] -> [(k, [v])]
+    groupToAssocWith (key, value)
+        -- The `groupBy` should return `[NonEmpty a]`.
+        = map      (\every@ (selected : _) -> (key selected, map value every))
+        . groupBy  ((==) `on` key)
+        . sortWith  key
```
volhovm: You map values three times using `key` -- with `sortWith`, `groupBy` and `map` itself. Isn't it faster to first convert to a `(k,v)` list and then transform into the `(k,[v])`?

src/Dscp/Educator/Block/Type.hs
```diff
@@ -0,0 +1,82 @@
...
+
+-- | Get previous block header, if previous block exists,
+-- 'Nothing' otherwise.
+getPrevBlockRefMaybe :: PrivateBlockHeader -> Maybe PrivateHeaderHash
+getPrevBlockRefMaybe PrivateBlockHeader {..} =
+    if _pbhPrevBlock == genesisHeaderHash
+    then Nothing
+    else Just _pbhPrevBlock
+
+-- | Private block body. Contains only private transactions (for now).
```
volhovm: >"for now"

Anything else is planned to be stored here? What am I missing?

src/Dscp/Educator/Txs/Type.hs
```diff
@@ -0,0 +1,62 @@
...
+-- 'PrivateTxId' is basically a hash of all transaction contents,
+-- so it's sufficient to sign only that.
+type PrivateTxSigData = PrivateTxId
+
+-- | Type alias for private tx signature.
+type PrivateTxSig = Signature PrivateTxSigData
+
+-- | Witness contains data required to verify transaction.
+-- Included 'PublicKey' belongs either to Student or Educator.
+-- TODO: maybe we can say that Educator's key is already known
```
volhovm: Yes. That's important -- there is no point in storing educator signatures on the educator's side sometimes.

src/Dscp/DB/SQLite/Instances.hs
```diff
@@ -28,16 +32,21 @@ instance FromField Course            where fromField f = Course            <$> f
 instance FromField Grade             where fromField f = UnsafeGrade       <$> fromField f
 instance FromField AssignmentType    where fromField f = Codec.deserialise <$> fromField f
 instance FromField SubmissionWitness where fromField f = Codec.deserialise <$> fromField f
+instance FromField ATGDelta          where fromField f = Codec.deserialise <$> fromField f
 
 instance ToField   (Hash a)          where toField = toField . Codec.serialise
 instance ToField   (Signature a)     where toField = toField . Codec.serialise
+instance ToField (MerkleSignature a) where toField = toField . Codec.serialise
+instance Serialise a =>
+         ToField   (MerkleTree a)    where toField = toField . Codec.serialise
```
flyingleafe: Won't work as expected - `MerkleTree a` has actual values of `a` inside it (which in our case are actual transactions). We definitely don't want to serialise transactions themselves into this field.
The simplest way to do it right, imho, is to have a `newtype EmptyMerkleTree a = Empty (MerkleTree ())` with functions like `getEmptyMerkleTree :: MerkleTree a -> EmptyMerkleTree a` and `fillEmptyMerkleTree :: Map LeafIndex a -> EmptyMerkleTree a -> MerkleProof a`, and serialise this newtype into SQLite instead, making it from a full Merkle tree on block creation and combining it with fetched transactions to get a proof after fetching.

src/Dscp/DB/SQLite/Queries.hs
```diff
@@ -140,6 +149,134 @@ getStudentTransactions student = do
...
+            , block^.pbHeader.pbhAtgDelta
+            , root
+            , tree
+            )
+
+        for_ txs' $ \(idx, tid) -> do
+            _ <- getTransaction tid `assertJust` TransactionDoesNotExist tid
+
+            execute setTxIndexRequest    (idx :: Word32, tid)
+            execute assignToBlockRequest (bid, tid)
```
flyingleafe: Why not do both of these with one `update`?

src/Dscp/DB/SQLite/Queries.hs
```diff
@@ -140,6 +149,134 @@ getStudentTransactions student = do
...
+                    prev_hash
+                    atg_delta
+                    mroot
+                    mtree
+            from    Blocks
+            where   idx = ?
+        |]
+
+createBlock :: DBM m => PrivateBlock -> m (Id PrivateBlock)
+createBlock block = do
```
flyingleafe: The block will always be constructed from the set of all currently blockless transactions (those which don't have a corresponding record in `BlockTxs` table). Why then not have the type `createBlock :: DBM m => m PrivateBlock` and perform fetching these them and constructing the block in the same DB transaction with saving it back to the DB?

src/Dscp/Educator/Block/Type.hs
flyingleafe: This is my comment, moved from the old file. Yes, cannot really think about anything else now, "for now" can be removed.

src/Dscp/Educator/Txs/Type.hs
flyingleafe: Actually, this is some redundant stuff - we have student's signature in `SignedSubmission` datatype anyway. Types `PrivateTxWitness` and `PrivateTxAux` should be removed altogether (dunno how they're still here)

src/Dscp/DB/SQLite/Queries.hs
flyingleafe: ok, fair enough :facepalm:

Heimdell:
Will fix

Heimdell:
Update to "will fix": actually, I don't fill `plugs` directly. I do (watch my hands!)
```haskell
let
    -- get all keys from the map
    keySet = Set.fromList (keys plugs)

    -- construct proof from tree of `()`s, containing no more paths than to `keySet`
    proof  = mkMerkleProof sieve keySet

    -- fill that already pruned (!) proof with values
    filled = fill <$> proof
in
    filled
```

Thats why

>  idea is to fill only those leaves of empty merkle tree which are included in plugs, leaving all others pruned

is already done. But, of course, I need to make a test to prove that.

src/Dscp/DB/SQLite/Instances.hs
Heimdell: Done.

src/Dscp/DB/SQLite/Queries.hs
Heimdell: I left one `Maybe ATGDelta` param there.

Heimdell:
Update to update: I removed length comparisons here.

lib/src/Dscp/DB/SQLite/Queries.hs
```diff
@@ -155,6 +224,184 @@ getStudentTransactions student = do
...
+            return (pruned, indiced)
+
+        return [(proof, txs) | (Just proof, txs) <- results]
+
+  where
+    groupToAssocWith :: Ord k => Ord k => (a -> k, a -> v) -> [a] -> [(k, [v])]
+    groupToAssocWith (key, value) =
+        Map.toList . foldr' push Map.empty
+      where
+        push a = Map.insertWith (++) (key a) [value a]
```
volhovm: Why not `Map.insertWith (flip (:)) (key a) (value a)` ?

lib/src/Dscp/DB/SQLite/Queries.hs
```diff
@@ -155,6 +224,184 @@ getStudentTransactions student = do
...
+                  and  Transactions.idx                <> -1
+        |]
+
+    -- Returns `PrivateBlock` in normalized format, with metadata.
+    getBlockData :: DBM m => Word32 -> m BlockData
+    getBlockData blockIdx = do
+        (listToMaybe <$> query getBlockDataQuery (Only blockIdx))
+            `assertJust` BlockWithIndexDoesNotExist blockIdx
+      where
+        -- TODO: prune fields that aren't needed.
```
volhovm: Is it still relevant?

lib/src/Dscp/DB/SQLite/Queries.hs
```diff
@@ -155,6 +224,184 @@ getStudentTransactions student = do
...
+        order by  idx desc
+        limit     1
+    |] ()
+
+createBlock :: DBM m => Maybe ATGDelta -> m ()
+createBlock delta = do
+    (prev, idx) <- getLastBlockIdAndIdx
+    txs         <- getAllNonChainedTransactions
+
+    transaction $ do
```
volhovm: I guess you can execute some of the `let` bindings strictly outside of the `transaction`, but performance is not a problem now.

lib/src/Dscp/Educator/Txs/Instances.hs
```diff
@@ -0,0 +1,21 @@
...
+
+-- | Transactions
+instance Serialise PrivateTx
+instance Serialise PrivateTxWitness
+instance Serialise PrivateTxAux
+
+instance HasId PrivateTx where
+    type Id PrivateTx = Hash PrivateTx
+    getId = hash
+
```
volhovm: :eyes:

lib/tests/Test/Common.hs
```diff
@@ -300,3 +300,9 @@ throwsPrism
     => Prism' e a -> m b -> m Bool
 throwsPrism excL action = do
     catchJust (^? excL) (action $> False) (\_ -> return True)
+
+puts :: MonadIO m => String -> m ()
+puts = liftIO . putStrLn
+
+put' :: MonadIO m => Show a => a -> m ()
+put' = liftIO . print
```
volhovm:
```
Prelude Universum> :t Universum.print
Universum.print :: (Show a, MonadIO m) => a -> m ()
```

lib/tests/Test/Common.hs
```diff
@@ -300,3 +300,9 @@ throwsPrism
     => Prism' e a -> m b -> m Bool
 throwsPrism excL action = do
     catchJust (^? excL) (action $> False) (\_ -> return True)
+
+puts :: MonadIO m => String -> m ()
```
volhovm: `putText`

lib/src/Dscp/Educator/Block/Type.hs
```diff
@@ -0,0 +1,82 @@
...
+
+-- | Header of a private block. There's no signatures here, as it's private anyway.
+-- During publishing, Educator will provide a signature as a part of transaction.
+data PrivateBlockHeader = PrivateBlockHeader
+    { _pbhPrevBlock :: !PrivateHeaderHash
+    -- ^ Previous header in the chain
+    , _pbhBodyProof :: !(MerkleSignature PrivateTx)
+    -- ^ Body payload proof (for now - only root of sized Merkle tree
+    -- over private transactions)
+    , _pbhAtgDelta  :: !ATGDelta
```
volhovm: Can you comment it out for now? We're not yet sure how to resolve it in public chain.

lib/src/Dscp/DB/SQLite/Queries.hs
Heimdell: It will reverse order (I don't know if we need to preserve it).

lib/src/Dscp/DB/SQLite/Queries.hs
Heimdell: Nope. I choose to hold all data in th end.

lib/src/Dscp/DB/SQLite/Queries.hs
Heimdell: The `let` binding is almost no-op, it doesn't execute - it just binds. Wherever you declare `foo`, it will be evaluated on the first place it is called.

And I propose optimising after we have benchmarks.

lib/src/Dscp/Educator/Txs/Instances.hs
Heimdell: What?

lib/tests/Test/Common.hs
Heimdell: Will remove these declarations.

lib/tests/Test/Common.hs
Heimdell: The same.

lib/src/Dscp/DB/SQLite/Queries.hs
Heimdell: ... up to the first constructor. We need benchmarks - this might be non-bottleneck.

lib/src/Dscp/DB/SQLite/Queries.hs
Heimdell: I used `flip (++)`, because `Map.insertWith` expects `v -> v -> v` there.

lib/src/Dscp/DB/SQLite/Queries.hs
Heimdell: And how does it improve performance? The evaluation has to be done no matter when.

PR #61, opened 2018-07-20
[DSCP-156] Add loot-configurations carried with reflection

Martoon-00:
Dunno what I'm doing here :man_shrugging: 
At least one can add his stuff to config now easily.

Disses are welcome.


lib/src/Dscp/Config/AppDir.hs
```diff
@@ -0,0 +1,48 @@
...
+-- | Create application directory if absent.
+prepareAppDir
+    :: MonadIO m
+    => AppDirectoryParam -> m FilePath
+prepareAppDir param = do
+    appDir <- case param of
+        AppDirectoryOS            -> getOSAppDir
+        AppDirectorySpecific path -> pure path
+    -- we would unlikely have logging context here
+    putTextLn $ "Application home directory will be at "+|appDir|+""
```
volhovm: That's why you shouldn't initialise as a "init resource", but rather as a separate entity.

Martoon-00:
Ueeee :ny_veselimsya:

PR #63, opened 2018-07-22
[AD-307] Create basic REPL wallet with stub API implementations

goodsoft:
Running: `stack exec disciplina-wallet`
Available commands: `help`, `gen-key-pair`, `send-tx`, `get-balance`

API method stubs are in `Dscp.Wallet.Backend`.

Code in `Ariadne` directory is copy-pasted from Ariadne almost verbatim, except for CLI, and will be moved back, when we decouple Ariadne components more and remove Cardano dependencies.


wallet/src/Dscp/Wallet/Knit.hs
```diff
@@ -0,0 +1,170 @@
...
+      , toToken . TokenSecretKey <$> pSecretKey
+      ]
+    where
+      pAddress :: Tokenizer Address
+      pAddress = do
+        str <- pSomeAlphaNum
+        either (fail . show) return $ addrFromText str
+
+      pCoin :: Tokenizer Coin
+      pCoin = Coin <$> P.signed (return ()) P.decimal
```
Martoon-00: `Coin` is still unsigned

wallet/src/Dscp/Wallet/Knit.hs
```diff
@@ -0,0 +1,170 @@
...
+        , cpHelp = "Generate a key pair."
+        }
+    , CommandProc
+        { cpName = "send-tx"
+        , cpArgumentPrepare = identity
+        , cpArgumentConsumer = do
+            passPhrase <- getArg tyString "passphrase"
+            secretKey <- getArg tySecretKey "secretkey"
+            address <- getArg tyAddress "address"
+            value <- getArg tyCoin "value"
```
Martoon-00: Btw is it possible to override argument consumer so that user can specify either `[(Address, Coin)]` or `(Address, Coin)` if the latter is easier?
We'd like to provide multi-output transactions support, although one-output transaction is still the most common use case.

wallet/src/Dscp/Wallet/Knit.hs
```diff
@@ -0,0 +1,170 @@
...
+      pAddress = do
+        str <- pSomeAlphaNum
+        either (fail . show) return $ addrFromText str
+
+      pCoin :: Tokenizer Coin
+      pCoin = Coin <$> P.signed (return ()) P.decimal
+
+      pSecretKey :: Tokenizer ByteString
+      pSecretKey = do
+        str <- (<>) <$> P.takeP (Just "base64") 91 <*> P.string "="
```
Martoon-00: Why can't you read till e.g. a non-letter/EOF?
If that's needed for performance, please move the constant next to `SecretKey` type.

Martoon-00: Btw encoded secret key takes 44 chars for me with the trailing `=`, how could it be :thinking:

wallet/src/Dscp/Wallet/Knit.hs
goodsoft: Right, I forgot it's `Word64`, not `Scientific`

wallet/src/Dscp/Wallet/Knit.hs
goodsoft: In Cardano backend it looks like this:
`send <wallet id, pass, etc> out:(tx-out <address> <value>) out:(tx-out <address> <value>) out:...`
Not sure I'll be able to make it better looking

wallet/src/Dscp/Wallet/Knit.hs
goodsoft: Unfortunately, it will consider _any_ base64 string as secret key then.
Knit is still somewhat weirdly typed language.

For Cardano we "solved" it by requiring `#` symbol before it.
I think in this case I'll just require all keys and addresses to require quotes around them as normal strings, and leave all validation to command representation below, like with passphrases.

goodsoft: I generated secret key using `keyGen`, `mkPassPhrase` and `encrypt`, got a string like this:
`UH1OoUs/eDxxKHMlLy09+m5Qw67i+iDVvJ6wZora1TDY6FgggZoDtW8FW7TD0iUx99vPg4q4mUydmMRZLiT9bA+sEug=`

goodsoft: Actually, disregard that, I can do it the same way as addresses.

Martoon-00: > I generated secret key using keyGen, mkPassPhrase and encrypt, got a string like this:
UH1OoUs/eDxxKHMlLy09+m5Qw67i+iDVvJ6wZora1TDY6FgggZoDtW8FW7TD0iUx99vPg4q4mUydmMRZLiT9bA+sEug=
Right, I forgot to encrypt secret key :man_facepalming:

wallet/src/Dscp/Wallet/Knit.hs
goodsoft:
```
knit> send-tx "asdfasdfasdfasdf" UH1OoUs/eDxxKHMlLy09+m5Qw67i+iDVvJ6wZora1TDY6FgggZoDtW8FW7TD0iUx99vPg4q4mUydmMRZaLiT9b+sEug= (tx-out LL4qKtG4mmZF4nC8hMiTpmsG7noszDtWZhiVHuqQsEqtghWdiCFmkkMp 10) (tx-out LL4qKtG4mmZF4nC8hMiTpmsG7noszDtWZhiVHuqQsEqtghWdiCFmkkMp 20)
Not implemented: sendTx
   <passphrase>
   UH1OoUs/eDxxKHMlLy09+m5Qw67i+iDVvJ6wZora1TDY6FgggZoDtW8FW7TD0iUx99vPg4q4mUydmMRZaLiT9b+sEug=
   TxOut {txOutAddr = Address {addrHash = AbstractHash ee70896bece23911966a2a7522a8cc11c9cefb58afed3f3d13bfe65a244865aa}, txOutValue = Coin {unCoin = 10}}
   TxOut {txOutAddr = Address {addrHash = AbstractHash ee70896bece23911966a2a7522a8cc11c9cefb58afed3f3d13bfe65a244865aa}, txOutValue = Coin {unCoin = 20}}

CallStack (from HasCallStack):
  error, called at src/Universum/Debug.hs:57:11 in universum-1.1.0-EGc4GwPNR66DFTSyrseD1k:Universum.Debug
  error, called at src/Dscp/Wallet/Backend.hs:20:36 in disciplina-wallet-0.1.0.0-nMR7bvFJpEFS9IJUSXcmM:Dscp.Wallet.Backend
knit> 
```

PR #64, opened 2018-07-23
[DSCP-159] Integrated validators and changed them a bit

Heimdell:
- Changed `HasPrism` to `HasGetter` for all the validator properties because we don't need to reconstruct `proof` from anything. I blindly copied `HasPrism` constraints back then. Thats what reading SD code does to people.
- Implemented all validatior requirements on `Proofs`.
- Changed requirements to `Expander` to signed `(TxId, PublicKey)`, since `Address` is wrapper around `Hash PublicKey` and `AuthorId` can be retrieved from that.
- Pulled authentication in validator into separate method.

### DIFF
--- a/lib/src/Dscp/Core/Types.hs
+++ b/lib/src/Dscp/Core/Types.hs
@@ -327,10 +327,10 @@ type TxId = Hash Tx
 txId :: (Serialise Tx) => Tx -> TxId
 txId = hash
 
--- | Transaction witness. We sign a pair of transaction hash and input
--- account. The second element of the pair is redundand, and only
--- included to speed up and simplify transaction validation
--- (@volhovm). Public key should correspond to the input address.
+-- | Transaction witness. We sign a pair of transaction hash and private
+-- key. The second element is there to authenticate proposed changes
+-- (@kirill.andreev). Public key hash should be equal to the input address.
+-- Also, public key should be the same which used to validate signature.
 data TxWitness = TxWitness
     { txwSig :: Signature (TxId, PublicKey)
     , txwPk  :: PublicKey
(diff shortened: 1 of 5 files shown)


lib/src/Dscp/Core/Types.hs
```diff
@@ -332,7 +332,7 @@ txId = hash
 -- included to speed up and simplify transaction validation
 -- (@volhovm). Public key should correspond to the input address.
 data TxWitness = TxWitness
-    { txwSig :: Signature (TxId, TxInAcc)
+    { txwSig :: Signature (TxId, PublicKey)
```
volhovm: You should also change the comment above.

lib/src/Dscp/Snowdrop/AccountValidation.hs
```diff
@@ -83,86 +84,82 @@ assertSigned WithSignature {..} message = do
...
+    -> ERoComp      e         id       value ctx (AccountId addr, Account)
+authenticate proof = do
+    let signedHash :: WithSignature pk sig (hash, pk)
+            = gett proof
 
-        realHashfromExpander
-          :: hash
-          <- requirePart txProof SignatureIsMissing
+        realHashfromExpander :: hash
+            = gett proof
```
volhovm: Why do you wrap lines like this? :thinking:

Heimdell: I also do
```
data Foo
    = Bar
    | Qux
```
And in the language I'm doing in my free time, i put `;` before definitions. That's just I.

lib/src/Dscp/Core/Types.hs
Heimdell: Ok

lib/src/Dscp/Snowdrop/AccountValidation.hs
volhovm: Please try to stick to the suggested coding style though, `hindent` does its job pretty well.

PR #65, opened 2018-07-23
[DSCP-157] Unite witness'es and educator's secret keys

Martoon-00:
Use common code for keyfile of witness and educator.

I inevitably had to make them distinguishable in nodes context and saw two approaches to do that:
1. `instance HasLens (KeyResourcesTag node) ctx KeyResources`
2. `instance HasLens' ctx (KeyResources node)`, thus adding a phantom type to `KeyResources` itself.
Eventually I chose the latter approach as soon as it prevents passing `KeyResources` wrong around. 

In many places I had to neglect your design decisions, @volhovm, because other ones has already been defined by @flyingleafe in https://issues.serokell.io/issue/DSCP-118, although you're the lead man now so feel free to override those, I'll fix :trollface:


lib/src/Dscp/CLI/Common.hs
```diff
@@ -70,16 +74,26 @@ versionOption = infoOption ("disciplina-" <> (showVersion version)) $
...
+         long [qc|{who}-keyfile-path|] <>
+         metavar "FILEPATH" <>
+         help [qc|Path to the secret key of {who}.|]
+    kpGenKeyParser = switch $
+         long [qc|{who}-generate-new-key|] <>
+         help [qc|Generate the key and write it to '{who}-keyfile-path' path.|]
+    kpPassphraseParser = option passphraseReadM $
+         long [qc|{who}-keyfile-password|] <>
+         metavar "PASSWORD" <>
+         help "Password from secret key."
```
flyingleafe: Looks grammatically invalid, maybe `Password _for_ secret key`?

lib/src/Dscp/Resource/Keys/Types.hs
```diff
@@ -0,0 +1,88 @@
...
+data KeyJson = KeyJson
+    { kjEncSecretKey :: Encrypted SecretKey
+    } deriving (Eq, Show)
+
+-- | What exactly lies in the store.
+type KeyfileContent = Versioned KeyJson
+
+instance ToJSON KeyJson where
+    toJSON KeyJson{..} = object
+        [ "educator_secret" .= kjEncSecretKey
```
flyingleafe: Ha, gotcha! (rename this to simply `secret`)

lib/src/Dscp/Educator/Launcher/Runner.hs
```diff
@@ -15,7 +15,7 @@ import Dscp.Witness.Launcher.Runner (formWitnessContext)
 formEducatorContext :: EducatorParams -> EducatorResources -> IO EducatorContext
 formEducatorContext _ecParams EducatorResources{..} = do
     let _ecDB = _erDB
-    let _ecSecret = _erSecret
+    let _ecKeys = _erKeys
```
Heimdell: It is too near to `let x = x`.

I'm still against `RecordWildCards`.

Not only it moves assignments away from records that receives them, but it also makes it possible to create `forall a. a` object (which is impossible with normal record assignment).

And, also. The code is supposed to be read by people, so lets not fill it with "connect property to corresponding record" riddles.

lib/src/Dscp/Resource/Keys/Error.hs
```diff
@@ -0,0 +1,37 @@
...
+
+import Dscp.Crypto (DecryptionError)
+import Dscp.Util (wrapRethrow)
+
+-- | Exception during secret key extraction from storage.
+data KeyInitError
+    = SecretWrongPassPhraseError !DecryptionError
+    | SecretParseError !Text
+    | SecretFileExistsError !FilePath
+    | SecretIOError !Text
```
Heimdell: `{-# LANGUAGE StrictData #-}`?

lib/src/Dscp/Resource/Keys/Functions.hs
```diff
@@ -0,0 +1,127 @@
...
+module Dscp.Resource.Keys.Functions
+    ( linkStore
+    ) where
+
+import Data.Aeson (eitherDecode', encode)
+import qualified Data.ByteString.Lazy as LBS
+import Fmt ((+||), (|+), (||+))
+import Loot.Log (MonadLogging, logDebug, logInfo)
+import System.Directory (doesFileExist)
+import System.Directory as D
```
Heimdell: First line 
```haskell
import System.Directory (doesFileExist)
```
does nothing, as second one imports `doesFileExist` in both qualified and unqualified modes.

lib/src/Dscp/CLI/Common.hs
Martoon-00: Yeah, sounds better

lib/src/Dscp/Resource/Keys/Functions.hs
Martoon-00: Pff, right :man_facepalming:

lib/src/Dscp/Resource/Keys/Error.hs
Martoon-00: Cool, didn't know about it

lib/src/Dscp/Educator/Launcher/Runner.hs
Martoon-00: Okay, here it really would look better without such `let`s

lib/src/Dscp/CLI/Common.hs
```diff
@@ -110,9 +124,9 @@ netServParamsParser = NetServParams <$> peersParser <*> ourZTNodeIdParser
 ---------------------------------------------------------------------------
 
 parseNetAddr :: String -> Either String NetworkAddress
-parseNetAddr str =
-    let host = toText $ takeWhile (/= ':') str
-        portS = drop (length host + 1) str
+parseNetAddr st =
```
volhovm: Such change.

lib/src/Dscp/Educator/Launcher/Mode.hs
```diff
@@ -66,11 +67,11 @@ type CombinedWorkMode ctx m =
 -- TODO Separate resources and non-resources.
 data EducatorContext = EducatorContext
     {
-      _ecParams     :: EducatorParams
-    , _ecDB         :: SQLiteDB
-    , _ecSecret     :: EducatorSecret
+      _ecParams     :: !EducatorParams
+    , _ecDB         :: !SQLiteDB
+    , _ecKeys       :: !(KeyResources EducatorNode)
```
volhovm: I wanted to change it many times before but tried to avoid conflicts, now is the time. Please take a look at witness mode context -- it essentially consists of three things: parameters, resources and variables. SQLiteDB, KeyResources are Resources, so please put EducatorResources directly into EducatorContext and fix related HasLens instances.

lib/src/Dscp/CLI/Common.hs
Martoon-00: It conflicted with `str` from `opt-parse` :man_shrugging:

PR #67, opened 2018-07-24
[DSCP-151] Address late comments on Student API stubs

nasportivnoy:
PR consists of parseNetAddr (CLI/Common) reimplementation using parsec and haddock for APIError addition


lib/src/Dscp/CLI/Common.hs
```diff
@@ -32,6 +32,10 @@ import Dscp.Resource.Network (NetCliParams (..), NetServParams (..))
 import Dscp.Web (NetworkAddress (..))
 import Paths_disciplina (version)
 
+import qualified Text.Parsec.String as Parsec
```
volhovm: This should not be a separate section, but rather united with a first one (project-independent).

PR #69, opened 2018-07-24
[AD-313] Make secret key encryption optional

goodsoft:
Secret key is now encrypted/decrypted only if passphrase is provided as optional parameter.

### DIFF
--- a/wallet/src/Dscp/Wallet/Knit.hs
+++ b/wallet/src/Dscp/Wallet/Knit.hs
@@ -2,13 +2,13 @@ module Dscp.Wallet.Knit where
 
 import Prelude hiding (preview)
 
-import Codec.Serialise (deserialiseOrFail)
+import Codec.Serialise (Serialise, deserialiseOrFail, serialise)
 import Control.Exception
 import Control.Lens
 import Data.Char (isSpace)
 import Data.Scientific (toBoundedInteger)
 import Dscp.Core (TxOut(..), addrFromText, coinToInteger, txId)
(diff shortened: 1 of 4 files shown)


wallet/src/Dscp/Wallet/Knit.hs
```diff
@@ -153,16 +155,25 @@ instance AllConstrained (Elem components) '[Wallet, Core] => ComponentCommandPro
...
-            pure (passPhrase, secretKey, outs)
-        , cpRepr = \(passPhrase, secretKey, outs) -> CommandAction $ \WalletFace{..} -> do
-            pp <- either throwIO return . mkPassPhrase . encodeUtf8 $ passPhrase
-            sk <- either throwIO return . deserialiseOrFail $ BSL.fromStrict secretKey
-            toValue . ValueString . show . txId <$> walletSendTx pp sk outs
+            pure (passString, secretKeyString, outs)
+        , cpRepr = \(passString, secretKeyString, outs) -> CommandAction $ \WalletFace{..} -> do
+            let
+              decodeSK :: Serialise a => ByteString -> IO a
+              decodeSK = either throwIO return . deserialiseOrFail . BSL.fromStrict
```
volhovm: I don't quite understand why `serialise` is used here at all. You should parse the secret key from the string instead of deserialising. It can be done with `fromByteArray`.

goodsoft: You're right.
`Encoded SecretKey` is still `serialise`d, though

PR #70, opened 2018-07-24
[DSCP-152] Wallet backend impementation

Martoon-00:
Implements simple wallet backend.

- [x] Implement backend
- [x] Add servant server for backend
- [x] Add servant client requests
- [x] Copy-paste code of servant requests logging
- [ ] ~Add mempool support~ separate PR
- [ ] ~Show tx list in account state~ in unforeseen future

### DIFF
--- a/.stylish-haskell.yaml
+++ b/.stylish-haskell.yaml
@@ -237,6 +237,7 @@ language_extensions:
   - NamedFieldPuns
   - NoImplicitPrelude
   - OverloadedStrings
+  - PolyKinds
   - RecordWildCards
   - ScopedTypeVariables
   - StandaloneDeriving
--- a/core/src/Dscp/Core/Address.hs
+++ b/core/src/Dscp/Core/Address.hs
@@ -6,6 +6,7 @@ module Dscp.Core.Address
        , addrToBase58
        , addrFromBase58
        , addrFromText
+       , addrToText
        ) where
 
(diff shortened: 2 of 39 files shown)


goodsoft:
I thought a bit about point 1, and I think the server part would fit perfectly in `Dscp.Web.Wallet` (instead of current `Dscp.Wallet.Web`)

Martoon-00:
Could you explain such choice?

goodsoft:
Because if there are multiple HTTP-endpoints in Disciplina, they could all fit under `Dscp.Web` namespace along with all required `Types` and all.

Martoon-00:
Both points fixed

witness/src/Dscp/Witness/Web/Wallet/API.hs
```diff
@@ -0,0 +1,37 @@
...
+import Servant.Generic
+
+import Dscp.Core.Types
+import Dscp.Witness.Web.Wallet.Types
+
+-- | API/implementation of wallet endpoints.
+data WalletEndpoints route = WalletEndpoints
+    { -- shortened prefix for purpose, one will most probably use type aliases
+      -- refering this type
+      wGetNextNonce :: route
```
volhovm: Why not return the nonce with the account state?

Martoon-00: Oki, fixed

PR #72, opened 2018-07-25
[DSCP-158] Split lib into core/ witness/ educator/

volhovm:


core/tests/Test/Dscp/Core/Instances.hs
```diff
@@ -0,0 +1,3 @@
+module Test.Dscp.Core.Instances
+    (
+    ) where
```
Martoon-00: Memno

volhovm: Fixed.

educator/src/Dscp/Educator/Secret/Class.hs
```diff
@@ -0,0 +1,20 @@
...
+import Dscp.Educator.Secret.Types (EducatorSecret (..))
+import Dscp.Launcher.Rio (RIO)
+
+class MonadEducatorSecret m where
+    -- | Fetch secret key of educator.
+    getEducatorSecret :: m SecretKey
+
+instance HasLens' ctx EducatorSecret =>
+         MonadEducatorSecret (RIO ctx) where
+    getEducatorSecret = views (lensOf @EducatorSecret) esSecretKey
```
Martoon-00: Everything under `Dscp.Educator.Secret` is redundant now

volhovm: Fixed.

PR #77, opened 2018-07-27
[AD-310] Ariadne Vty wallet

goodsoft:
`stack exec disciplina-wallet-vty -- --wallet-server 127.0.0.1:4020`

- [x] Copy Vty library from Ariadne
- [x] Make it compile
- [x] Show list of accounts
- [x] Selecting account from the list
- [x] Add new account form
- [x] Balance info
- [x] Send transaction form

### DIFF
--- a/wallet/app-vty/Glue.hs
+++ b/wallet/app-vty/Glue.hs
@@ -1,0 +1,287 @@
+-- | Glue code between the frontend and the backends.
+
+module Glue
+       (
+         -- * Knit ↔ Vty
+         knitFaceToUI
+
+         -- * Wallet ↔ Vty
+       , createWalletState
+       , walletEventToUI
+       , putWalletEventToUI
+
+         -- * Command history ↔ Vty
+       , historyToUI
+       ) where
+
+import Control.Exception (displayException)
+import Control.Lens (ix)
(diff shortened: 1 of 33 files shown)


wallet/app-vty/Glue.hs
```diff
@@ -0,0 +1,287 @@
...
+          )
+      UiRestoreWallet UiRestoreWalletArgs{..} -> do
+        Right $ Knit.ExprProcCall
+          (Knit.ProcCall "restore-key" $
+            [ Knit.ArgKw "secretkey" . Knit.ExprLit . Knit.toLit . Knit.LitString $ urwaMnemonic
+            ] ++
+            optString "name" urwaName ++
+            optString "pass" urwaPassphrase
+          )
+      _ -> Left "Not implemented yet"
```
volhovm: By the way, which commands are missing now?

goodsoft: New account, new address (for HD-wallets), rename, remove...

volhovm: Ah ok, these constructors are imported (copypasted) from ariadne i guess?

goodsoft: Yes.

PR #87, opened 2018-08-03
[DSCP-113] Educator api server with stubs

Martoon-00:


educator/src/Dscp/Educator/Web/Educator/API.hs
```diff
@@ -0,0 +1,230 @@
...
+        :- "students" :> Capture "studentAddr" Student
+        :> "courses"  :> Capture "courseId" Course
+        :> "proofs"
+        :> Summary "Get proofs of student's course progress"
+        :> Description "Gets student's course grades and submissions in form \
+                        \of private transactions, together with corresponding \
+                        \Merkle proofs."
+        :> Get '[JSON] [BlkProofInfo]
+
+   } deriving (Generic)
```
Heimdell: This looks nice; but is it possible to abstract "REST entity" of some kind?
I don't insist on doing this, though.

educator/src/Dscp/Educator/Web/Educator/Error.hs
```diff
@@ -0,0 +1,54 @@
...
+instance ToJSON APIError where
+    toJSON = error "Not implemented"
+
+---------------------------------------------------------------------------
+-- Functions
+---------------------------------------------------------------------------
+
+-- | Get HTTP error code of error.
+toServantErrNoReason :: APIError -> ServantErr
+toServantErrNoReason = error "Not implemented"
```
Heimdell: Why not 500?

educator/src/Dscp/Educator/Web/Educator/Types.hs
```diff
@@ -0,0 +1,93 @@
...
+    , aiDesc         :: !Text
+    } deriving (Show, Eq, Generic)
+
+data SubmissionEducatorInfo = SubmissionEducatorInfo
+    { siHash           :: !(Hash Submission)
+    , siContentsHash   :: !(Hash Raw)
+    , siAssignmentHash :: !(Hash Assignment)
+    , siGrade          :: !(Maybe GradeInfo)
+    , siWitness        :: !SubmissionWitness
+    } deriving (Show, Eq, Generic)
```
Heimdell: I recommend `{-# LANGUAGE StrictData #-}`.

educator/src/Dscp/Educator/Web/Server.hs
```diff
@@ -1,49 +1,58 @@
...
-import Dscp.Launcher.Rio (runRIO)
+import Dscp.Educator.Web.Student (StudentAPI, convertStudentApiHandler, studentAPI,
+                                  studentApiHandlers)
 import Dscp.Web (ServerParams (..), serveWeb)
 
-studentAPIServer
+type EducatorWebAPI = "api" :> (
+    "educator" :> EducatorAPI
+    :<|>
+    "student" :> StudentAPI
```
Heimdell: No `"common"`?

educator/src/Dscp/Educator/Web/Server.hs
```diff
@@ -1,49 +1,58 @@
...
     :: forall ctx m. EducatorWorkMode ctx m
     => (forall x. m x -> Handler x)
     -> Bool
     -> m (Server StudentAPI)
-studentAPIServer nat withBot = do
-    handlers <-
+mkStudentApiServer nat withBot = do
+    studentAndBotHandlers <-
         bool identity (>>= addBotHandlers) withBot
         (pure studentApiHandlers)
```
Heimdell: What is going on here?

educator/src/Dscp/Educator/Web/Student/Queries.hs
```diff
@@ -80,21 +78,21 @@ type MonadStudentAPIQuery m =
...
-    => Id Student -> Core.Course -> m Course
-getCourse studentId courseId =
+    => Id Student -> Course -> m CourseStudentInfo
+studentGetCourse studentId courseId =
     transaction $ do
         mcourse <-
             query queryText (Only courseId)
             >>= listToMaybeWarn "courses"
         Only mdesc <-
             pure mcourse `assertJust` CourseDoesNotExist courseId
```
Heimdell: Where does the warning go? Into logs?

educator/src/Dscp/Educator/Web/Student/Queries.hs
```diff
@@ -172,18 +170,18 @@ getLastAssignmentSubmission student assignH = do
...
         >>= listToMaybeWarn "assignments"
-    (aCourseId, aContentsHash, assignType, aDesc) <-
+    (aiCourseId, aiContentsHash, assignType, aiDesc) <-
         pure massign `assertJust` AssignmentDoesNotExist assignH
-    let IsFinal aIsFinal = assignType ^. assignmentTypeRaw
-    aLastSubmission <- getLastAssignmentSubmission student assignH
-    return Assignment{ aHash = assignH, .. }
+    let IsFinal aiIsFinal = assignType ^. assignmentTypeRaw
+    aiLastSubmission <- studentGetLastAssignmentSubmission student assignH
+    return AssignmentStudentInfo{ aiHash = assignH, .. }
```
Heimdell: Please, don't use `RecordWildCards` this way.
The data flow becomes very obscure with it. (You can rename `assignH` to `aiHash` to improve obscurity)

Also, there is `Id` type family from `Dscp.Util.HasId` typeclass; please use it when referring to entities, don't use their hashes. The `Id a` is the `Hash a` for most our types.

if there is no instance for some new entity, please make one.

It even has sane instances for all good old types once lived in `Dscp.Core.Types`.

educator/src/Dscp/Educator/Web/Student/Types.hs
```diff
@@ -1,178 +1,119 @@
...
+import Servant (FromHttpApiData)
 
-makePrisms ''IsFinal
+import Dscp.Core
+import Dscp.Crypto
+import Dscp.Educator.Web.Types
+import Dscp.Util (Id)
 
 -- | Whether student is enrolled into a course.
 newtype IsEnrolled = IsEnrolled { unIsEnrolled :: Bool }
```
Heimdell: Do we need this type? We protected from mixing it with any other `Bool`, but is it worth it?

educator/src/Dscp/Educator/Web/Types.hs
```diff
@@ -0,0 +1,79 @@
...
+
+import Dscp.Core
+import Dscp.Crypto
+import Dscp.Util.Aeson (AsHex)
+
+-- | Whether assignment is final in course.
+newtype IsFinal = IsFinal { unIsFinal :: Bool }
+    deriving (Eq, Show)
+
+makePrisms ''IsFinal
```
Heimdell: Leaving the nessessity of wrapper out of scope, do we need that prism? We can unwrap it in optic chain (and then add `._Just` because thats how prisms work) and have our `Bool`, yes.

But I thought that prisms are best for distinguishing constructors.

Feel free to prove me wrong. Maybe I'm missing something here.

P.S.: why not making it `data IsFinal = IsFinal | IsNotFinal` (and adding a lens to `Bool`?), if you want to have type that is not `Bool`?

educator/src/Dscp/Educator/Web/Types.hs
```diff
@@ -0,0 +1,79 @@
...
+
+assignmentTypeRaw :: Iso' AssignmentType IsFinal
+assignmentTypeRaw = iso forth back
+  where
+    back = \case
+        IsFinal False -> Regular
+        IsFinal True  -> CourseFinal
+    forth = \case
+        Regular     -> IsFinal False
+        CourseFinal -> IsFinal True
```
Heimdell: I started loosing track of new `Bool` types we have. Even if `AssignmentType` may become 3-type at some moment.

educator/tests/Test/Dscp/Educator/Web/Student/Queries.hs
```diff
@@ -157,27 +156,27 @@ spec_StudentApiQueries = describe "Basic database operations" $ do
...
     describe "getCourses" $ do
         it "Student is not enrolled initially" $
             sqliteProperty $ \() -> do
                 _ <- CoreDB.createStudent student1
                 _ <- createCourseSimple 1
 
-                courses <- sqlTx $ DB.getCourses student1 Nothing
-                return (all (not . cIsEnrolled) courses)
+                courses <- sqlTx $ studentGetCourses student1 Nothing
+                return (all (not . ciIsEnrolled) courses)
```
Heimdell: May be expressed as 
```
return $ not $ any ciIsEnrolled courses
```
To reduce amount of parenthes.

I don't insist here.

educator/tests/Test/Dscp/Educator/Web/Student/Queries.hs
```diff
@@ -335,16 +334,16 @@ spec_StudentApiQueries = describe "Basic database operations" $ do
...
 
                 let assignments' =
-                        applyFilterOn aCourseId courseIdF $
-                        applyFilterOn aDocumentType docTypeF $
-                        applyFilterOn (IsFinal . aIsFinal) isFinalF $
-                        map (\a -> liftAssignment a Nothing)
+                        applyFilterOn aiCourseId courseIdF $
+                        applyFilterOn saDocumentType docTypeF $
+                        applyFilterOn (IsFinal . aiIsFinal) isFinalF $
+                        map (\a -> studentLiftAssignment a Nothing)
```
Heimdell: Can be right section:
```
map (`studentLiftAssignment` Nothing)
```
but I don't insist.

educator/src/Dscp/Educator/Web/Student/Queries.hs
Heimdell: The reason the dataflow is obscure is that you meet `FooBar { .. }` and the flow of reading breaks: you have to go up in the text to collect everything that looks like `fbQux`.

You can't reconstruct data going in in this line just by looking on it.

educator/src/Dscp/Educator/Web/Educator/API.hs
Martoon-00: What do you understand under `REST entity`?

educator/src/Dscp/Educator/Web/Educator/Error.hs
Martoon-00: Will be along with implementation, it was lazier to remove copy-pasted `err500` than to import it )

educator/src/Dscp/Educator/Web/Server.hs
Martoon-00: Waiting for @volhovm to create some ticket about it

educator/src/Dscp/Educator/Web/Server.hs
Martoon-00: I wanted it to look like a pipeline so that it was easy to build handlers from many pieces, one piece per line

educator/src/Dscp/Educator/Web/Student/Queries.hs
Martoon-00: Yep

educator/src/Dscp/Educator/Web/Student/Queries.hs
Martoon-00: But the point of this function and all others in this module is entirely to construct a response, seems to be justified use of `RecordWildCards`

educator/src/Dscp/Educator/Web/Student/Types.hs
Martoon-00: I feel bad when I see function of type `Bool -> Bool -> Bool -> m Something` (same endpoints would look like that), and making mistake here is easy and difficult to find, so I think it's worth it.

Martoon-00: Instead of creating newtype I can switch to `named` library, if it would look better

educator/src/Dscp/Educator/Web/Types.hs
Martoon-00: I also think prisms are, but `IsFinal` turns to be a stronger `Iso` here, so it works for wrapping and unwrapping safely.

> why not making it data IsFinal = IsFinal | IsNotFinal (and adding a lens to Bool?)
This way I would need to define custom `ToJSON / FromJSON / FromHttpApiData` instances for servant. Not saying that's is hard, I just found a balance between not operating with blind bools and transparency for servant library.

educator/tests/Test/Dscp/Educator/Web/Student/Queries.hs
Martoon-00: Meh, right

Martoon-00:
I actually forgot about `Id`, would be really nice to use it.

educator/src/Dscp/Educator/Web/Educator/API.hs
Heimdell: `POST /entity`
`GET /entity/`  -- get all
`GET /entity/:id`
`PUT /entity/:id`
`DELETE /entity/:id`

As one template or type in our case.

educator/src/Dscp/Educator/Web/Student/Queries.hs
Heimdell: I'm personally against usage of this extension. But don't insist here.

educator/src/Dscp/Educator/Web/Student/Types.hs
Heimdell: In that case, usage of `newtype`s is justified.

educator/src/Dscp/Educator/Web/Types.hs
Heimdell: Then its OK.

educator/src/Dscp/Educator/Web/Educator/API.hs
Martoon-00: Well, looks like each entity would have different set of endpoints corresponding to it (most have only all-getter, while many have no one-getter at all), so it's hard to extract such template.
I can visually group endpoints for each entity if you find that useful.

Martoon-00:
I seem to address major concerns. Do I deserve a green tick?

Martoon-00:
Btw I thought a bit about `s/Hash/Id`, if I do this change, then I should change field names as well: `submissionHash` -> `submissionId`. Then JSON serialisation will change, I will have to adjust swagger spec. @flyingleafe was working on it, and it is something I don't want to change in his absence.

educator/src/Dscp/Educator/Web/Server.hs
volhovm: Let it be separate. It doesn't matter that much.

educator/src/Dscp/Educator/Web/Student/Types.hs
volhovm: It's alright to use newtypes here imho. Especially if this type is a part of API.

PR #93, opened 2018-08-06
[DSCP-141] Add HTTP request signature verification

hithroc:
This PR adds authentication to all students API requests.

The client is supposed to make a signed JWT (http://jwt.io/) and put it into Authorization HTTP header. The JWT should contain only one claim called `dat` and inside it there should be an object with two fields `adTime` and `adPath`. `adTime` should have the time when the request is created. The JWT is considered invalid by the server if server's current time is 5 minutes (or more) ahead of the specified time in JWT (this is to avoid replay attacks). `adPath` should contain the URL path without host as it will be in the raw HTTP request. Example JWT object:
```
{ "dat": {"adTime": "2018-08-10T13:26:59.461998136Z", "adPath":"/api/student/v1/courses"} }
```
Here `dat` is an unregistered JWT claim and the object in it the authentication data that server uses.
The signed version of JWT above should look something like this
```
eyJhbGciOiJFZERTQSJ9.eyJkYXQiOnsiYWRQYXRoIjoiL2FwaS9zdHVkZW50L3YxL2NvdXJzZXMiLCJhZFRpbWUiOiIyMDE4LTA4LTEwVDEzOjI2OjU5LjQ2MTk5ODEzNloifX0.UWxBudf2Jckrf0PbFIGQ9I5fbFHy6DZUyxCl1F1rCf6WLop3yaT8-0c9rhoK7HDN0FRPdCkFnoU0eQbhlSa8BQ
```
After JWT is signed, the client should put it into `Authorization` HTTP Header with `Bearer` prefixed. Here's an example:
```
Authorization: Bearer eyJhbGciOiJFZERTQSJ9.eyJkYXQiOnsiYWRQYXRoIjoiL2FwaS9zdHVkZW50L3YxL2NvdXJzZXMiLCJhZFRpbWUiOiIyMDE4LTA4LTEwVDEzOjI2OjU5LjQ2MTk5ODEzNloifX0.UWxBudf2Jckrf0PbFIGQ9I5fbFHy6DZUyxCl1F1rCf6WLop3yaT8-0c9rhoK7HDN0FRPdCkFnoU0eQbhlSa8BQ"
```
And in `curl`
```
curl -v -H "Authorization: Bearer eyJhbGciOiJFZERTQSJ9.eyJkYXQiOnsiYWRQYXRoIjoiL2FwaS9zdHVkZW50L3YxL2NvdXJzZXMiLCJhZFRpbWUiOiIyMDE4LTA4LTEwVDEzOjI2OjU5LjQ2MTk5ODEzNloifX0.UWxBudf2Jckrf0PbFIGQ9I5fbFHy6DZUyxCl1F1rCf6WLop3yaT8-0c9rhoK7HDN0FRPdCkFnoU0eQbhlSa8BQ" 127.0.0.1:8090/api/student/v1/courses
```

Currently the server code doesn't actually use any sort of database of public key for authentication, and instead uses two dummy student keys. So to authorize your requests you have to use one of those keys:
```
{"encSecretKey":"UMIfCuZeHdVU8cbHwKvKAS1QigqZzU59UbLScT7NTHCu6lggKjLQ5Wqkw8NR51UmkMcxnjxLdz2GNmihDiIxEuzNAAw=","secretKey":"bbse7rPo6BVH8we6k0Ta5ihqkxlFjpWW8v/ENnDVxkg=","publicKey":"836d6a5908166c398db194e7af916827f63da021a6a3a1c3edf9ced6608adf2b","address":"LL4qKqLVPqrkx1ekKfQJo1uATwtQqVEXmdNanYc4BbgydaCuawVcChdf"}
```
```
{"encSecretKey":"UG3jdfg8xsu5WNSEXTWeLO9QigqZzU59UbLScT7NTHCu6lggNjPwrwA6C6Wm4chMUOUxgs8iUqrw1aeDVj+hM6PDSjU=","secretKey":"cbo+pNl2IHOw9ZrQU2ba+tsDto4zbVq0quJUFz/bjHE=","publicKey":"ed21126e18b5a074a06b3e1ba700c9bfb7aa56ebb4f9cde8735717b72811967f","address":"LL4qKnCgFHcaJEY7ivdkoaFLNw1B84nFiye8iykJH2WodrgyxnPRiJmi"}
```

### DIFF
--- a/educator/src/Dscp/Educator/Web/Server.hs
+++ b/educator/src/Dscp/Educator/Web/Server.hs
@@ -6,26 +6,33 @@ module Dscp.Educator.Web.Server
        ( serveStudentAPIReal
        ) where
 
+import Data.Proxy (Proxy (..))
 import Fmt ((+|), (|+))
 import Loot.Log (logInfo)
 import Network.Wai.Middleware.Cors (simpleCors)
-import Servant ((:<|>) (..), Handler, Server, hoistServer, serve)
+import Servant ((:<|>) (..), Context (..), Handler, Server, err401, hoistServer,
+                hoistServerWithContext, serveWithContext)
+import Servant.Auth.Server (AuthResult (..), CookieSettings, JWTSettings, defaultCookieSettings,
+                            defaultJWTSettings, generateKey, throwAll)
 import Servant.Generic (toServant)
 
+import Dscp.Core (Student)
+import Dscp.Crypto (PublicKey)
 import Dscp.Educator.Config (HasEducatorConfig)
 import Dscp.Educator.Launcher (EducatorRealMode, EducatorWorkMode)
-import Dscp.Educator.Web.Bot (EducatorBotSwitch (..), addBotHandlers)
+import Dscp.Educator.Web.Bot (EducatorBotSwitch (..), addBotHandlers, initializeBot)
 import Dscp.Educator.Web.Educator (EducatorAPI, convertEducatorApiHandler, educatorAPI,
                                    educatorApiHandlers)
 import Dscp.Educator.Web.Params (EducatorWebParams (..))
-import Dscp.Educator.Web.Student (StudentAPI, convertStudentApiHandler, studentAPI,
+import Dscp.Educator.Web.Student (GetStudentsAction (..), ProtectedStudentAPI, StudentAPI,
+                                  WithStudent (..), convertStudentApiHandler, studentAPI,
                                   studentApiHandlers)
 import Dscp.Web (ServerParams (..), serveWeb)
 
 type EducatorWebAPI =
     EducatorAPI
     :<|>
-    StudentAPI
+    ProtectedStudentAPI
 
 mkEducatorApiServer
     :: forall ctx m. EducatorWorkMode ctx m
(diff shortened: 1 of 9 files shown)


educator/src/Dscp/Educator/Web/Student/Auth.hs
```diff
@@ -0,0 +1,74 @@
...
+
+-- | A type that has a Student attached to it
+data WithStudent a = WithStudent Student a
+
+-- | A type that the student has sent to authenthicate
+data AuthData = AuthData
+  deriving Generic
+
+-- | Action to get all students
+type GetStudents = IO [PublicKey]
```
volhovm: newtype

educator/src/Dscp/Educator/Web/Student/Auth.hs
```diff
@@ -0,0 +1,74 @@
...
+---------------------------------------------------------------------------
+
+-- | Custom authenthication type
+data StudentAuth
+
+-- | A type that has a Student attached to it
+data WithStudent a = WithStudent Student a
+
+-- | A type that the student has sent to authenthicate
+data AuthData = AuthData
```
volhovm: So we should figure out what is it exactly, right?

hithroc: Yes.

hithroc: Also I probably should change this comment because right now it doesn't make sense 😅

educator/src/Dscp/Educator/Web/Student/Auth.hs
```diff
@@ -0,0 +1,74 @@
...
+       , WithStudent (..)
+       , AuthData (..)
+       , GetStudents
+       ) where
+
+import Crypto.JOSE.JWK hiding (hash)
+import Crypto.JWT hiding (hash)
+import Data.Aeson
+import Servant.Auth.Server
+import Servant.Auth.Server.Internal.Class
```
Heimdell: Please, don't import modules outside of the project like that, _especially_ `Internal` ones.
If possible, don't import `Internal` modules at all.
If you import, either put a list of names to import or do it in qualified way.

educator/src/Dscp/Educator/Web/Student/Auth.hs
```diff
@@ -0,0 +1,74 @@
...
+---------------------------------------------------------------------------
+
+-- | This function returns AuthCheck that checks the signature of the JWT.
+studentAuthCheck :: FromJWT usr => GetStudents -> AuthCheck (WithStudent usr)
+studentAuthCheck getStudents = do
+    students <- liftIO $ getStudents
+    let pubKeyToJwk (AbstractPK pub) =
+          fromKeyMaterial (OKPKeyMaterial $ Ed25519Key pub Nothing) & jwkKeyOps .~ Just [Verify]
+    let jwkPbs = map (\pub -> (pub, pubKeyToJwk pub)) students
+    usr <- foldr (\(p, j) a -> ((p,) <$> jwtAuthCheck (defaultJWTSettings j)) <|> a) empty jwkPbs
```
Heimdell: I suggest making a local function out of that (with properly named parameters).

Heimdell: So, it will be like
```haskell
usr <- foldr authOrTryNext empty jwkPbs
```

educator/src/Dscp/Educator/Web/Student/Handlers.hs
```diff
@@ -24,17 +23,14 @@ type StudentApiWorkMode m =
...
+-- TODO [DSCP-141]: remove this
 oneGeekSK :: SecretKey
 oneGeekSK = studentSKEx
 
 studentApiHandlers
     :: forall m. StudentApiWorkMode m
-    => StudentApiHandlers m
-studentApiHandlers =
+    => Student -> StudentApiHandlers m
+studentApiHandlers oneGeek = traceShow oneGeek $
```
Heimdell: Please, don't call students "geeks", some can become offended after they read the code. This is open source, after all.

educator/src/Dscp/Educator/Web/Server.hs
```diff
@@ -38,23 +45,54 @@ mkStudentApiServer
...
+        ppub = AbstractPK pub
+        ppub2 = AbstractPK pub2
+    let jwkSk = fromKeyMaterial (OKPKeyMaterial $ Ed25519Key pub $ Just sk) & jwkKeyOps .~ Just [Sign]
+    let jwkSk2 = fromKeyMaterial (OKPKeyMaterial $ Ed25519Key pub2 $ Just sk2) & jwkKeyOps .~ Just [Sign]
+    atomically $ modifyTVar' tvr (ppub :)
+    atomically $ modifyTVar' tvr (ppub2 :)
+    liftIO $ print $ hash ppub
+    liftIO $ print =<< (makeJWT AuthData (defaultJWTSettings jwkSk) Nothing)
+    liftIO $ print $ hash ppub2
+    liftIO $ print =<< (makeJWT AuthData (defaultJWTSettings jwkSk2) Nothing)
```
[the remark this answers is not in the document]
hithroc: This is just a quick test code to check that everything indeed works (since it's still WIP), this part will be completely removed in the final PR.

educator/src/Dscp/Educator/Web/Student/Auth.hs
hithroc: This type represents an action to get the list of students (or rather, their public keys). That's why I thought `GetStudents` name is fitting. I can make it `GetStudentsAction` I guess.

hithroc: I don't really see a difference between `IO` action and any other value, why is it a big deal to not have it behind a type?

I'm using an `IO` action here, instead of something like `TVar [PublicKey]` because that allows the auth checker to not depend or even care about the implementation details on how it is actually going to get the public keys (is it TVar? Is it a database connection? etc), it just executes the action and gets the thing. All implementation details are left to whoever provides the servant context.

educator/src/Dscp/Educator/Web/Student/Handlers.hs
Martoon-00: Loool, it just was temporal constant so I called it funny, now you'd better to replace it with something plain ))

educator/src/Dscp/Educator/Web/Student/Auth.hs
volhovm: Yeah, I didn't leave a comment on it, but unqualified imports from external libraries are considered a bad practice, so please spend several minutes to make them qualified.

educator/src/Dscp/Educator/Web/Student/Auth.hs
hithroc: Yeah, it will be a `newtype` after I update and finalize the PR.

PR #98, opened 2018-08-07
[DSCP-172] Swagger spec for Block Explorer API

goodsoft:
First draft

### DIFF
--- a/specs/disciplina/witness/api/explorer.yaml
+++ b/specs/disciplina/witness/api/explorer.yaml
@@ -1,0 +1,178 @@
+swagger: "2.0"
+info:
+  description: "This is an HTTP API for interacting with Disciplina Witness node. It is used by Block Explorer frontend to request info about blocks, addresses and transactions."
+  version: 0.1.0
+  title: Disciplina Block Explorer API
+  termsOfService: "https://disciplina.io/tnc.pdf"
+  contact:
+    email: hi@serokell.io
+  license:
+    name: Apache 2.0
+    url: "http://www.apache.org/licenses/LICENSE-2.0.html"
+host: "localhost:8091"
+basePath: /api/witness/v1
+schemes:
+  - https
+  - http
+paths:
+  /blocks:
+    get:
+      summary: Get list of blocks (paginated).
+      operationId: getBlocks
+      produces:
+      - application/json
+      parameters:
+      - in: query
+        name: before
+        description: Get only blocks before (not including) the specified one. If not specified, get the most recent blocks.
+        required: false
+        type: string
+        format: byte
+      responses:
+        200:
+          description: Successful operation
+          schema:
+            type: array
+            items:
+              $ref: "#/definitions/BlockInfo"
+        404:
+          description: Block not found
+  /blocks/{headerHash}:
+    get:
+      summary: Get information about a block.
+      operationId: getBlock
+      produces: 
+      - application/json
+      parameters:
+      - in: path
+        name: headerHash
+        description: Block identifier (hash of the block header).
+        required: true
+        type: string
+        format: byte
+      responses:
+        200:
+          description: Successful operation
+          schema:
+            $ref: "#/definitions/BlockInfo"
+        404:
+          description: Block not found
+  /addresses/{address}:
+    get:
+      summary: Get information about an address.
+      operationId: getAddress
+      produces: 
+      - application/json
+      parameters:
+      - in: path
+        name: address
+        description: Address being requested.
+        required: true
+        type: string
+        format: byte
+      responses:
+        200:
+          description: Successful operation
+          schema:
+            $ref: "#/definitions/AddressInfo"
+        404:
+          description: Address not found
+  /transactions/{transactionHash}:
+    get:
+      summary: Get information about a transaction.
+      operationId: getTransaction
+      produces: 
+      - application/json
+      parameters:
+      - in: path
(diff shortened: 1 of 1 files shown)


specs/disciplina/witness/api/explorer.yaml
```diff
@@ -0,0 +1,178 @@
...
+      responses:
+        200:
+          description: Successful operation
+          schema:
+            $ref: "#/definitions/AddressInfo"
+        404:
+          description: Address not found
+  /transactions/{transactionHash}:
+    get:
+      summary: Get information about a transaction.
```
Martoon-00: I think it's worth being specific on which transactions would be returned

specs/disciplina/witness/api/explorer.yaml
```diff
@@ -0,0 +1,178 @@
...
+      operationId: getBlocks
+      produces:
+      - application/json
+      parameters:
+      - in: query
+        name: before
+        description: Get only blocks before (not including) the specified one. If not specified, get the most recent blocks.
+        required: false
+        type: string
+        format: byte
```
Martoon-00: What exactly is passed in `before` seems ambiguous

specs/disciplina/witness/api/explorer.yaml
```diff
@@ -0,0 +1,178 @@
...
+      responses:
+        200:
+          description: Successful operation
+          schema:
+            type: array
+            items:
+              $ref: "#/definitions/BlockInfo"
+        404:
+          description: Block not found
```
Martoon-00: Minor: add `400 (Invalid format)`

specs/disciplina/witness/api/explorer.yaml
```diff
@@ -0,0 +1,215 @@
...
+    required:
+      - error
+    properties:
+      error:
+        type: string
+        enum:
+          - InvalidFormat
+        description: >-
+          Error type:
+            * `InvalidFormat` - Failed to deserialise one of parameters.
```
Martoon-00: I do not insist on having this type, but it's expected that you either enlist all errors on none, isn't it? ))

specs/disciplina/witness/api/explorer.yaml
```diff
@@ -0,0 +1,215 @@
...
+basePath: /api/witness/v1
+schemes:
+  - https
+  - http
+paths:
+  /blocks:
+    get:
+      summary: >-
+        Get list of blocks.
+        Only the batch of N most recent blocks are returned at once.
```
volhovm: Of how many?

specs/disciplina/witness/api/explorer.yaml
```diff
@@ -0,0 +1,215 @@
...
+  description: "This is an HTTP API for interacting with Disciplina Witness node. It is used by Block Explorer frontend to request info about blocks, addresses and transactions."
+  version: 0.1.0
+  title: Disciplina Block Explorer API
+  termsOfService: "https://disciplina.io/tnc.pdf"
+  contact:
+    email: hi@serokell.io
+  license:
+    name: Apache 2.0
+    url: "http://www.apache.org/licenses/LICENSE-2.0.html"
+host: "localhost:8091"
```
volhovm: I wouldn't fix this.

specs/disciplina/witness/api/explorer.yaml
```diff
@@ -0,0 +1,215 @@
...
+schemes:
+  - https
+  - http
+paths:
+  /blocks:
+    get:
+      summary: >-
+        Get list of blocks.
+        Only the batch of N most recent blocks are returned at once.
+        The hash of the last block in the batch can be passed as `before` parameter to get the next batch.
```
Martoon-00: It may be my and some other guys' subjective opinion, but `summary` is usually one-sentence note, remaining goes to `description`.

goodsoft: oh, right `description`

specs/disciplina/witness/api/explorer.yaml
goodsoft: Please choose an arbitrary number.

specs/disciplina/witness/api/explorer.yaml
goodsoft: Copied that from other specs.
Should I remove it at all?

goodsoft:
> I also do not see proper error handling. What happens if requested blocks are not present, for example?

Basically we only have two error conditions:
* the requested item does not exist
* the request is malformed

First one is covered by `404` error (but I will add `ErrResponse` structure to the body, as in Educator API).
`400` is already added for the second one.

specs/disciplina/witness/api/explorer.yaml
volhovm: Let it be a parameter maybe?

goodsoft: We still need to have some default amount or hard limit, as we probably don't want to serve the entire blockchain via this API

volhovm: Yes, let's say that the limit is 100, but let it be a parameter anyway.

specs/disciplina/witness/api/explorer.yaml
Martoon-00: Oh lol, ignore this, dunno where I look at :man_facepalming:

PR #101, opened 2018-08-08
[DSCP-180] Keep just reference on `Assignment` in `Submission`

Martoon-00:
Actually this required completely another approach to testing, because now there is no large data like `SignedSubmission` which would keep all test case input.
So I artificially introduced datatype which would keep all possible data test case may require.

### DIFF
--- a/educator/src/Dscp/DB/DSL/Interpret/SimpleTxDB.hs
+++ b/educator/src/Dscp/DB/DSL/Interpret/SimpleTxDB.hs
@@ -6,7 +6,7 @@ import Control.Lens (filtered, makePrisms, traversed)
 import Data.List (intersect, union)
 import Data.Map.Strict (Map)
 
-import Dscp.Core (Assignment (..), Course (..), SignedSubmission (..), Subject, Submission (..),
+import Dscp.Core (Assignment (..), Course (..), SignedSubmission (..), Subject,
                   activityTypeGraphIndexed, hasPathFromTo)
 import Dscp.Core.Foundation.Educator (PrivateTx (..))
 import Dscp.Crypto (hash)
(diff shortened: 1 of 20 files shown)


educator/src/Dscp/DB/DSL/Interpret/SimpleTxDB.hs
```diff
@@ -142,4 +142,6 @@ runSimpleTxDBQuery dbTx dbObj query =
         cId5 = Course 5
 
 getTxCourseId :: PrivateTx -> Id Course
-getTxCourseId tx = _aCourseId (_sAssignment (_ssSubmission (_ptSignedSubmission tx)))
+getTxCourseId tx = _aCourseId (getAssignment (_ssSubmission (_ptSignedSubmission tx)))
+  where
+    getAssignment = error "Some magic should happen here here"
```
volhovm: But it doesn't happen. Is it OK?

educator/tests/Test/Dscp/Educator/Web/Student/Queries.hs
```diff
@@ -348,17 +350,21 @@ spec_StudentApiQueries = describe "Basic database operations" $ do
...
+                 (genCoreTestEnv simpleCoreTestParams) -> env
                ) -> do
-                prepareAndCreateSubmissions sigSubmissions
+                let student = tiOne $ cteStudents env
+                    sigSubs = take 3 $ tiList $ cteSignedSubmissions env
+
+                prepareForSubmissions env
+                forM_ (nub $ toList sigSubs) $ \sigSub -> do
+                    void $ sqlTx $ CoreDB.createSignedSubmission sigSub
+                    liftIO $ threadDelay 10000  -- sqlite keeps time in ms precision
```
volhovm: Eeh, hacky. I guess it's not bad that some txs will have the same timestamp, but we'd like to avoid this for tests only?

educator/src/Dscp/DB/DSL/Interpret/SimpleTxDB.hs
Martoon-00: My feeling is that there is no need to maintain this module, but waiting for response of @Heimdell

PR #102, opened 2018-08-08
Reduce use of stateful NIX_PATH

yegortimoshenko:


default.nix
```diff
@@ -1,8 +1,14 @@
-with import <nixpkgs> { overlays = [ (import <serokell-overlay/pkgs>) ]; };
+with import (fetchTarball "https://github.com/serokell/nixpkgs/archive/master.tar.gz") {
+  config.allowUnfree = true;
+  overlays = [ (import "${fetchGit "ssh://git@github.com:/serokell/serokell-overlay"}/pkgs") ];
```
yorickvP: does this need that "${...}" around it?

yegortimoshenko: Yes, because it has `/pkgs"` at the end (we need specific folder in `serokell-overlay`, not its root.

PR #104, opened 2018-08-08
[DSCP-163] Do not create initial data if already exists

Martoon-00:


educator/src/Dscp/Educator/Web/Bot/Setting.hs
```diff
@@ -186,8 +187,10 @@ delayed action
 
 botPrepareInitialData :: (BotWorkMode m, HasBotSetting) => m ()
 botPrepareInitialData = do
-    mapM_ (\(c, t, s) -> createCourse c (Just t) s) (bsCourses botSetting)
-    mapM_ createAssignment (bsAssignments botSetting)
+    exists <- existsCourse (head . Exts.fromList $ bsBasicCourses botSetting)
+    unless exists $ do
+        mapM_ (\(c, t, s) -> createCourse c (Just t) s) (bsCourses botSetting)
```
Heimdell: Not an issue, but it can be
```haskell
forM_ (bsCourses botSetting) $ \(c, t, s) ->
    createCourse c (Just t) s
```

PR #107, opened 2018-08-09
[DSCP-174] [WIP] Implement Block Explorer API

goodsoft:
TODO:
- [ ] `/addresses` endpoint


specs/disciplina/witness/api/explorer.yaml
```diff
@@ -172,16 +179,12 @@ definitions:
         format: int64
       prevHash:
         $ref: "#/definitions/Hash"
-  AddressInfo:
+  AccountInfo:
     type: object
     properties:
       balance:
```
volhovm: This is irrelevant, but why don't we return the nonce?

witness/src/Dscp/Snowdrop/Configuration.hs
```diff
@@ -91,13 +97,15 @@ data Values
     = TipValueVal       (TipValue HeaderHash)
     | BlundVal           SBlund
     | AccountOutVal      Account
+    | TxVal              T.GTxWitnessed
```
volhovm: This way we store all transactions twice. Maybe it's better to define mapping `tx -> headerHash`?

witness/src/Dscp/Snowdrop/Expanders.hs
```diff
@@ -139,7 +139,7 @@ seqExpandersBalanceTx =
         let inpNewBal     = aBalance inpPrevAcc + inputBack - inputSent
         let newInpAccount = Account
                 { aBalance = inpNewBal
-                , aNonce   = tiaNonce (txInAcc twTx) + 1
+                , aNonce   = tiaNonce (txInAcc twTx)
```
volhovm: Wait, what?

specs/disciplina/witness/api/explorer.yaml
goodsoft: we do, I'll update the spec later

witness/src/Dscp/Snowdrop/Configuration.hs
goodsoft: Yeah, right, we also want to return block reference for `/transactions` endpoint anyway.

witness/src/Dscp/Snowdrop/Expanders.hs
goodsoft: The nonce was incremented in two places and caused off-by-one errors.
Probably, this wasn't the best place to fix it, I'll try making it a bit more clear.

goodsoft: By using more descriptive names, like, lastNonce and nextNonce, among other things :)

goodsoft: Oh, actually this was the right place to change it.
This one here is the new account _after_ tx application.
Accounts store last used nonce, so there was no sense in incrementing it here.

witness/src/Dscp/Snowdrop/Configuration.hs
```diff
@@ -56,27 +57,45 @@ publicationOfPrefix = Prefix 4
...
     = TipKeyIds          TipKey
     | BlockRefIds       (BlockRef  HeaderHash)
     | AccountInIds       AccountId
-    | PublicationOfIds   T.PublicationsOf
-    | PublicationHeadIds T.PublicationHead
+    | TxIds              T.GTxId
+    | TxOfIds            TxsOf
+    | TxHeadIds          TxHead
+    | PublicationOfIds   PublicationsOf
+    | PublicationHeadIds PublicationHead
```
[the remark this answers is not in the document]
Heimdell: First is for relation (`Address` -> first `Publication`), second is for relation (`Publication` -> prev `Publication`)

witness/src/Dscp/Witness/Logic/Traversals.hs
```diff
@@ -131,3 +132,28 @@ getBlocksFromTo (headerHash -> olderH) (headerHash -> newerH) = runExceptT $ do
...
+
+    let retrievedOldest = NE.head (unOldestFirst blocks)
+
+    -- sanity checks, remove them later
+    when (NE.length (unOldestFirst blocks) < depthDiff &&
+          retrievedOldest /= genesisBlock) $
+        error "getBlocksFromTo: retrieved less than expected"
+
+    when (NE.length (unOldestFirst blocks) > depthDiff) $
+        error "getBlocksFromTo: retrieved too many"
```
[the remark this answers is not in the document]
volhovm: Yes, we'll get rid of it.

PR #119, opened 2018-08-13
Frontend for faucet app

romkaspb:


goodsoft:
`Generate wallet` popup shows only an address and a public key.
We need to show private key too, for user to be able to restore this account in the wallet app.
Appropriate warnings about security should be present, I think.

faucet/frontend/src/locales/cn.json
```diff
@@ -0,0 +1,31 @@
+{
+  "welcomeMsg" : "在這裡你可以獲得令牌來測試DISCIPLINA區塊鏈",
```
volhovm: What a nice Chinese localisation :+1:

faucet/frontend/src/assets/api/student.js
```diff
@@ -0,0 +1,36 @@
+const _assignments = [
```
volhovm: What do you need this stub data for?

faucet/frontend/src/locales/cn.json
mkaito: I hope that doesn't accidentally tell someone to go pound sand.

PR #121, opened 2018-08-14
 [DSCP-196] Add Educator authorization check

hithroc:
This PR depends on #120

### DIFF
--- a/educator/src/Dscp/Educator/Web/Auth.hs
+++ b/educator/src/Dscp/Educator/Web/Auth.hs
@@ -1,0 +1,76 @@
+{-# LANGUAGE DataKinds     #-}
+{-# LANGUAGE TypeOperators #-}
+
+-- | Necessary types and implementation for web server authenthication
+module Dscp.Educator.Web.Auth
+       ( Auth'
+       , AuthData (..)
+       , checkAuthData
+       , WithCommonAuthData (..)
+       ) where
+
+import Data.Aeson (FromJSON)
+import Data.Time.Clock (UTCTime, diffUTCTime, getCurrentTime)
+import Network.Wai (rawPathInfo)
+import Servant ((:>), HasServer, ServantErr (..), ServerT, err401, hoistServerWithContext, route)
+import Servant.Auth.Server (AuthCheck (..), AuthResult (..), FromJWT)
+import Servant.Auth.Server.Internal.Class (AreAuths (..), runAuths)
+import Servant.Server.Internal.RoutingApplication (DelayedIO, addAuthCheck, delayedFailFatal,
+                                                   withRequest)
+
+---------------------------------------------------------------------------
+-- Data types
+---------------------------------------------------------------------------
+
+-- | Custom authetication API type.
+data Auth' auth res
+
+-- | A type that reperesents the data that the client has sent to authenthicate
+data AuthData = AuthData
+    { adPath :: Text
+    , adTime :: UTCTime
+    }
+    deriving Generic
+
+-- | A data type that contains common authenthication data ('AuthData')
+data WithCommonAuthData a = WithCommonAuthData AuthData a
+
+---------------------------------------------------------------------------
+-- Instances
+---------------------------------------------------------------------------
+
+instance ( HasServer api ctxs
+         , AreAuths '[auth] ctxs (WithCommonAuthData res)
+         ) => HasServer (Auth' auth res :> api) ctxs where
+    type ServerT (Auth' auth res :> api) m = res -> ServerT api m
+
+    hoistServerWithContext _ pc nt s = hoistServerWithContext (Proxy :: Proxy api) pc nt . s
+
+    route _ context subserver = route (Proxy :: Proxy api) context (addAuthCheck subserver authCheck)
+      where
+        authMsg :: ByteString
+        authMsg = "Set Authorization header with a proper JWT"
+        authCheck :: DelayedIO res
+        authCheck = withRequest $ \req -> do
+            let authChecks = runAuths (Proxy :: Proxy '[auth]) context
+            authRes <- liftIO $ runAuthCheck (authChecks >>= checkAuthData) req
+            case authRes of
+                (Authenticated (r :: res)) -> return r
+                _ -> delayedFailFatal $ err401 { errHeaders = [("WWW-Authenticate", authMsg)] }
+
+instance FromJSON AuthData
+instance FromJWT AuthData
+
+---------------------------------------------------------------------------
+-- Helpers
+---------------------------------------------------------------------------
+
+checkAuthData :: WithCommonAuthData r -> AuthCheck r
+checkAuthData (WithCommonAuthData AuthData {..} r) = do
+    request <- ask
+    -- request path verification
+    guard (adPath == decodeUtf8 (rawPathInfo request))
+    -- time verification
+    curTime <- liftIO $ getCurrentTime
+    guard (diffUTCTime curTime adTime <= 300)
+    return r
--- a/educator/src/Dscp/Educator/Web/Educator/Auth.hs
+++ b/educator/src/Dscp/Educator/Web/Educator/Auth.hs
@@ -1,10 +1,10 @@
 {-# LANGUAGE DataKinds     #-}
 {-# LANGUAGE TypeOperators #-}
 
--- | Necessary types and implementation for Student authenthication
-module Dscp.Educator.Web.Student.Auth
-       ( GetStudentsAction (..)
-       , StudentAuth
+-- | Necessary types and implementation for Educator authenthication
+module Dscp.Educator.Web.Educator.Auth
+       ( EducatorPublicKey (..)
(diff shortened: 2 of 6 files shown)


educator/src/Dscp/Educator/Web/Educator/Auth.hs
```diff
@@ -0,0 +1,42 @@
+{-# LANGUAGE DataKinds     #-}
+{-# LANGUAGE TypeOperators #-}
+-- | Nessesary types and implementation for Educator authenthication
```
volhovm: Necessary.

educator/src/Dscp/Educator/Web/Auth.hs
```diff
@@ -0,0 +1,72 @@
+{-# LANGUAGE DataKinds     #-}
+{-# LANGUAGE TypeOperators #-}
+-- | Nessesary types and implementation for web server authenthication
```
volhovm: Necessary. Also it's subjective, but the comment usually sticks to the `module` line like this:
```
{-# LANGUAGE Pragma1 #-}

-- | Comment
module Dscp.mem where ...
```

educator/src/Dscp/Educator/Web/Server.hs
```diff
@@ -49,35 +52,34 @@ mkStudentApiServer
...
+        (Proxy :: Proxy '[GetStudentsAction])
         nat
         (toServant handlers)
-    addAuth :: (Student -> Server StudentAPI) -> Server ProtectedStudentAPI
-    addAuth h = \case
-        Authenticated (WithStudent student _) -> h student
-        _ -> throwAll err401
+
+-- This is a temporary function that provides a dummy GetStudentAction
+createGetStudentsAction :: IO GetStudentsAction
```
volhovm: How do we know which secret keys were generated? Post them somewhere maybe?

hithroc: This change is part of DSCP-141 which is still not merged yet. Here I just separated the key generation into its own function as @Martoon-00 requested, and instead of generating just two I generate 100. Two of them are posted in #120 description.

educator/src/Dscp/Educator/Web/Educator/Auth.hs
hithroc: I feel ashamed.

volhovm: Don't be, @pva701 does it even worse :peka:

educator/src/Dscp/Educator/Web/Auth.hs
```diff
@@ -0,0 +1,72 @@
...
+
+---------------------------------------------------------------------------
+-- Helpers
+---------------------------------------------------------------------------
+
+checkAuthData :: (AuthData, r) -> AuthCheck r
+checkAuthData (AuthData {..}, r) = do
+    request <- ask
+    -- request path verification
+    when (adPath /= decodeUtf8 (rawPathInfo request)) mzero
```
Heimdell: Why not
```haskell
guard (adPath == decodeUtf8 (rawPathInfo request))
```
?

volhovm: Yes, please use guards.

educator/src/Dscp/Educator/Web/Auth.hs
```diff
@@ -0,0 +1,72 @@
...
+
+instance FromJSON AuthData
+instance FromJWT AuthData
+
+---------------------------------------------------------------------------
+-- Helpers
+---------------------------------------------------------------------------
+
+checkAuthData :: (AuthData, r) -> AuthCheck r
+checkAuthData (AuthData {..}, r) = do
```
volhovm: This doesn't make any sense. Why do you just return the `r`? This also forces you to write a pretty weird `instance IsAuth EducatorAuth (AuthData, ())`. Can't it be solved trivially with basic tools like typeclasses and adts?

educator/src/Dscp/Educator/Web/Educator/Auth.hs
```diff
@@ -0,0 +1,42 @@
...
+-- Data types
+---------------------------------------------------------------------------
+
+-- | Custom authentication type for auth-servant
+data EducatorAuth
+
+-- | Type that holds Educator's public key
+newtype EducatorPublicKey = EducatorPublicKey PublicKey
+
+instance IsAuth EducatorAuth (AuthData, ()) where
```
volhovm: Again -- this looks kind of weird. I'd expect `IsAuth EducatorAuth EducatorAuthData` or something, but not a `()`.

educator/src/Dscp/Educator/Web/Auth.hs
hithroc: But why? What does that achieve?

volhovm: I've already answered that in the first comment.

hithroc: Your first comment basically says "it looks weird" and not much more. That is subjective, because it doesn't to me. I am not convinced that what you're proposing is better, at least not from how I understood it.

I want `HasServer` instance to be generic and work the same for anything that ever wants to use Auth. That way I know for sure that the general part of authentication, `AuthData` is always checked, because it is ensured by this instance. For that to work, we separate the return type of the JWT signature checking into two things: AuthData that is generalized auth and checked by `HasServer` instance and what is the API endpoint handler is actually going to receive which is represented by this tuple.

Besides, this is not much different from how the original student auth worked. Look at `WithStudent` type, it is pretty much the same tuple. That is _exactly the same thing_, but just instead of having `WithStudent AuthData` we now have basically `WithAuth Student` in form of a tuple.

volhovm: >That is subjective

All code standards are. What's better, `instance MyClass (Datatype, (), (), Int)` or `instance MyClass MyDatatype`?

>I want HasServer instance to be generic 

There's no need to want that, but if you _really do_, you can use typeclasses or ADTs. Or, at least, you can write `instance IsAuth EducatorAuth (GenericAuth ())`, which is better than the pair. Pairs are pretty unobvious.

I got it, but is it possible to define a type alias?

volhovm: Sure, yes.

specs/disciplina/educator/api/educator.yaml
```diff
@@ -695,7 +745,39 @@ paths:
...
+      Authentication system requires a signed JWT in Authorization header in `Bearer <JWT>` format. Example header:
+      ```
+      Authorization: Bearer eyJhbGciOiJFZERTQSJ9.eyJkYXQiOnsiYWRQYXRoIjoiL2FwaS9lZHVjYXRvci92MS9zdHVkZW50cyIsImFkVGltZSI6IjIwOTktMDgtMTBUMTM6MjY6NTkuNDYxOTk4MTM2WiJ9fQ.gU032BGkGwllzPtBLxsvXNZttI6qmEh4M2DGfdiIV6k2L_SuPNR8-YHGphQ2_bytOSKl4Hf8LKPt9g0wNjJbAQ
+      ```
+      The JWT should be signed by the educator's private key and contain one claim "dat". The claim should have an object inside with two fields "adPath" and "adTime'. "adPath" should contain the path to the endpoint as it is going to be in the raw request and "adTime" should have the current time in ISO-8601 format. Example JWT object:
+      ```
+      {
+        "dat": {
+          "adTime": "2018-08-10T13:22:40.461998136Z",
+          "adPath": "/api/educator/v1/students"
```
Martoon-00: Typo?

educator/src/Dscp/Educator/Web/Student/Auth.hs
```diff
@@ -81,18 +35,11 @@ instance IsAuth StudentAuth (WithStudent AuthData) where
...
-    let tryAuth :: PublicKey -> AuthCheck (PublicKey, AuthData)
         tryAuth pub = fmap (pub,) . jwtAuthCheck . defaultJWTSettings . pubKeyToJwk $ pub
     (publicKey, authData) <- asum . map tryAuth $ students
-    -- request path verification
-    when (adPath authData /= decodeUtf8 (rawPathInfo request)) mzero
-    -- time verification
-    curTime <- liftIO $ getCurrentTime
-    when (diffUTCTime curTime (adTime authData) > 300) mzero
-    return (WithStudent (Address $ hash publicKey) authData)
+    return $ WithCommonAuthData authData (Address $ hash publicKey)
```
Martoon-00: There is `mkAddr`

specs/disciplina/educator/api/educator.yaml
hithroc: Why? Isn't that a valid endpoint for Educator?

Martoon-00: Hm, okay

hithroc:
I signed the commits.

PR #122, opened 2018-08-14
[DSCP-192] Support bind-internal CLI option

volhovm:


witness/src/Dscp/Web/Server.hs
```diff
@@ -41,7 +40,9 @@ buildServantLogConfig
 buildServantLogConfig modifyName = do
     origNameSel <- Log.logName
     logger <- views (lensOf @(Logging IO)) Log._log
-    let origName = fromMaybe mempty $ origNameSel ^? _GivenName
+    let origName = case origNameSel of
+                       GivenName x -> x
```
Heimdell: Why indent right so far?

volhovm: 4 spaces from `case`, that's the standard practice.

PR #125, opened 2018-08-14
[DSCP-126] [DSCP-209] Fee calculation

Heimdell:


witness/src/Dscp/Snowdrop/AccountValidation.hs
```diff
@@ -152,34 +165,22 @@ assertExists thing message = do
...
---     () <- validateIff SumMustBeNonNegative (difference >= 0)
---     return difference
+-- TODO: this and '{A,a}uthenticate' should be in Dscp.SD.Util
+canAffordFees
+    :: forall ctx
+    .  Account
+    -> Word32
+    -> ERoComp Exceptions Ids Values ctx ()
+canAffordFees account size = do
+    validateIff CannotAffordFees $ aBalance account >= fromIntegral size
```
Martoon-00: Shouldn't it be `aBalance account >= size + txInValue`?

witness/src/Dscp/Snowdrop/AccountValidation.hs
```diff
@@ -63,6 +65,13 @@ validateSimpleMoneyTransfer = mkValidator ty
   where
     ty = StateTxType $ getId (Proxy @TxIds) AccountTxTypeId
 
+data Authenticated payload = Authenticated
```
Martoon-00: Leave a comment please, this is going to be used all over the project iiuc

witness/src/Dscp/Snowdrop/AccountValidation.hs
Heimdell: No, this is for account state with already retracted payment.

Heimdell: Thus we can use it for transactions w/o payment.

witness/src/Dscp/Snowdrop/AccountValidation.hs
Heimdell: Ok. Didn't know we want to spill snowdrop validtors all over the place, but hey.

Martoon-00: Right, I just saw a familiar pattern, but in other places we use `sign` :man_facepalming:

core/src/Dscp/Core/Foundation/Witness.hs
```diff
@@ -236,26 +242,50 @@ newtype GTxId = GTxId (Hash GTx)
 
 -- | Compute tx id.
 toGTxId :: (Serialise Tx, Serialise PublicationTx) => GTx -> GTxId
-toGTxId (GMoneyTx tx) = GTxId . unsafeCastHash . toTxId $ tx
+toGTxId (GMoneyTx tx)       = GTxId . unsafeCastHash . toTxId $ tx
 toGTxId (GPublicationTx tx) = GTxId . unsafeCastHash . toPtxId $ tx
 
-data GTxWitnessed
+data WithSize a = WithSize
```
volhovm: Please stop putting utility datatypes into the `Witness` module, it's designed to carry only the logic-related types of witness, not serialisation wrappers.

core/src/Dscp/Core/Foundation/Witness.hs
```diff
@@ -236,26 +242,50 @@ newtype GTxId = GTxId (Hash GTx)
...
+    deriving (Generic, Eq, Show)
+
+withSize :: Serialise a => a -> WithSize a
+withSize payload =
+    WithSize (fromIntegral $ BS.length (serialise payload)) payload
+
+instance Buildable a => Buildable (WithSize a) where
+    build (WithSize size a) = "WithSize {" +| size |+ ", " +| a |+ "}"
+
+type GTxWitnessed = WithSize GTxWitnessedRaw
```
volhovm: This doesn't look right.

Heimdell: This is not serialisation wrapper, it it from our domain of modelling. We have fees, and we get them from tx size. We can't put size inside the transaction, because tx should know nothing about its size and fees. But the transaction should carry size anyway.

You are free to show better way to do this.

volhovm: Transaction carries size by itself, implicitly.

core/src/Dscp/Core/Foundation/Witness.hs
```diff
@@ -236,26 +246,62 @@ newtype GTxId = GTxId (Hash GTx)
 
 -- | Compute tx id.
 toGTxId :: (Serialise Tx, Serialise PublicationTx) => GTx -> GTxId
-toGTxId (GMoneyTx tx) = GTxId . unsafeCastHash . toTxId $ tx
+toGTxId (GMoneyTx tx)       = GTxId . unsafeCastHash . toTxId $ tx
 toGTxId (GPublicationTx tx) = GTxId . unsafeCastHash . toPtxId $ tx
 
-data GTxWitnessed
-    = GMoneyTxWitnessed TxWitnessed
+data CarriedKind = Size | Fee
```
goodsoft: ELI5, why is fee and size stored separately?

Also, as far as I see, we now store tx with size (not fee, see `GTxWitnessed`) in block, is this intended?
When I have a block with transactions (e.g. in block explorer), how do I get the fee for a particular transactions?

Heimdell: Below there is `turnIntoFees` function that performs `Carrying 'Size GTxwitnessedRaw -> Carrying 'Fee GTxwitnessedRaw`.

The reason I did it like that is because size is only needed to calculate fees, and reason to calculated fee only emerges in (Tx -> SD.Tx) expander, where I use `turnIntoFees`.

Fee and size represent different stages of fee calculation and they are stored disjunctively, not separately. The whole `CarriedKind` thing is so you can't calculate fees twice or zero times.

goodsoft: I, for one, don't see any particular use for a transaction with size _and_ without fee.
Why not calculate and store them both at the same time?

And, once again, how do I reliably get the transaction fee from `GTxWitnessed` in block explorer?

goodsoft: Moreover, there is a use for a function, which directly calculates fee for a transaction.
It will be needed in wallet app, when we show fee preview for outgoing transactions.

core/src/Dscp/Core/Genesis.hs
```diff
@@ -116,7 +116,8 @@ formGenesisInfo GenesisConfig{..} =
                           , txOuts = txOutputs }
                 twWitness = TxWitness { txwSig = sign sk (toTxId twTx, pk, ())
                                       , txwPk = pk }
-            in GMoneyTxWitnessed $ TxWitnessed {..}
+            -- Zero here means minimal fee, since its a genesis.
+            in carryingNothing $ GMoneyTxWitnessed TxWitnessed {..}
```
volhovm: By the way, the genesis block is applied without checks, so tx validator won't kick in anyway.

witness/src/Dscp/Snowdrop/AccountValidation.hs
```diff
@@ -152,27 +165,14 @@ assertExists thing message = do
...
---     () <- validateIff SumMustBeNonNegative (difference >= 0)
---     return difference
+-- TODO: this and '{A,a}uthenticate' should be in Dscp.SD.Util
+canAffordFees
+    :: forall ctx
+    .  Account
+    -> Word32
+    -> ERoComp Exceptions Ids Values ctx ()
+canAffordFees account fee = do
+    validateIff CannotAffordFees $ aBalance account >= fromIntegral fee
```
volhovm: Calculate fee here, also return a specific mismatch pair `a < b` if it fails.

witness/src/Dscp/Snowdrop/Expanders.hs
```diff
@@ -45,92 +47,115 @@ expandGTxs ::
...
 
--- | Expand list of global txs.
+-- | Expand one global tx.
 expandGTx ::
        ( Default (ChgAccum ctx)
        , HasLens ctx RestrictCtx
        , HasLens ctx (ChgAccumCtx ctx)
        )
-    => GTxWitnessed
+    => FeeCoefficients
```
volhovm: What's the hassle here? Why do you pass `FeeCoefficients` directly instead of using reflection? It is obviously inconvenient. Don't be afraid to pull the `WitnessContext` altogether, it was designed exactly for this type of usage.

witness/src/Dscp/Snowdrop/Expanders.hs
```diff
@@ -45,92 +47,115 @@ expandGTxs ::
...
-
-toProofBalanceTx :: TxWitnessed -> (StateTxType, Proofs)
-toProofBalanceTx (TxWitnessed tx (TxWitness {..})) =
-    (txType, AddressTxWitness $ WithSignature {..})
+       GTxWitnessedWithFee -> SeqExpanders Exceptions Ids Proofs Values ctx GTxWitnessedWithFee
+seqExpandersGTx (Carrying _ (GMoneyTxWitnessed _)) =
+    flip contramap seqExpandersBalanceTx $
+        \(Carrying fee (GMoneyTxWitnessed tx)) ->
+            Carrying fee tx
+
```
volhovm: Two lines between different pattern-match cases of the same function :thinking:

witness/src/Dscp/Witness/CLI.hs
```diff
@@ -96,6 +97,15 @@ witnessKeyParamsParser = do
     wkpCommittee <- optional committeeParamsParser
     pure $ WitnessKeyParams {..}
 
+feeCoefficientsParser :: Parser FeeCoefficients
```
volhovm: Noooooo, it should be in configuration. Configuration is a set of "cluster-related" values -- it doesn't make sense to change this value for a single cluster user to anything else -- you'll either pay more or your tx will be rejected, or you'll create an incorrect block.

PS: core configuration. Wallet (that doesn't want to know anything about witness' code) should be able to create a proper tx.

witness/src/Dscp/Snowdrop/Configuration.hs
```diff
@@ -141,11 +141,14 @@ type CanVerifyPayload txid payload =
         (txid, PublicKey, payload)
 
 
-type PersonalisedProof txid  payload =
-    WithSignature
+data PersonalisedProof txid payload = PersonalisedProof
```
volhovm: So this seems to be a proof for tx, right? Why `ppFee` is `Word32`? Can we rename it into something better or at least add a comment?

witness/src/Dscp/Snowdrop/Expanders.hs
```diff
@@ -33,10 +35,11 @@ expandBlock ::
        , HasLens ctx RestrictCtx
        , HasLens ctx (ChgAccumCtx ctx)
        )
-    => Block
+    => FeeCoefficients
```
volhovm: Didn't I ask to use reflection? Or there's a reason not to do that?

witness/src/Dscp/Snowdrop/Expanders.hs
volhovm: This ^

witness/src/Dscp/Snowdrop/Expanders.hs
```diff
@@ -100,28 +112,42 @@ seqExpandersPublicationTx =
...
+                                Upd account
+                                    -- You can say that `PubOf ==> LastPub`
+                                    -- mechanism will prevent double-pub
+                                    -- but lets not depend on that impl detail.
+                                    { aNonce   = aNonce + 1
+                                    , aBalance = aBalance - fromIntegral fee
+                                    }
+                          ]
+
+                    Nothing -> do
```
volhovm: Redundant do.

witness/src/Dscp/Snowdrop/Expanders.hs
```diff
@@ -100,28 +112,42 @@ seqExpandersPublicationTx =
...
-                  [ PublicationsOf  ptAuthor ==> change (LastPublication ptBlock)
-                  , PublicationHead ptBlock  ==> New    (PublicationNext ptPrevBlock)
-                  ]
+                queryOne (AccountId ptAuthor) >>= \case
+                    Just (account@ Account {..}) -> do
+                        let change = if isJust maybePub then Upd else New
+                        return $ mkDiffCS $ Map.fromList
+                          [ PublicationsOf  ptAuthor ==> change (LastPublication ptBlock)
+                          , PublicationHead ptBlock  ==> New    (PublicationNext ptPrevBlock)
+                          , AccountId ptAuthor ==>
```
volhovm: Wait, but who gets the fee?

witness/src/Dscp/Snowdrop/Expanders.hs
```diff
@@ -100,28 +112,42 @@ seqExpandersPublicationTx =
...
-    contramap (\(GMoneyTxWitnessed tx) -> tx) seqExpandersBalanceTx
-seqExpandersGTx (GPublicationTxWitnessed _) =
-    contramap (\(GPublicationTxWitnessed ptx) -> ptx) seqExpandersPublicationTx
+seqExpandersGTx
+    :: Word32
+    -> GTxWitnessed
+    -> SeqExpanders Exceptions Ids Proofs Values ctx GTxWitnessed
+seqExpandersGTx _ (GMoneyTxWitnessed _) =
+    flip contramap seqExpandersBalanceTx $ \(GMoneyTxWitnessed tx) -> tx
+
```
volhovm: No need for the extra line.

witness/src/Dscp/Snowdrop/Expanders.hs
```diff
@@ -100,28 +112,42 @@ seqExpandersPublicationTx =
...
-seqExpandersGTx (GMoneyTxWitnessed _) =
-    contramap (\(GMoneyTxWitnessed tx) -> tx) seqExpandersBalanceTx
-seqExpandersGTx (GPublicationTxWitnessed _) =
-    contramap (\(GPublicationTxWitnessed ptx) -> ptx) seqExpandersPublicationTx
+seqExpandersGTx
+    :: Word32
+    -> GTxWitnessed
+    -> SeqExpanders Exceptions Ids Proofs Values ctx GTxWitnessed
+seqExpandersGTx _ (GMoneyTxWitnessed _) =
+    flip contramap seqExpandersBalanceTx $ \(GMoneyTxWitnessed tx) -> tx
```
volhovm: I don't get it, why do we not use the `fee` parameter here?

witness/src/Dscp/Snowdrop/Types.hs
```diff
@@ -45,6 +46,8 @@ data AccountValidationException
     | ReceiverMustIncreaseBalance  -- ^ Receiver cannot decrease in its 'aBalance'.
     | SumMustBeNonNegative         -- ^ Amount of money sent must be greater of equal
                                    -- to the total amount received.
+    | CannotAffordFees             -- ^ Given account state cannot afford given fees.
+    | BalanceCannotBecomeNegative
```
volhovm: Comment.

witness/src/Dscp/Witness/Logic/Processing.hs
```diff
@@ -64,23 +63,26 @@ applyBlockRaw toVerify block = do
...
   where
     blockPrefixes = S.fromList [tipPrefix, blockPrefix, txPrefix]
 
 applyBlock :: WitnessWorkMode ctx m => Block -> m AvlProof
-applyBlock = applyBlockRaw True
+applyBlock = applyBlockRaw feeCoefficients True
 
 applyGenesisBlock :: WitnessWorkMode ctx m => m ()
-applyGenesisBlock = void $ applyBlockRaw False genesisBlock
+applyGenesisBlock = void $ applyBlockRaw noFees False genesisBlock
```
volhovm: Please elaborate on why did you eventually decide to pass fees directly instead of figuring out why block validator kicks in when we don't ask it to.

witness/src/Dscp/Snowdrop/Expanders.hs
Heimdell: We have Genesis block application, which has to be not covered by fees, so reflection is used in `Logic.Processing`.

witness/src/Dscp/Snowdrop/Expanders.hs
Heimdell: Same question. Nobody gets fee in money transfer as well.

witness/src/Dscp/Snowdrop/Expanders.hs
Heimdell: Fee is not needed in money transfer expander, since we're already doing changes in author account.
It is used from proof in validator.

witness/src/Dscp/Witness/Logic/Processing.hs
Heimdell: > Genesis
> noFees

volhovm: When i pass `noFees` validator should not kick in. Why did you decide to add kostyl instead of figuring out the problem?

witness/src/Dscp/Snowdrop/Expanders.hs
Heimdell: How do we know current block leader? Should I take it from slot info?

Heimdell: So, question "how" remains.

witness/src/Dscp/Snowdrop/Configuration.hs
Heimdell: It is `Word32` because I wanted some non-negative thing.

witness/src/Dscp/Snowdrop/Expanders.hs
volhovm: As an argument.

witness/src/Dscp/Snowdrop/Configuration.hs
volhovm: But `Coin` is `Word64`.

goodsoft:
So, we're back to storing `[GTxWitnessed]` in a block?

Can you, please, elaborate on how will I be able to get the actual calculated fee for a transaction in a block in this case (to be used in Block Explorer and wallet UI)?

witness/src/Dscp/Snowdrop/Expanders.hs
```diff
@@ -44,51 +47,60 @@ expandGTxs ::
...
-        let (a, b) = toProofPublicationTx ptx
-        in  (a, b, seqExpandersGTx t)
-
-toProofPublicationTx :: PublicationTxWitnessed -> (StateTxType, Proofs)
-toProofPublicationTx (PublicationTxWitnessed tx (PublicationTxWitness {..})) =
+       , SeqExpanders Exceptions Ids Proofs Values ctx GTxWitnessed
+       )
+getByGTx t =
+    let size = BS.length $ serialise t
+        fee  = calculateFee feeCoefficients size
```
flyingleafe: By design, the fee for publications is calculated differently than fee for money transactions - it should depend on the size of the private block, encoded in the private header. We can leave it like this for alpha, but you should bear this in mind.

witness/src/Dscp/Snowdrop/Expanders.hs
flyingleafe: We can simply burn the fee for alpha release, who even cares, Cardano mainnet is still doing this.

Heimdell: Too late, send it to issure in the last commit.

witness/src/Dscp/Snowdrop/AccountValidation.hs
```diff
@@ -112,89 +119,58 @@ preValidateSimpleMoneyTransfer =
...
+        let balanceBefore = aBalance before
+        -- Total amount of coins sent (sum of outputs)
+        let receivedTotal = sum allOutputs
+        -- Previous value minus fees
+        let receivedNoFees = receivedTotal - coinToInteger (unFees fees)
 
-        validateIff SumMustBeNonNegative $
-            paid - sum deltas == 0
+        unless (balanceBefore - receivedNoFees >= 0) $ throwLocalError SumMustBeNonNegative
+        unless (balanceBefore - receivedTotal >= 0) $ throwLocalError CannotAffordFees
```
Heimdell: If it works, its OK. But I expected here `something >= fees`.

witness/src/Dscp/Snowdrop/AccountValidation.hs
```diff
@@ -112,89 +119,58 @@ preValidateSimpleMoneyTransfer =
...
-
-      Nothing -> do
-        throwLocalError message
+requirePart whole message = maybe (throwLocalError message) pure $ proj whole
 
 -- | Require that id exists in a database or throw error.
-assertExists
-    :: ( HasExceptions e [e1, StatePException]
+assertExists ::
+       ( HasExceptions e '[ e1, StatePException]
```
Heimdell: &lt;style-war&gt;
Why carrying first operator there, if all following operators are in the starts of their lines?
&lt;/style-war&gt;
Why manually lifting the list? I thought it worked fine w/o `'`.

volhovm: This was your change.

witness/src/Dscp/Snowdrop/AccountValidation.hs
```diff
@@ -112,89 +119,58 @@ preValidateSimpleMoneyTransfer =
...
+        allOutputs <- mapM validateSaneArrival recipients
 
-        (paid, deltas) <-
-            (,) <$> validateSaneDeparture payer before
-                <*> mapM validateSaneArrival recipients
+        let balanceBefore = aBalance before
+        -- Total amount of coins sent (sum of outputs)
+        let receivedTotal = sum allOutputs
+        -- Previous value minus fees
+        let receivedNoFees = receivedTotal - coinToInteger (unFees fees)
```
Martoon-00: I'm stuck here (
What physical sense `sum outs - fee` has?

witness/src/Dscp/Snowdrop/Configuration.hs
```diff
@@ -256,8 +264,9 @@ data Proofs
 ----------------------------------------------------------------------------
 
 data ExpanderException =
-    MTxDuplicateOutputs
+      MTxDuplicateOutputs
     | CantResolveSender
+    | ExpanderInternalError String
```
Martoon-00: I personally don't think that internal errors should appear in `ExceptT` or be thrown with `throwM` as exceptions, `error` is better, because when I catch exceptions in upper logic I shouldn't care or know about internal exceptions. We even had this rule in cardano's exceptions guide.

witness/src/Dscp/Snowdrop/Expanders.hs
```diff
@@ -170,16 +227,45 @@ seqExpandersBalanceTx =
...
+                    [ TxsOf  txAddr      ==> Upd (LastTx txId)
+                    , TxHead txAddr txId ==> New (TxNext prevTxId)
                     ]
 
-        pure $ mkDiffCS $ Map.fromList $ inp : outs ++ concat accList
+        -- Money flow related changes (w/o fees)
+        let changesList = inp : outs ++ concat miscChanges
+
+        -- Checking for duplicates
+        let hasDuplicates l = length l /= length (ordNub l)
```
Martoon-00: I would appreciate if you move it to `Dscp.Util`, it's often useful

core/src/Dscp/Core/Fees.hs
```diff
@@ -0,0 +1,77 @@
...
+newtype Fees = Fees { unFees :: Coin }
+    deriving (Eq, Ord, Show, Generic, Hashable, Bounded)
+
+instance Serialise Fees
+
+-- | Setup for fee calculation.
+data FeeCoefficients = FeeCoefficients
+    { fcMultiplier :: Float   -- ^ How much tokens taken per byte of tx?
+    , fcMinimal    :: Coin    -- ^ How much is the base fee?
+    } deriving (Eq, Show)
```
Martoon-00: Probably we won't return to fees soon again, so I think there should a support for multiple fee policies for forward compatibility. Something like 
```
data FeePolicy
    = LinearFeePolicy FeeCoefficients
    | UknownFeePolicy
```
so that it `LinearFeePolicy coefs` serialises to `"0" <> encode coefs`, and carry this `FeePolicy` around instead of `FeeCoefficients`.

wallet/src/Dscp/Wallet/Backend.hs
```diff
@@ -60,24 +64,28 @@ restoreKey sendEvent mName eSecretKey mPassPhrase = do
...
-    nonce <- aiCurrentNonce <$> wGetAccount wc address False
+    nonce <- fromInteger . aiCurrentNonce <$> wGetAccount wc address False
 
-    let inAcc   = TxInAcc { tiaNonce = nonce, tiaAddr   = address }
-        tx      = Tx      { txInAcc  = inAcc, txInValue = inValue, txOuts = outs }
-        inValue = Coin $ sum $ unCoin . txOutValue <$> outs
+    let txSize  = precomputeTxSize (length outs)
+        fees    = calculateFee feeCoefficients txSize
+        inAcc   = TxInAcc { tiaNonce = nonce, tiaAddr = address }
+        inValue = Coin (sum $ map (unCoin . txOutValue) outs) `unsafeAddCoin` unFees fees
```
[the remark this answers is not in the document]
Martoon-00: Right, didn't come to my mind

witness/src/Dscp/Snowdrop/AccountValidation.hs
volhovm: No, that's not a bug. Expander adds fees sent as a separate output, so in the validator we have an extra output, which is counted in `receivedTotal`. I've added a comment about that.

witness/src/Dscp/Snowdrop/Configuration.hs
volhovm: Not sure I understand you. We had `DBMalformedException` which was not an `error` thing.

witness/src/Dscp/Snowdrop/Expanders.hs
volhovm: It seemed that we had it in srk-utils at some point :thinking:

witness/src/Dscp/Snowdrop/AccountValidation.hs
flyingleafe: Okay, it's still not clear from the comment that fees are _already_ included in the outputs after expansion, but fine.

witness/src/Dscp/Snowdrop/Configuration.hs
Martoon-00: Hm, right, it took place.

witness/src/Dscp/Snowdrop/AccountValidation.hs
volhovm: It is not clear from the comment, but the careful_programmer^tm should also check the expander, because some checks and predicates are located there (because expander/verifier framework is not ideal). So when he reads the expander he'll get why it is made so.
)

core/src/Dscp/Core/Fees.hs
```diff
@@ -0,0 +1,75 @@
...
+
+-- | Amount of transaction fees.
+newtype Fees = Fees { unFees :: Coin }
+    deriving (Eq, Ord, Show, Generic, Hashable, Bounded)
+
+instance Serialise Fees
+
+-- | Setup for fee calculation.
+data FeeCoefficients = FeeCoefficients
+    { fcMinimal       :: Coin  -- ^ How much is the base fee?
```
flyingleafe: I believe that minimal fee for publications should be different from minimal fee for money transactions.

PR #126, opened 2018-08-14
[DSCP-199] /hash/{hash} endpoint for requesting hash type

goodsoft:


witness/src/Dscp/Witness/Web/Logic.hs
```diff
@@ -94,3 +95,24 @@ getTransactionInfo =
...
+        (fromHex @HeaderHash)
+        (runSdMRead . getBlockMaybe)
+    isAddress = is
+        HashIsAddress
+        addrFromText
+        getAccountMaybe
+    isTx = is
+        HashIsTx
+        fromHex
+        (runSdMRead . getTxMaybe . GTxId)
```
Martoon-00: Is `getMaybe` part really required?

witness/src/Dscp/Witness/Web/Types.hs
```diff
@@ -22,14 +23,14 @@ data Balances = Balances
       -- ^ From blocks + mempool
     }
 
--- | All what user may wish to know about an account.
```
Martoon-00: Memno

witness/src/Dscp/Witness/Web/Logic.hs
goodsoft: Otherwise, I can't distinguish between block and tx hashes, they are exactly same.

witness/src/Dscp/Witness/Web/Logic.hs
```diff
@@ -94,3 +95,24 @@ getTransactionInfo =
...
     return . toTxInfo
+
+getHashType :: WitnessWorkMode ctx m => Text -> m HashIs
+getHashType hash = fmap (fromMaybe HashIsUnknown) . runMaybeT . asum $
+    [isBlock, isAddress, isTx] <*> [hash]
+  where
+    is hashType decode getMaybe =
+        MaybeT . return . rightToMaybe . decode >=>
+        MaybeT . getMaybe >=>
+        MaybeT . return . Just . const hashType
```
Martoon-00: This is subjective, but here you seem to write it in the long way in purpose.
First line can be written like `Right id <- decode text`, and the third one is just `pure hashType` if written with do-notation.

witness/src/Dscp/Witness/Web/Logic.hs
Martoon-00: Right. I think you should leave a comment about purpose you mentioned, and also mention in swagger that entity should exist.

witness/src/Dscp/Witness/Web/Logic.hs
goodsoft: Not sure I get your idea, but this is as far as my Haskell jutsu gets without multi-storey `case`s

witness/src/Dscp/Witness/Web/Logic.hs
Martoon-00: It's ambiguous whether entity is unknown or type of entity is unknown. I personally though the former at first.

goodsoft: `description: Get type of entity identified by given hash`
If we don't know any entity with this hash, we don't the type of it, and vice versa.
This does not hold for addresses, as their format is less ambiguous, but it's irrelevant for the purposes of this API.

witness/src/Dscp/Witness/Web/Logic.hs
Martoon-00: Since you are in `MaybeT` anyway you could just use partial pattern-matches of do-notation which use `fail` - instead of `case`, but nvm

PR #127, opened 2018-08-14
[DSCP-200] /transactions endpoint for N recent transactions

goodsoft:


witness/src/Dscp/Witness/Logic/Traversals.hs
```diff
@@ -157,3 +159,21 @@ getBlocksBefore depthDiff (headerHash -> newerH) = runExceptT $ do
...
     pure blocks
+
+-- | Retrieves the requested amount of recent transactions
+getTxs ::
+       HasWitnessConfig
+    => Int
+    -> SdM (OldestFirst [] GTxInBlock)
+getTxs depth = getTipBlock >>= map OldestFirst . loadTxs [] depth
+  where
+    loadTxs !res depthLeft Block{..}
```
Martoon-00: This bang pattern doesn't help match since you add several transactions per iteration :thinking:

goodsoft: Don't think we'll have too many transactions per block any time soon.

PR #128, opened 2018-08-15
[DSCP-192] Solve NAT-related issue, add host resolving

volhovm:


witness/src/Dscp/Resource/Network.hs
```diff
@@ -83,8 +84,10 @@ instance WithNetLogging => AllocResource NetCliParams NetCliResources where
         buildComponentR "netcli" allocate release
       where
         allocate = do
+            ncPeers' <- liftIO $ Set.fromList <$> mapM mkZTNodeId (Set.toList ncPeers)
```
Heimdell: Had to look in to recognise that this is a hand-made `mapM` on `Set`.

I'd make `mapMSet` function somewhere and move it later to some `*.Util`.

witness/src/Dscp/Resource/Network.hs
```diff
@@ -123,9 +126,15 @@ instance WithNetLogging => AllocResource NetServParams NetServResources where
         buildComponentR "netcli" allocate release
       where
         allocate = do
+            nsPeers' <- liftIO $ Set.fromList <$> mapM mkZTNodeId (Set.toList nsPeers)
+            nsOurAddress' <- liftIO $ mkZTNodeId nsOurAddress
+            nsInternalAddress' <-
+                liftIO $ maybe (pure Nothing) (fmap Just . mkZTNodeId) nsInternalAddress
```
Heimdell: Isn't `maybe (... Nothing) (... Just ...)` a `fmap x`? Or, maybe whole expression can be written as `(fmap.fmap) mkZTNodeId nsInternalAddress`?

witness/src/Dscp/Resource/Network.hs
```diff
@@ -123,9 +126,15 @@ instance WithNetLogging => AllocResource NetServParams NetServResources where
...
+
             global <- ztGlobalEnv (unNetLogging netLogging)
-            cli <- createNetCliEnv global nsPeers
-            serv <- createNetServEnv global nsOurAddress nsInternalAddress
+            cli <- createNetCliEnv global nsPeers'
+            serv <- createNetServEnv global nsOurAddress' nsInternalAddress'
```
Heimdell: Also, lets not use names with `'`. Its easy to forgot `'` and get yourself incorrect behaviour or bottom and an hour of wtf-debug.

Lets call them `new-` and `old-` instead. Or `src-`/`tmp-`/`dst-`.

volhovm: ok

PR #131, opened 2018-08-15
[DSCP-201] Add missing stuff to BE API

goodsoft:


core/src/Dscp/Core/Slotting.hs
```diff
@@ -25,6 +26,9 @@ getCurrentSlot = do
     curTime <- getTimeMcs
     pure $ SlotId $ curTime `div` slotLength
 
+getSlotSince :: HasCoreConfig => SlotId -> Word64
```
volhovm: Comment?

core/src/Dscp/Core/Foundation/Witness.hs
```diff
@@ -254,7 +254,7 @@ unGTxWitnessed (GPublicationTxWitnessed tw) = GPublicationTx (ptwTx tw)
 
 -- | Transaction with reference to block it is published in
 data GTxInBlock = GTxInBlock
-    { tbHeaderHash :: Maybe HeaderHash
+    { tbBlock :: Maybe Block
```
volhovm: Please explain this change. Also why is it in `core/../Witness.hs`?

witness/src/Dscp/Witness/Web/Types.hs
```diff
@@ -55,6 +55,7 @@ data HashIs
     | HashIsBlock
     | HashIsAddress
     | HashIsTx
```
volhovm: Then this should be `HashIsMoneyTx` right?

witness/src/Dscp/Witness/Logic/Traversals.hs
```diff
@@ -164,16 +165,26 @@ getBlocksBefore depthDiff (headerHash -> newerH) = runExceptT $ do
 getTxs ::
```
volhovm: Add a comment.

witness/src/Dscp/Witness/Web/Logic.hs
```diff
@@ -77,11 +77,11 @@ toTxInfo tx = TxInfo
 
 getBlocks :: WitnessWorkMode ctx m => Maybe Int -> Maybe HeaderHash -> m [BlockInfo]
 getBlocks mCount mFrom = do
-    let count = max 100 $ fromMaybe 100 mCount
+    let count = min 100 $ fromMaybe 100 mCount
     from <- maybe (runSdMRead getTipHash) return mFrom
     runSdMRead (getBlockMaybe from) >>= void . nothingToThrow BlockNotFound
     eBlocks <- runSdMRead $ getBlocksBefore count from
-    either (throwM . InternalError) (return . map (toBlockInfo False) . toList) eBlocks
+    either (throwM . InternalError) (return . map (toBlockInfo False) . reverse . toList) eBlocks
```
volhovm: Are you sure? :thinking:

goodsoft: Yes, because frontend needs newest-first lists.

core/src/Dscp/Core/Foundation/Witness.hs
goodsoft: When I added this type, I only returned hash of block header for transaction in API, but now I need more info about block.
Not sure, whether it belongs to `core`, should I create something like `Dscp.Witness.Types`?

PR #132, opened 2018-08-16
[DSCP-201] BE API Fixes

goodsoft:
- [x] Include sum of tx outputs
- [x] Enable CORS


witness/src/Dscp/Witness/Web/Types.hs
```diff
@@ -115,6 +115,7 @@ instance ToJSON TxInfo where
                 [ "txId" .= toTxId tx
                 , "txType" .= ("money" :: Text)
                 , "money" .= tx
+                , "outValue" .= (Coin . sum . map (unCoin . txOutValue) . txOuts $ tx)
```
Martoon-00: There is `sumCoins`, but ok :man_shrugging:

goodsoft: :man_facepalming:

PR #135, opened 2018-08-16
[DSCP-183] Add `isFinished` flag to `CourseInfo`

Martoon-00:


educator/src/Dscp/Educator/Web/Student/Queries.hs
```diff
@@ -76,6 +76,26 @@ type MonadStudentAPIQuery m =
     , MonadCatch m
     )
 
+studentIsCourseFinished
+    :: MonadStudentAPIQuery m
+    => Id Student -> Course -> m Bool
+studentIsCourseFinished studentId courseId = do
+    finalAssignmentsGrades <- query queryText (CourseFinal, courseId, studentId)
+    return $ any isPositiveGrade finalAssignmentsGrades
```
Heimdell: It says `...IsCourseFinished`, but inside is `any isPositiveGrade`.

So, state where course is not yet succeeded and one where it is ended with negative grade aren't distinguishable.

What about returning `data CourseState = Succeeded | Failed | InProcess`?

volhovm: Failed courses are for now the same as "in process".

PR #136, opened 2018-08-16
[DSCP-197] Add metrics reporting support

hithroc:


witness/src/Dscp/Util/Servant.hs
```diff
@@ -443,3 +447,17 @@ instance (FromJSON a, Reifies err String) => MimeUnrender (SimpleJSON err) a whe
...
         let errMsg = reflect (Proxy @err)
         in first (\_ -> errMsg) . eitherDecode
+
+-------------------------------------------------------------------------
+-- Response time metrics
+-------------------------------------------------------------------------
+
+responseTimeMetric :: Endpoint -> Middleware
+responseTimeMetric endpoint app = \request f -> do
+    t <- getPOSIXTime
```
volhovm: Abstract this and reuse it to measure response time of workers/listeners.

witness/src/Dscp/Witness/CLI.hs
```diff
@@ -70,6 +71,27 @@ netServParamsParser :: Parser NetServParams
...
+----------------------------------------------------------------------------
+-- Metrics server
+----------------------------------------------------------------------------
+
+metricsServerParser :: Parser (Text, Int)
+metricsServerParser =
+  option (maybeReader parseMetricsServer)
+         (long "metrics-server" <>
+          metavar "HOST:PORT" <>
+          value ("127.0.0.1", 8125) <>
```
volhovm: Why is this a default value?

witness/src/Dscp/Witness/CLI.hs
```diff
@@ -70,6 +71,27 @@ netServParamsParser :: Parser NetServParams
...
+          help "Server to send metrics information to"
+         )
+  where
+    parseMetricsServer :: String -> Maybe (Text, Int)
```
volhovm: We already have `host:port` parser implemented in `CommonCLI` module.

witness/src/Dscp/Witness/CLI.hs
hithroc: What else should be the default?

volhovm: There should be no default, the parameter must be optional.

witness/src/Dscp/Witness/CLI.hs
```diff
@@ -70,6 +73,14 @@ netServParamsParser :: Parser NetServParams
 netServParamsParser =
     NetServParams <$> peersParser <*> ourZTNodeIdParser <*> optional internalZTNodeIdParser
 
+----------------------------------------------------------------------------
+-- Metrics server
+----------------------------------------------------------------------------
+
+metricsServerParser :: Parser (Maybe Endpoint)
```
volhovm: If you define `metricsServerParser` without `optional`, it'll be more flexible to use `metricsServerParser` from the outside -- you can always add `optional` in the direct place of usage.

witness/src/Dscp/Witness/Workers/Worker.hs
```diff
@@ -91,9 +91,11 @@ blockUpdateWorker =
            | otherwise -> logDebug "blockUpdateWorker: block is useless"
 
     applyNewBlock block = do
+        params <- view (lensOf @WitnessParams)
         logDebug $ "Block " +| hashF (headerHash block) |+
                   " is a direct continuation of our tip, applying"
-        proof <- applyBlock block
+        proof <- maybe id (reportTime "disciplina.timer.block_apply") (wpMetricsEndpoint params) $
```
volhovm: You can create a wrapper that uses the fact we have `Maybe Endpoint` in the context to avoid this `maybe id`.

PR #138, opened 2018-08-17
TankovaE/faucet-frontend-fix Added background for faucet result block

TankovaE:


romkaspb:
This PR adds transparent background for result block. Shows only when user claimed tokens and got success/error msg.

PR #140, opened 2018-08-17
[DSCP-194] Consider mempool everywhere

Martoon-00:
I advice to review on per-commit basis, each commit has own description.
Also, I wish all the code here to be taken critically (especially in cosmetics), I don't know what I was doing.

Tested:
- [x] Wallet shows "pending" expenses
* [x] It shows them as pending temporally
* [x] Several payments from same account can be made withing just one slot.

### DIFF
--- a/witness/src/Dscp/Witness/SDLock.hs
+++ b/witness/src/Dscp/Witness/SDLock.hs
@@ -1,17 +1,31 @@
 
 module Dscp.Witness.SDLock
     ( SDLock
+    , WithinReadSDLock, WithinWriteSDLock
     , readingSDLock, readingSDLockOf
     , writingSDLock, writingSDLockOf
     , newSDLock
     ) where
 
 import qualified Control.Concurrent.ReadWriteLock as RawLock
+import Data.Reflection (Given, give)
 import qualified UnliftIO
 
 import Loot.Base.HasLens (HasLens (..), HasLens')
 
 newtype SDLock = SDLock RawLock.RWLock
+
+data ReadSDLockTag = ReadSDLockTag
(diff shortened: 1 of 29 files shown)


Martoon-00:
Could you elaborate on "aligning code in columns"?

goodsoft:
I mean this: 
image

Martoon-00:
Meh, I see. `stylish-haskell` does it automatically for me according to config file in repo.

Martoon-00:
AFAUI and if I don't mix anything up, this happens because wallet can't communicate with witnesses on cluster.

goodsoft:
I think it might be because the wallet in this PR is not compatible with the deployed cluster.
You should run your own witness node from this PR.

goodsoft:
Actually, @Martoon-00, please update `explorer.yaml` spec with regards to your changes.

Martoon-00:
Right, thanks :+1:

witness/src/Dscp/Witness/SDLock.hs
```diff
@@ -1,63 +1,80 @@
...
+import Loot.Log (MonadLogging)
 import qualified UnliftIO
 
-import Loot.Base.HasLens (HasLens (..), HasLens')
+import Dscp.Util.TimeLimit (logWarningWaitInf)
 
+-- | Lock for SD operations.
 newtype SDLock = SDLock RawLock.RWLock
 
+data ReadSDLockTag = ReadSDLockTag
```
volhovm: Oh my god, this is actually a very good idea. cc @nasportivnoy

Martoon-00:
Thanks, fixed them.

PR #142, opened 2018-08-19
[DSCP-191] Move AppDirectory from configuration to CLI params

nasportivnoy:


yorickvP:
As far as I can tell, the app directory isn't used for anything. Correct?

nasportivnoy:
@yorickvP the directory is used in `../(Witness/Faucet/Educator)/Launcher/Resource.hs` which import `link Store` function from`Dscp.Resource.Keys.Functions` to  find where keyfile lies.

yorickvP:
Then what are the various --keyfile options for?

nasportivnoy:
Do you mean `--witness-keyfile`? Afaiu this option does nothing. I've just launched the demo script and submitted invalid witness key file in `--witness-keyfile` and it worked fine.

witness/src/Dscp/CommonCLI.hs
```diff
@@ -77,6 +79,14 @@ baseKeyParamsParser who = do
          help "Password of secret key."
     passphraseReadM = leftToFail . first pretty . mkPassPhrase =<< str
 
+appDirParamParser :: Parser AppDirParam
+appDirParamParser = AppDirectorySpecific <$>
+                        (strOption $
+                        long "path" <>
```
volhovm: Any other parameter name maybe?

witness/src/Dscp/Resource/Keys/Functions.hs
```diff
@@ -47,11 +48,8 @@ fromSecretJson pp KeyJson{..} = do
 -- | Where keyfile would lie.
 storePath
     :: (HasWitnessConfig, Buildable (Proxy node))
-    => BaseKeyParams -> Proxy node -> FilePath
-storePath BaseKeyParams{..} nodeNameP =
-    fromMaybe defPath bkpPath
-  where
-    defPath = giveLC @'["core", "generated", "home"] @WitnessConfig </> (nodeNameP |+ ".key")
+    => AppDir -> Proxy node -> FilePath
+storePath appDir nodeNameP = appDir </> (nodeNameP |+ ".key")
```
goodsoft: @nasportivnoy, removing `fromMaybe` here broke `--faucet-keyfile` and similar options which allowed specifying custom secrets location.
I fixed that regression in #174.
Still, if you had some ulterior motives for that, some explanation would really help.

PR #145, opened 2018-08-19
[SRK-8] Make `--witness-listen` optional

Martoon-00:


Martoon-00:
@volhovm Please check this PR again, I added witness workers and listeners to Main of educator, and also added witness api into educator server.

educator/src/Dscp/Educator/Web/Server.hs
```diff
@@ -66,16 +68,17 @@ createGetStudentsAction = do
...
     let srvCtx = getStudents :. EmptyContext
 
     logInfo $ "Serving Student API on "+|spAddr|+""
-    eCtx <- ask
-    let educatorApiServer = mkEducatorApiServer (convertEducatorApiHandler eCtx)
-    studentApiServer <- mkStudentApiServer (convertStudentApiHandler eCtx) ewpBotParams
+    unliftIO <- askUnliftIO
+    let educatorApiServer = mkEducatorApiServer (convertEducatorApiHandler unliftIO)
+    studentApiServer <- mkStudentApiServer (convertStudentApiHandler unliftIO) ewpBotParams
+    let witnessApiServer = mkWitnessAPIServer (convertWitnessHandler unliftIO)
```
volhovm: So again, it just serves there independently of the parameter passed, which is not what we want. Just make the `--witness-listen` optional:
* If it's not present, we don't launch the api
* If it's present and different from educator api, we launch it separately
* Otherwise we launch it at the same port

Martoon-00:
Like this?

educator/src/Dscp/Educator/Launcher/Entry.hs
```diff
@@ -0,0 +1,43 @@
...
+        whenJust (wpWitnessServerParams witnessParams) $ \serverParams ->
+            when separateWitnessServer $ do
+                logInfo "Forking witness API server"
+                void . async $
+                    serveWitnessAPIReal serverParams
+
+        logInfo "Forking student API"
+        void . async $ serveEducatorAPIsReal
+            (not separateWitnessServer)
+            (epWebParams educatorParams)
```
volhovm: How is it going to work if the same port is specified for educator API and witness API?

PR #146, opened 2018-08-20
[DSCP-208] Prevented application/validation exns from leaking out

Heimdell:
- Also `failedTxs` `TVar` is now filled with txs.
- But is never cleaned.

### DIFF
--- a/witness/src/Dscp/Witness/Mempool/Logic.hs
+++ b/witness/src/Dscp/Witness/Mempool/Logic.hs
@@ -8,6 +8,7 @@ module Dscp.Witness.Mempool.Logic
     ) where
 
 import UnliftIO (MonadUnliftIO)
+import Control.Monad.Except (catchError)
 
 import Loot.Base.HasLens (HasLens', lensOf)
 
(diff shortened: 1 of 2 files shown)


witness/src/Dscp/Witness/Mempool/Logic.hs
```diff
@@ -44,10 +45,13 @@ addTxToMempool
     .  MempoolCtx ctx m
     => MempoolVar
     -> GTxWitnessed
-    -> m ()
-addTxToMempool (Mempool pool conf, lock) tx =
-    writeToMempool @ctx pool lock
-        $ Pool.processTxAndInsertToMempool conf tx
+    -> m Bool
+addTxToMempool (Mempool pool conf, lock) tx = do
+    writeToMempool @ctx pool lock $ do
```
volhovm:
```
writeToMempool @ctx pool lock $ 
  True <$ Pool.processTxAndInsertToMempool conf tx `catchError` (\_ -> pure False)
```

witness/src/Dscp/Witness/Workers/Worker.hs
```diff
@@ -144,5 +144,9 @@ makeRelay (RelayState input pipe failedTxs) =
             isThere <- isInMempool @ctx pool tx
 
             unless isThere $ do
-                addTxToMempool @ctx pool tx
+                good <- addTxToMempool @ctx pool tx
+
+                unless good $ do
```
volhovm: Redundant do.

witness/src/Dscp/Witness/Mempool/Logic.hs
Heimdell: This looks more like obfuscation, then style improvement.

flyingleafe: Dunno, maybe:
```
writeToMempool @ctx pool lock $
    True <$ Pool.processTxAndInsertToMempool conf tx
  `catchError`
    \_ -> pure False
```
?
:man_shrugging: :woman_shrugging: :man_shrugging:

flyingleafe: Frankly, everything looks bad, but @Heimdell's variant is easier to read.

PR #147, opened 2018-08-20
[SRK-8] Fix advanced courses in bot

Martoon-00:
Since student cannot send more than one submission for the same assignment for now,
easter egg with providing advanced courses on 3 submissions for the same assignment
didn't work. But not anymore.

### DIFF
--- a/educator/src/Dscp/Educator/Web/Bot/Handlers.hs
+++ b/educator/src/Dscp/Educator/Web/Bot/Handlers.hs
@@ -54,17 +54,22 @@ addBotHandlers student StudentApiEndpoints{..} = botEndpoints
             botProvideInitSetting student
             res <- sMakeSubmission newSub
 
-            delayed $ requestToSignedSubmission newSub >>= botGradeSubmission
+            delayed $ do
+                requestToSignedSubmission newSub >>= botGradeSubmission
 
-            allAssigns <- sGetAssignments Nothing Nothing Nothing
-            botProvideUnlockedAssignments student res allAssigns
+                allAssigns <- sGetAssignments Nothing Nothing Nothing
+                botProvideUnlockedAssignments student res allAssigns
 
-            -- cheat: on 3 submissions for the same assignment unlock all courses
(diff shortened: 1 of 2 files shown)


educator/src/Dscp/Educator/Web/Bot/Handlers.hs
```diff
@@ -52,19 +52,21 @@ addBotHandlers student StudentApiEndpoints{..} = botEndpoints
...
         , sMakeSubmission = \newSub -> do
             botProvideInitSetting student
-            res <- sMakeSubmission newSub
-
-            delayed $ requestToSignedSubmission newSub >>= botGradeSubmission
-
-            allAssigns <- sGetAssignments Nothing Nothing Nothing
-            botProvideUnlockedAssignments student res allAssigns
 
             -- cheat: on 3 submissions for the same assignment unlock all courses
```
volhovm: I do not understand how's that a fix. You've just moved a piece of code giving default assignments after the `when (length ...)` check.

flyingleafe: :heavy_plus_sign: , can you elaborate how this allows for multiple submissions for assignments?

Martoon-00: It does not (and afaiu we decided not to support multiple submissions on same assignment).
It just allows to make consequent failing submissions of the same assignment which would unlock advanced courses.

volhovm:
You say:

> (the) easter egg with providing advanced courses on 3 submissions for the same assignment
didn't work. But not anymore.

So you imply now we have this easter egg working. But in reality, this PR doesn't even change bot's behaviour.

educator/src/Dscp/Educator/Web/Bot/Handlers.hs
volhovm: >It just allows to make consequent failing submissions of the same assignment which would unlock advanced courses.

Can we just find a better way to unlock extra courses?

PR #151, opened 2018-08-21
[DSCP-195] Store transaction lists in pure storage instead of AVL

goodsoft:
Blergh


goodsoft:
I forgot to move `tx -> headerhash` expansion to expander, so WIP

PR #152, opened 2018-08-21
[DSCP-209] Check account has enough money on witness API side

Martoon-00:
Mim

### DIFF
--- a/specs/disciplina/witness/api/witness.yaml
+++ b/specs/disciplina/witness/api/witness.yaml
@@ -367,6 +367,7 @@ definitions:
             * `InvalidFormat` - Failed to deserialise one of parameters.
 
             * `BlockNotFound` - Specified block does not exist.
+            * `NotEnoughMoney` - Transaction's source account has not enough money to make this transaction.
             * `AccountNotFound` - Specified account does not exist.
             * `TransactionNotFound` - Specified transaction does not exist.
 
--- a/specs/disciplina/witness/api/witness.yaml
+++ b/specs/disciplina/witness/api/witness.yaml
@@ -14,6 +14,13 @@ schemes:
   - https
   - http
 paths:
+  /ping:
+    get:
+      summary: Ping a node
+      operationId: ping
+      responses:
+        200:
+          description: Successful operation.
   /hash/{hash}:
     get:
       summary: Get type of entity identified by given hash
@@ -180,6 +187,31 @@ paths:
                   - $ref: "#/definitions/TransactionBlockInfo"
               nextId:
                 $ref: "#/definitions/Hash"
+
+    post:
+      summary: Submit a formed transaction.
+      operationId: submitTx
+      produces:
+      - application/json
+      parameters:
+      - in: body
+        name: txWitnessed
+        schema:
+          $ref: "#/definitions/TxWitnessed"
+      responses:
+        201:
+          description: Transaction passed validation. If no transactions spending money from the same account are submitted concurrently to other witness node, transaction will eventually appear in blockchain. Returns transaction id.
+          schema:
+            $ref: "#/definitions/Hash"
+        400:
+          description: Formed transaction is invalid.
+          schema:
+            $ref: '#/definitions/TxValidationError'
+        403:
+          description: Transaction cannot be applied.
+          schema:
+            $ref: '#/definitions/TxValidationError'
+
   /transactions/{transactionHash}:
     get:
       summary: Get information about a transaction.
@@ -211,11 +243,28 @@ paths:
           description: Transaction not found
           schema:
             $ref: "#/definitions/ErrResponse"
+  /transactions/async:
+    post:
+      summary: Submit a formed transaction asynchronously.
+      operationId: submitTxAsync
+      produces:
+      - application/json
+      parameters:
+      - in: body
+        name: txWitnessed
+        schema:
+          $ref: "#/definitions/TxWitnessed"
+      responses:
+        202:
+          description: Transaction accepted. Returns transaction id.
+          schema:
+            $ref: "#/definitions/Hash"
 
 definitions:
   Hash:
     type: string
     format: byte
+    description: Hex-encoded hash value.
   Address:
     type: string
     format: byte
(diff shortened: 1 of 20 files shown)


witness/src/Dscp/Witness/Web/Logic.hs
```diff
@@ -29,17 +29,31 @@ import Dscp.Witness.Web.Types
...
 
+-- | Check that source account has enough money to make transaction.
+ensureEnoughMoney :: WitnessWorkMode ctx m => TxWitnessed -> m ()
+ensureEnoughMoney tw = do
+    let tx     = twTx tw
+        spent  = txInValue tx
+        source = tiaAddr (txInAcc tx)
+    account <- fromMaybe def <$> getMempoolAccountMaybe source
+
+    when (aBalance account < coinToInteger spent) $
```
goodsoft: Does this take fees into account?

specs/disciplina/witness/api/explorer.yaml
```diff
@@ -367,6 +367,7 @@ definitions:
             * `InvalidFormat` - Failed to deserialise one of parameters.
 
             * `BlockNotFound` - Specified block does not exist.
+            * `NotEnoughMoney` - Transaction's source account has not enough money to make this transaction.
```
goodsoft: We should probably add tx submitting endpoint to the docs too, at this point

witness/src/Dscp/Witness/Web/Error.hs
```diff
@@ -28,6 +29,7 @@ instance Buildable WitnessAPIError where
     build = \case
         BlockNotFound -> "Specified block does not exist."
         TransactionNotFound -> "Specified transaction does not exist."
+        NotEnoughMoney -> "Account has not enough money to make such transaction"
```
goodsoft: `Not enough money to perform this transaction.`, imho

witness/src/Dscp/Witness/Web/Logic.hs
```diff
@@ -29,17 +29,31 @@ import Dscp.Witness.Web.Types
...
+submitUserTx tw = do
+    ensureEnoughMoney tw
     Relay.relayTx (GMoneyTxWitnessed tw) >>= wait @"tx in mempool"
 
 -- | Applies transaction, but does not wait for a whole cycle of transaction
 -- application.
 submitUserTxAsync :: WitnessWorkMode ctx m => TxWitnessed -> m ()
-submitUserTxAsync = void . Relay.relayTx . GMoneyTxWitnessed
+submitUserTxAsync tw = do
+    ensureEnoughMoney tw
```
Heimdell: By the way, I did it in validator.

Martoon-00: Yes, but as far as I understand it happens beyond relay, and transaction will just silently fail unless I check it once again here.

specs/disciplina/witness/api/explorer.yaml
Martoon-00: Oh, it's not there after all.
Would it be fair to add such endpoint to `explorer.yaml`? :thinking:

witness/src/Dscp/Witness/Web/Logic.hs
Martoon-00: Right. I guess I'll wait until #125 get merged

specs/disciplina/witness/api/explorer.yaml
goodsoft: Actually, there's no particular reason to consider "block explorer API" some special part of witness API, so this should probably be renamed to `witness.yaml` or something

witness/src/Dscp/Witness/Web/Logic.hs
Heimdell: They'd rather be put into `failedTxs` map.

Martoon-00: Would would be a proper way to do that?
AFAIU `failedTxs` is available in worker context only, should I move it to node context?

flyingleafe: This seems like a kostyl. I think that the proper way of doing this is to rethrow exceptions through relay somehow. Otherwise we'll have to build such kostyls for every reasonable error which might happen with a transaction (I guess "invalid signature" goes here too, i. e. now API would just silently fail).
We might leave it as is if doing things properly turns out to be really complicated, but we _need_ to do it eventually.

Martoon-00:
That's weird. We put transactions into channel and let worker process them sequentially just to... later take a lock, block until worker notifies about transaction being put in mempool and transfer exceptions from behind the relay.
But ok, let me come up with something.

flyingleafe:
@Martoon-00 Relay exists for proper _asynchronous_ processing of transactions, I assume, so when we'd like to process them synchronously we need to do this weird stuff, right.

witness/src/Dscp/Witness/Web/Logic.hs
```diff
@@ -33,13 +33,14 @@ import Dscp.Witness.Web.Types
...
-submitUserTx tw =
+submitUserTx tw = do
     Relay.relayTx (GMoneyTxWitnessed tw) >>= wait @"tx in mempool"
 
 -- | Applies transaction, but does not wait for a whole cycle of transaction
 -- application.
 submitUserTxAsync :: WitnessWorkMode ctx m => TxWitnessed -> m ()
-submitUserTxAsync = void . Relay.relayTx . GMoneyTxWitnessed
+submitUserTxAsync tw = do
+    void . Relay.relayTx $ GMoneyTxWitnessed tw
```
goodsoft: Why those redundant `do`s?

specs/disciplina/witness/api/witness.yaml
```diff
@@ -180,6 +187,29 @@ paths:
...
+      produces:
+      - application/json
+      parameters:
+      - in: body
+        name: txWitnessed
+        schema:
+          $ref: "#/definitions/TxWitnessed"
+      responses:
+        200:
+          description: Transaction passed validation. If no transactions spending money from the same account are submitted concurrently to other witness node, transaction will eventually appear in blockchain.
```
goodsoft: We may probably want to return a tx id _just in case_
Also, wasn't it `201`?

Martoon-00:
Yep, something like that.

Martoon-00:
For other reviewers: briefly speaking, now I had to describe and handle all kinds of `AccountValidationException` (with hope that other snowdrop exceptions are programmer errors, assuming transaction `expansion` is performed by our node, thus they treated like internal errors in witness API), thus PR became a bit big.

specs/disciplina/witness/api/witness.yaml
```diff
@@ -180,6 +187,31 @@ paths:
                   - $ref: "#/definitions/TransactionBlockInfo"
               nextId:
                 $ref: "#/definitions/Hash"
+
+    post:
```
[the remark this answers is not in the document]
Martoon-00: Lol, noice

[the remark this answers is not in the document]
Martoon-00: Looks like a weird bug in github actually, my local version is different, and if you view this file not in diff then you'll see it correct.

Martoon-00:
Thanks, looks much better now.

specs/disciplina/witness/api/witness.yaml
```diff
@@ -367,10 +462,36 @@ definitions:
...
+          - SumMustBeNonNegative
+        description: >-
+          Error type:
+            * `InvalidFormat` - Failed to deserialise one of parameters.
+            * `AuthorDoesNotExist` - Source account has never received money.
+            * `SignatureIsCorrupted` - Bad signature provided.
+            * `NotASingletonSelfUpdate` - Cannot send money to yourself more than once in transaction.
+            * `NonceMustBeIncremented` - Bad nonce for source account is used. Should be exactly 'nextNonce' returned by /accounts/{address} endpoint.
+            * `PaymentMustBePositive` - You can only decrease your funds by outcoming transaction.
+            * `ReceiverMustIncreaseBalance` - You can only increase funds of receiver by outcoming transaction.
```
Heimdell: "increase _the_ funds", my bad.

specs/disciplina/witness/api/witness.yaml
```diff
@@ -367,10 +462,36 @@ definitions:
...
+            * `SumMustBeNonNegative` - Sum of payments plus fee must be less or equal than amount of money you spend.
```
Heimdell: "less or equal _to_"

specs/disciplina/witness/api/witness.yaml
Martoon-00: I wonder how many articles are actually missing in the spec :thinking:

PR #155, opened 2018-08-23
[SD-133] Make disciplina node implementation work with latest `master` of snowdrop

pva701:


witness/src/Dscp/Witness/Logic/Processing.hs
```diff
@@ -57,17 +58,20 @@ createBlock newSlot = do
...
+    let stateDBM = nsStateDBActions sdActions (SD.RememberForProof True)
 
     (blockCS, stateCS) <- reify blockPrefixes $ \(_ :: Proxy ps) ->
         let actions = SD.constructCompositeActions @ps (SD.dmaAccessActions blockDBM)
                                                         (SD.dmaAccessActions stateDBM)
             rwComp = do
               sblock <- SD.liftERoComp $ expandBlock block
-              SD.applyBlockImpl toVerify blkStateConfig sblock
+              -- getCurrentTime requires MonadIO
+              let osParams = SD.unOSParamsBuilder sdOSParamsBuilder $ UTCTime (toEnum 0) (toEnum 0)
```
georgeee: Looks weird... Maybe we quickly perform refactoring for osparams to integrate them properly

pva701: ahahahah nu nu, quickly.

pva701: it looks as weird as implementation of os params in snowdrop, imho

pva701: >Maybe we quickly perform refactoring for osparams to integrate them properly

it's not so easy, I think. you should somehow manage os params, I mean create them, recalculate in some moments, IMHO, it's not so easy to design them properly. abstract type is the smallest thing in this refactoring.

georgeee: Let's at least pass start time as UtcTime 0 0 so it's clear it's not used anyhow

pva701:
@georgeee done

PR #156, opened 2018-08-23
[SD-135] Make disciplina node implementation work with latest `master` of snowdrop

astiunov:
Continuation of #155


volhovm:
This should be pointed towards `master`, we do not plan to have rollbacks in alpha.

core/src/Dscp/Crypto/Impl.hs
```diff
@@ -75,6 +80,27 @@ type PublicKey = AbstractPK SigScheme
 type SecretKey = AbstractSK SigScheme
 type Signature a = AbstractSig SigScheme a
 
+data DscpSigScheme
```
volhovm: Snowdrop is only allowed to be used in `witness`, and it'd be better if it's only inside `Dscp.Snowdrop`, so move this elsewhere pls.

core/src/Dscp/Crypto/Impl.hs
```diff
@@ -75,6 +80,27 @@ type PublicKey = AbstractPK SigScheme
...
+data instance SD.PublicKey DscpSigScheme = DscpPK PublicKey
```
volhovm: Also, I do not understand this design decision. Why is it an open data family? Why not a closed one? Why not a type family?

witness/src/Dscp/Snowdrop/Actions.hs
```diff
@@ -40,23 +39,24 @@ initSDActions ::
...
             (respV :: AVL.MapLayer AvlHash Ids Values AvlHash) <- deserialiseM resp'
             let respProof = AVL.Proof . Free $ pure <$> respV
             when (not $ AVL.checkProof h respProof) $ throwM BrokenProofError
           pure resp
-    let (rootHash :: RootHash) = gett avlInitState
+    let (_rootHash :: RootHash AvlHash) = gett avlInitState
 
     -- What am I supposed to do with it?
-    _clientStDba <- avlClientDbActions @Ids @Values retrieveF rootHash
+    -- pva701: it will be used in a AVL client, keep it compatible
```
volhovm: I know it's for "AVL client" so I keep it compatible, but I doubt we'll ever need AVL client.

witness/src/Dscp/Snowdrop/Configuration.hs
```diff
@@ -26,17 +27,15 @@ import Dscp.Witness.Logic.Exceptions (LogicException)
...
-data SPayload = SPayload { sPayStateTxs :: [StateTx Ids Proofs Values]
-                         , sPayOrigBody :: T.BlockBody
+data SPayload = SPayload { sPayStateTxs     :: ![StateTx Ids Proofs Values]
+                         , sPayOrigBodyHash :: !(Hash T.BlockBody)
                          } deriving (Eq, Show, Generic)
 type SBlock   = Block SHeader SPayload
-type SUndo    = ChangeSet Ids Values
-type SBlund   = Blund SHeader SPayload SUndo
+type SUndo    = Undo Ids Values
+type SBlund   = Blund SHeader T.BlockBody SUndo
```
volhovm: Shylo na mylo.

witness/src/Dscp/Snowdrop/Storage/Avlp.hs
```diff
@@ -1,440 +0,0 @@
...
-{-# LANGUAGE NamedFieldPuns      #-}
-{-# LANGUAGE ScopedTypeVariables #-}
-
--- This module is a complete BARDAQ, please rewrite it.  And after you
--- do, add docs and sections. It's harder to read/understand than you
--- think!
-
--- | AVL plus integration into snowdrop.
-
-module Dscp.Snowdrop.Storage.Avlp
```
volhovm: Akhem:
* Was this ever discussed before? 
* Did you already break everything or we still have time to bring this module back?

cc @Heimdell

witness/src/Dscp/Snowdrop/Validators.hs
```diff
@@ -92,51 +74,24 @@ _baseValidator =
...
-
--- | Same as inmemoryBlkStateConfiguration, but works for our patched
--- SPayload. DSCP-175
-simpleBlkStateConfiguration ::
-       SD.BlkConfiguration SHeader SPayload HeaderHash
-    -> SD.Validator Exceptions Ids Proofs Values (IOCtx chgAccum)
-    -> SD.BlkStateConfiguration SHeader SPayload SUndo HeaderHash
+blkStateConfig
+    :: HasWitnessConfig
+    => SD.BlkStateConfiguration SHeader SPayload BlockBody (SD.Undo Ids Values) HeaderHash
```
volhovm: `SD.Undo Ids Values` is `SUndo`

witness/src/Dscp/Snowdrop/Validators.hs
```diff
@@ -92,51 +74,24 @@ _baseValidator =
...
-    , bscSetTip = \newTip' -> do
-          let newTip = SD.inj $ SD.TipValue newTip'
-          let tipChg = \cons -> SD.ChangeSet $ M.singleton (SD.inj SD.TipKey) (cons newTip)
-          oldTipMb <- SD.liftERoComp $ SD.queryOne SD.TipKey
-          -- TODO check that tip corresponds to blund storage
-          case oldTipMb of
-              Nothing                            -> SD.modifyRwCompChgAccum $ tipChg SD.New
-              Just (_ :: SD.TipValue HeaderHash) -> SD.modifyRwCompChgAccum $ tipChg SD.Upd
+blkStateConfig = blkStateConfig'
+    { SD.bscApplyPayload = \(SPayload txs _) -> SD.bscApplyPayload blkStateConfig' txs
```
volhovm: I do not understand why do we apply payload here with `bcBlkVerify` turned off.

witness/src/Dscp/Witness/Launcher/Mode.hs
```diff
@@ -74,12 +79,13 @@ type WitnessWorkMode ctx m =
...
-    , _wcMempool    :: !MempoolVar
-    , _wcSDActions  :: !SDActions
-    , _wcRelayState :: !RelayState
-    , _wcSDLock     :: !SDLock
+    { _wcParams          :: !WitnessParams     -- ^ Parameters witness was started with.
+    , _wcResources       :: !WitnessResources  -- ^ Resources, allocated from params.
+    , _wcMempool         :: !MempoolVar
+    , _wcSDActions       :: !SDActions
+    , _wcRelayState      :: !RelayState
+    , _wcSDParamsBuilder :: !SD.OSParamsBuilder
```
volhovm: Rename `SDActions` into `SDVars` and put this `OSParamsBuilder` inside.

witness/src/Dscp/Witness/Logic/Processing.hs
```diff
@@ -57,17 +58,20 @@ createBlock newSlot = do
...
+    let stateDBM = nsStateDBActions sdActions (SD.RememberForProof True)
 
     (blockCS, stateCS) <- reify blockPrefixes $ \(_ :: Proxy ps) ->
         let actions = SD.constructCompositeActions @ps (SD.dmaAccessActions blockDBM)
                                                         (SD.dmaAccessActions stateDBM)
             rwComp = do
               sblock <- SD.liftERoComp $ expandBlock block
-              SD.applyBlockImpl toVerify blkStateConfig sblock
+              -- getCurrentTime requires MonadIO
+              let osParams = SD.unOSParamsBuilder sdOSParamsBuilder $ UTCTime (toEnum 0) (toEnum 0)
```
volhovm: What does it mean/do?

witness/src/Dscp/Witness/Mempool/Logic.hs
```diff
@@ -129,7 +130,7 @@ readFromMempoolLocked pool lock action = do
 writeToMempool
     :: forall ctx m a
     .  (MempoolCtx ctx m, WithinWriteSDLock)
-    => Pool.Mempool Ids Values (AVLP.AVLChgAccum Ids Values) GTxWitnessed
+    => Pool.Mempool Ids Values (AVLP.AVLChgAccum AvlHash Ids Values) GTxWitnessed
```
volhovm: Please introduce decent type aliases for chgaccums (both), statetx and don't forget undo i've mentioned earlier.

volhovm:
Oh yeah, after a moment of consideration I think it makes sense to merge this PR to alpha branch anyway. It doesn't introduce any changes as it is, though having DSCP compatible with SD is good, as SD devs will be able to fix problems more quickly if they arise.

witness/src/Dscp/Snowdrop/Storage/Avlp.hs
georgeee: What? You had a terrible copy-pasta in this module. What do you want to discuss?

witness/src/Dscp/Witness/Logic/Processing.hs
georgeee: Now it does nothing. This feature is called OS paras handling and is needed e.g. to verify that all blocks in fork are not from the future slot. Unfortunately the way the feature was implemented doesn't allow for easy integration with DSCP, so we left this placeholder for now (and another one in `bcValidateFork`.

witness/src/Dscp/Snowdrop/Validators.hs
georgeee: Sorry, didn't understand the question

core/src/Dscp/Crypto/Impl.hs
georgeee: > Why not a type family?

There were some problems with `VerifySign` type class and making `PublicKey` and `Signature` data families was the easiest (while still seamless) solution.

> Why not a closed one?

Well, because there could be a universe of various signature schemes.

core/src/Dscp/Crypto/Impl.hs
georgeee: @astiunov please fix

witness/src/Dscp/Snowdrop/Validators.hs
georgeee: @astiunov please fix

witness/src/Dscp/Witness/Launcher/Mode.hs
georgeee: @astiunov please fix

witness/src/Dscp/Witness/Mempool/Logic.hs
georgeee: @astiunov please fix

core/src/Dscp/Crypto/Impl.hs
volhovm: No, i mean the `class SigScheme where data SSPub, data SSpriv`... Currently i can define public key without defining a secret one, which is weird.

witness/src/Dscp/Snowdrop/Validators.hs
volhovm: I asked someone to explain this logic. Why `bscApplyPayload` uses modified `blkStateConfig`?

witness/src/Dscp/Snowdrop/Storage/Avlp.hs
volhovm: At least we had control over it. It's up to Kirill to decide, I think.

core/src/Dscp/Crypto/Impl.hs
georgeee: > Currently i can define public key without defining a secret one, which is weird.

Because for the sake of validation (which is main concern of snowdrop to date) we don't need a secret key.

witness/src/Dscp/Snowdrop/Storage/Avlp.hs
georgeee: I disagree. "We control" is not an excuse for maintaining a number of code duplicates. You're always welcome to make a PR against snowdrop if some change is desirable.

witness/src/Dscp/Snowdrop/Validators.hs
georgeee: Oh, I see. A very good point indeed!

@pva701 or @astiunov ?

core/src/Dscp/Crypto/Impl.hs
volhovm: Yes, but the signature scheme with pk without sk doesn't make any sense at all. It doesn't have anything to do with what you ask in your methods, it's just there should be a single place to define pk and sk. 

Let's end this conversation, we got sidetracked a bit.

witness/src/Dscp/Snowdrop/Validators.hs
astiunov: This part from @pva701

core/src/Dscp/Crypto/Impl.hs
astiunov: But it's already used in https://github.com/DisciplinaOU/disciplina/blob/release-alpha/core/src/Dscp/Util.hs#L62

witness/src/Dscp/Snowdrop/Validators.hs
pva701: @volhovm `bscApplyPayload` doesn't use `bcBlkVerify`, it can't use it even in theory, because payload is part of block, but `bcBlkVerify` verifies whole block. 

`blkStateConfig'` is needed to match type of `inmemoryBlkStateConfiguration` and `blkStateConfig`

core/src/Dscp/Crypto/Impl.hs
volhovm: It's completely different. You should have put `OldestFirst` and `NewestFirst` in lootbox instead, but since you define it in sd we're forced to re-export it from our utility module.

witness/src/Dscp/Snowdrop/Validators.hs
volhovm: >because payload is part of (a) block, but bcBlkVerify verifies (the) whole block.

So why do you change it to `mempty` then?

> A is needed to match (the) type of B

I don't understand this pattern-matching game, can you please elaborate on the exact semantics of this change?

pva701: >So why do you change it to mempty then?

because `bcBlkVerify` isn't used in `inmemoryBlkStateConfiguration` anyway, the only place where it's used it's setting `bscConfig`, but I overwrite it

>I don't understand this pattern-matching game, can you please elaborate on the exact semantics of this change?

the exact semantic is: the logic of `blkStateConfig` is the same as `inmemoryBlkStateConfiguration` and we want to abuse this fact, but `inmemoryBlkStateConfiguration` has a little bit type of payload, and I hacked all places which are affected by the payload type (`bscApplyPayload` and `bscConfig`). It can be done in this way because `cfg` isn't used in functions of `BlkStateConfiguration` (ok it's used but in only one place where we store blund, but it's ok). It's done to avoid copy-pasting of code from snowdrop.

pva701: ok, we are going to fix it @volhovm

shell.nix
```diff
@@ -26,6 +26,8 @@ in
       binutils
       git
       rocksdb
+      snappy
+      google-gflags
```
volhovm: Sorry, what exactly wants these deps?

witness/package.yaml
```diff
@@ -12,6 +12,7 @@ library:
     - autoexporter
     - AVL
     - bytestring
+    - cereal
```
volhovm: We don't use `cereal`, we only use `serialise`.

witness/src/Dscp/Snowdrop/Validators.hs
```diff
@@ -6,16 +6,14 @@ module Dscp.Snowdrop.Validators
...
 import qualified Snowdrop.Core as SD
-import qualified Snowdrop.Model.Block as SD
 import qualified Snowdrop.Util as SD
 
 import Dscp.Core hiding (PublicationTxWitness)
-import Dscp.Crypto (PublicKey, Signature, hash, verify)
+import Dscp.Crypto (PublicKey, hash, verify)
 import Dscp.Snowdrop.AccountValidation
-import Dscp.Snowdrop.Configuration
+import Dscp.Snowdrop.Configuration hiding (PublicKey)
```
volhovm: `Dscp.Snowdrop.Configuration` doesn't export `PublicKey`. If it does, that's a mistake.

witness/src/Dscp/Witness/Launcher/Mode.hs
```diff
@@ -74,12 +78,12 @@ type WitnessWorkMode ctx m =
...
-    , _wcMempool    :: !MempoolVar
-    , _wcSDActions  :: !SDActions
-    , _wcRelayState :: !RelayState
-    , _wcSDLock     :: !SDLock
+    { _wcParams          :: !WitnessParams     -- ^ Parameters witness was started with.
+    , _wcResources       :: !WitnessResources  -- ^ Resources, allocated from params.
+    , _wcMempool         :: !MempoolVar
+    , _wcSDActions       :: !SDVars
+    , _wcRelayState      :: !RelayState
+    , _wcSDLock          :: !SDLock
```
volhovm: Good news: these spaces are not needed.
Bad news: somebody still hasn't added `stylish-haskell` to save hook.

PR #163, opened 2018-08-24
[DSCP-207] Refresh balance after successful tx and also periodically

goodsoft:


Martoon-00:
Ueeee, kaef

wallet/exec/Main.hs
```diff
@@ -54,4 +57,10 @@ initializeEverything serverAddress = do
         uiAction :: IO ()
         uiAction = mkUiAction (knitFaceToUI uiWalletState uiFace walletFace knitFace)
 
-    return uiAction
+        refreshAction :: IO ()
+        refreshAction = forever $ do
+            delay <- randomRIO (15, 45)
```
Martoon-00: Slot duration in production will be 20s, I believe we can make this less.
Poor @ArgentumHeart may just fall asleep while waiting for 40 seconds.

goodsoft: 15..25 then

goodsoft:
Dunno, we have feature freeze and all.
I'm committing everything to `master` unless specifically told to do otherwise.

PR #166, opened 2018-08-27
[DSCP-189] Safer CORS

Martoon-00:


faucet/src/Dscp/Faucet/Web/Server.hs
```diff
@@ -30,15 +31,21 @@ mkFaucetApiServer
 mkFaucetApiServer nat =
     hoistServer faucetAPI nat (toServant faucetApiHandlers)
 
+-- | CORS is enabled to ease development for frontend team.
+ourCors :: Middleware
```
volhovm: Why not `faucetCors` then? Two functions with the same name and different semantics is something my jump-to-definition won't understand :)

Martoon-00: Right, this doesn't make sense

PR #172, opened 2018-08-28
Merge `release-alpha` into `master`

goodsoft:
Now this one should work.


goodsoft:
There were some conflicts:
1. `UnliftIO` in `master` vs `Rio` import fix in `release-alpha` (chose `master`)
2. `wpWalletServerParams → wpWitnessServerParams` vs other params addition, merged
3. `release-alpha` introduced explicit exports in `Snowdrop.Configuration`, which didn't include `blockPrefixes` in `master`.

Other than that it was rather smooth.

goodsoft:
@flyingleafe the less conflicting changes there are, the easier it is to merge branches.
We can easily make another merge PR when all these are merged, but frontend guys will get the stuff they need already now.
Though, ok, I'll wait for the first two to be merged and will update this PR.

PR #178, opened 2018-08-29
[DSCP-???] Hotfixing underflow

Heimdell:


Martoon-00:
Tested, it works (once it builds)

Martoon-00:
Although if account is new, I get `Internal error: account is not registered in chain`

PR #180, opened 2018-08-29
[DSCP-233] No mention of other currencies in Ariadne

goodsoft:


goodsoft:
Oops, broke Ariadne, will fix shortly

PR #187, opened 2018-08-30
[DSCP-234] Implement nextHash field in BE BE

goodsoft:


witness/src/Dscp/Witness/Logic/Processing.hs
```diff
@@ -64,12 +63,16 @@ applyBlockRaw toVerify block = do
...
     (blockCS, stateCS) <- reify blockPrefixes $ \(_ :: Proxy ps) ->
         let actions = SD.constructCompositeActions @ps (SD.dmaAccessActions blockDBM)
-                                                        (SD.dmaAccessActions stateDBM)
+                                                       (SD.dmaAccessActions stateDBM)
             rwComp = do
               sblock <- SD.liftERoComp $ expandBlock block
               -- getCurrentTime requires MonadIO
               let osParams = SD.unOSParamsBuilder sdOSParamsBuilder $ UTCTime (toEnum 0) (toEnum 0)
               SD.applyBlockImpl toVerify osParams blkStateConfig (bBody block) sblock
+              void $ SD.modifyRwCompChgAccum $ SD.CAMChange $ SD.ChangeSet $
```
volhovm: I think this should go to `expandBlock`.

goodsoft: I'm too retarded to fit it there right now, so I'd leave it for later to be moved along with the transactions below.
Will leave a comment though.

volhovm: Yeah but we don't have atomicity this way.

PR #188, opened 2018-08-30
[DSCP-241] Some fixes

Martoon-00:
This does not fix original problem, but still may be useful. 

1. Since `checkThenRepublish` is called in several places, it should be made
atomic.
2. Not always exceptions are caught from `checkThenRepublish`.


witness/src/Dscp/Witness/Workers/Worker.hs
```diff
@@ -148,6 +148,7 @@ makeRelay (RelayState input pipe failedTxs) =
             dieGracefully "tx retranslation repeater" $ forever $ do
                 (_, PubTx tx) <- cliRecvUpdate btq (-1)
                 checkThenRepublish tx
+                    `catchAny` \_ -> pass
```
volhovm: shouldn't you log a warning at least?

Martoon-00:
Fixed. We are going to see many warnings under load though :peka:

PR #189, opened 2018-08-30
[DSCP-239] Fix nonce inconsistency

volhovm:


witness/src/Dscp/Witness/Web/Types.hs
```diff
@@ -87,7 +87,7 @@ instance Buildable a => Buildable (BlocksOrMempool a) where
 instance Buildable (ForResponseLog AccountInfo) where
     build (ForResponseLog AccountInfo{..}) =
         "{ balances = " +| aiBalances |+
-        ", next nonce = " +| aiNextNonce |+
+        ", next nonce = " +| aiCurrentNonce |+
```
Martoon-00: Typo

Martoon-00:
Althoug, as it was sayd, fix `witness.yaml` with swagger doc

PR #196, opened 2018-08-31
LingoHub based on release-alpha

romkaspb:
Pull request created by LingoHub based on branch release-alpha


romkaspb:
Would be reopened again :(

PR #202, opened 2018-08-31
Fix time format

romkaspb:


mkaito:
Already merged into `develop`.

PR #203, opened 2018-08-31
[DSCP-206] Proper block paging in block explorer

goodsoft:


witness/src/Dscp/Witness/Logic/Traversals.hs
```diff
@@ -136,6 +137,32 @@ getBlocksFromTo (headerHash -> olderH) (headerHash -> newerH) = runExceptT $ do
...
+    -> SdM (Either Text (OldestFirst NonEmpty Block))
+getBlocksFrom skip depthDiff = runExceptT $ do
+    tipIdx <- unDifficulty . hDifficulty <$> lift getTipHeader
+
+    newerH <- ExceptT $ maybeToRight "Can't skip so many blocks" <$> SD.queryOne (Difficulty $ tipIdx - skip)
+
+    let loadCond _block depth = depth < depthDiff
+    blocks <- lift $ loadBlocksDownWhile newerH loadCond
+
+    let retrievedOldest = NE.head (unOldestFirst blocks)
```
Martoon-00: `head` and `length` from Universum should work as well

goodsoft: this is mindless copy-pasta

PR #205, opened 2018-09-03
Autoupdate transactions list (each 20 sec.)

romkaspb:


mkaito:
Already merged into `develop`.

PR #211, opened 2018-09-05
[DSCP-242] Throw exception if transaction queue is overflown instead of blocking

flyingleafe:
This avoids the situation when during high transactions load Servant handlers block on `writeTBQueue` and do not return any response until the queue isn't cleared.

### DIFF
--- a/tools/src/txperf/Dscp/TxPerf/Tx.hs
+++ b/tools/src/txperf/Dscp/TxPerf/Tx.hs
@@ -33,7 +33,10 @@ genTx wc async = do
     let toCount = length accounts `min` balance accFrom
     accsTo <- liftIO . generate . resize toCount . listOf1 $ elements $ delete accFrom accounts
 
-    void . liftIO $ sendTx wc async accFrom (map (, Coin 1) . ordNub $ accsTo)
+    liftIO $
+        void (sendTx wc async accFrom (map (, Coin 1) . ordNub $ accsTo))
+        `catch` \(e :: WitnessClientError) ->
+                    putStrLn $ "Tx sending error: " ++ show e
 
     let upd =
             (\acc -> if acc == accFrom
--- a/tools/src/txperf/Dscp/TxPerf/Tx.hs
+++ b/tools/src/txperf/Dscp/TxPerf/Tx.hs
@@ -36,7 +36,7 @@ genTx wc async = do
     liftIO $
         void (sendTx wc async accFrom (map (, Coin 1) . ordNub $ accsTo))
         `catch` \(e :: WitnessClientError) ->
-                    putStrLn $ "Tx sending error: " ++ show e
+                    putStrLn $ "Tx sending error: " <> pretty e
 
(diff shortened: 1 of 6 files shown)


volhovm (approved):

tools/src/txperf/Dscp/TxPerf/Tx.hs
```diff
@@ -33,7 +33,10 @@ genTx wc async = do
     let toCount = length accounts `min` balance accFrom
     accsTo <- liftIO . generate . resize toCount . listOf1 $ elements $ delete accFrom accounts
 
-    void . liftIO $ sendTx wc async accFrom (map (, Coin 1) . ordNub $ accsTo)
+    liftIO $
+        void (sendTx wc async accFrom (map (, Coin 1) . ordNub $ accsTo))
+        `catch` \(e :: WitnessClientError) ->
+                    putStrLn $ "Tx sending error: " ++ show e
```
Martoon-00: Better `pretty e` here.

Martoon-00 (approved):
Reasonable fixes :+1: 
Just see some minor comments.

Martoon-00 (changes requested):
Or wait a sec. I don't think that `TxQueueFull` should be displayed as internal error to the client, there is even special HTTP code for such cases after all.

Please add constructor with `RelayException` into `WitnessAPIError` and modify `fromException` for `WitnessAPIError` accordingly.

witness/src/Dscp/Witness/Web/Error.hs
```diff
@@ -53,22 +56,37 @@ data ErrResponse = ErrResponse
...
+prefixUnavailable :: Text -> Text
+prefixUnavailable = (<>) uaPrefix
+
+unprefixUnavailable :: Text -> Maybe Text
+unprefixUnavailable txt =
+    if T.take l txt == uaPrefix
+    then Just $ T.drop l txt
+    else Nothing
+  where
+    l = length uaPrefix
```
Martoon-00: Yeah, live is hard :pepe_sad:

witness/src/Dscp/Witness/Web/Server.hs
```diff
@@ -34,8 +35,11 @@ convertWitnessHandler
     -> Handler a
 convertWitnessHandler (UnliftIO unliftIO) handler =
     liftIO (unliftIO handler)
-        `catch` (throwError . witnessToServantErr)
-        `catchAny` (throwError . witnessToServantErr . InternalError . show)
+        `catch` throwServant
+        `catch` (throwServant . ServiceUnavailable . show @Text @RelayException)
```
Martoon-00: This way we would need to list all possible exceptions at every `catch` site (at client for instance, if it will be needed).
Better add `ServiceUnavailable <$> fromException e` here https://git.io/fAudK instead

Martoon-00 (approved):
LGTM, thanks for addressing my comments

PR #213, opened 2018-09-06
[DSCP-249] Profile block apply

Martoon-00:
Makes profiling more convenient.
For profiling report see https://issues.serokell.io/issue/DSCP-249.

### DIFF
--- a/tools/singleton-txperf.json
+++ b/tools/singleton-txperf.json
@@ -1,0 +1,10 @@
+{
+  "witness": "localhost:8091",
+  "accKeys": [
+    "ffRB5uhVEV24NXsYUuw1irt9HF5i5IRWJXHJzagXU8Y=",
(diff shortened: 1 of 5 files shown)


tools/singleton-txperf.json
```diff
@@ -0,0 +1,10 @@
+{
+  "witness": "localhost:8091",
+  "accKeys": [
+    "ffRB5uhVEV24NXsYUuw1irt9HF5i5IRWJXHJzagXU8Y=",
+    "fOhZOFr6SSnYv1/xfKPPMHdWLdQDboysJ4Q794Gezmw="
+  ],
+  "genAccs": 2,
+  "txCount": 100,
```
flyingleafe: 100 transactions is not enough for loading witness significantly, but this parameter is going to be changed during load testing anyway, so I don't object.

PR #210, opened 2018-09-05
Remove frontend code from repo

flyingleafe:
Code is moved to these repos:
https://github.com/DisciplinaOU/disciplina-faucet-frontend
https://github.com/DisciplinaOU/disciplina-explorer-frontend

PR #215, opened 2018-09-06
nix: remove frontends from derivation

mkaito:


yorickvP:
Also, what's #210 ?

mkaito:
I didn't even see #210, but assuming we delete the same things, there should be no conflict.

PR #217, opened 2018-09-07
[DSCP-243] Remove witness backend ping on faucet startup

Martoon-00:
In contrast to original purpose this just delivered plenty of inconveniences to
devops team.

### DIFF
--- a/faucet/exec/Main.hs
+++ b/faucet/exec/Main.hs
@@ -2,10 +2,9 @@
 
 module Main where
 
-import Control.Lens (views)
 import Fmt ((+|), (|+))
 import Loot.Base.HasLens (lensOf)
-import Loot.Log (logError, logInfo)
+import Loot.Log (logInfo)
 import Options.Applicative (execParser, fullDesc, helper, info, progDesc)
 
 import Dscp.CommonCLI (versionOption)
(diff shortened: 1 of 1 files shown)


faucet/exec/Main.hs
```diff
@@ -2,42 +2,29 @@
...
   where
-    printSourceInfo params = do
-        wc <- views (lensOf @WitnessClient) (hoistWitnessClient liftIO)
+    printSourceInfo = do
         pk <- view $ lensOf @(KeyResources FaucetApp) . krPublicKey
         let addr = mkAddr pk
-
         logInfo $ "Faucet source address: " +| addr |+ ""
 
-        let DryRun dryRun = _fpDryRun params
```
goodsoft: shouldn't you nuke `_fpDryRun` now too?

Martoon-00: It was used primarily to test frontend making transactions without witness backend, not only to suppress this check at start.
Taking this into account, does it worth removing this flag anyway?

goodsoft: If it's used somewhere else, then probably it's not.
Otherwise, if it's not used anywhere, then there's no reason to keep it around at all.

Martoon-00: Ok, there is one usage site in tx making logic, thus I remain.

PR #218, opened 2018-09-07
Add wallet derivations

lukateras:


yorickvP:
error: anonymous function at /nix/store/xbjk1j3kaphydwqhnv96j81q2c6ma683-source/default.nix:2:1 called without required argument 'target', at /var/lib/buildkite-agent/builds/mars-1/serokell/disciplina/default.nix:75:33

yegortimoshenko:
@yorickvP Should I set `meta.platforms = platforms.darwin;` on macOS bundle? It won't build on Linux. How do we handle that?

yegortimoshenko:
@yorickvP Moved wallet derivations to `release.nix`. Also, this branch uses new fetcher.

educator/src/Dscp/Educator/Web/Educator/API.hs
```diff
@@ -40,7 +40,7 @@ protectedEducatorAPI = Proxy
 
 data EducatorApiEndpoints route = EducatorApiEndpoints
     {
-      -- * Students
```
yegortimoshenko: Those commits are invalid. There are much more of these, only cleaned up several files before switching to `doHaddock = false;`.

default.nix
```diff
@@ -1,58 +1,44 @@
...
+        buildTools = (super.buildTools or []) ++ [ autoexporter tasty-discover ];
+        testDepends = (super.testDepends or []) ++ [ hspec tasty tasty-discover tasty-hspec ];
+
+        doCheck = true;
+        doHaddock = false; # until syntax errors in comments are fixed
+
+        postInstall = ''
+          mkdir -p $out/share/disciplina/specs $out/etc/disciplina
+          cp ${src}/configuration.yaml $_
+          find ${src}/specs -name '*.yaml' -exec cp '{}' $out/share/disciplina/specs \;
```
yegortimoshenko: Do we need any of this?

yorickvP: I was using it in the CD profiles, since they kept changing the config. But it makes disciplina nix expressions depend on ./. and rebuild every time anything changes, so it's actually terrible. Any alternative besides hardcoding config in CD profiles?

yegortimoshenko: I think have a separate derivation (`disciplina-config`) that handles this.

yorickvP:
I will need disciplina-bin back :/

yegortimoshenko:
@yorickvP Why? You said it's just to save some compilation time. I've pushed all artifacts to https://serokell.cachix.org.

yegortimoshenko:
If you want to push smaller closures to NixOps, let's just add static build to all local packages.

yorickvP:
Yes, I want smaller closures but not twice the compilation time. Can't make them all justStaticExecutables, because they do link against eachother, so the libs are needed. The proper solution here is to have multiple outputs.

default.nix
yorickvP: Please do that :)

educator/src/Dscp/Educator/Web/Educator/API.hs
yorickvP: The comment changes should be in another PR

release.nix
```diff
@@ -0,0 +1,50 @@
+{ pkgs ? import ./nixpkgs.nix }: with pkgs;
+
+let
+  buildFlatpak = callPackage (fetchGit {
+    url = "https://github.com/serokell/nix-flatpak";
+    rev = "46a2aadf37981d6313621913cd2802debfb763fd";
```
yorickvP: I'm not sure if I'm happy with abolishing the serokell-overlay fetch, changing to 3 different fetches. Maybe These should be in serokell-public-overlay if that still exists.

wallet/profile.sb
```diff
@@ -6,7 +6,7 @@
   (literal "/dev/urandom")
   (subpath "/usr/lib")
   (subpath "/usr/share")
-  (subpath (param "root")))
+  (subpath (param "DYLD_ROOT_PATH")))
```
yorickvP: What's the sb file?

yegortimoshenko: It's a macOS sandbox policy in Scheme.

release.nix
yegortimoshenko: It still exists. Guess that would work.

PR #219, opened 2018-09-07
[DSCP-248] General fixes

Martoon-00:
Some small fixes I dunno how I came to doing this, but this may be helpful.
See on per commit basis, their descriptions are telling.


Martoon-00:
Lol, someone is going to have less fun :man_shrugging:

PR #221, opened 2018-09-11
[DSCP-245] Keep several SQLite connections used exclusively.

Martoon-00:
- [x] Implement SQLite context
- [x] Update the codebase
- [x] Test

Timeouts are left for separate PR


Martoon-00:
Nice, I'll proceed to updating remaining codebase then.

Martoon-00:
This is ready for review.

educator/tests/Test/Dscp/DB/SQLite/Transactions.hs
```diff
@@ -0,0 +1,96 @@
...
+    db' <- openSQLiteDB dbParams
+    runRIO db' $ invoke prepareSchema
+    closeSQLiteDB db'
+    bracket (openSQLiteDB dbParams)
+            (\db -> closeSQLiteDB db >> removeFile dbPath) $
+            \db -> runRIO db action
+  where
+    -- sad, but testing with in-memory database is not an option
+    -- because each connection would work with its own database in this case
+    dbPath = "./sql-transaction-test.db"
```
flyingleafe: This is so sad. Did you try to use single connection with WAL? Any chance it will work properly?

educator/src/Dscp/DB/SQLite/Functions.hs
```diff
@@ -4,76 +4,273 @@ module Dscp.DB.SQLite.Functions
...
+-- for details.
+data OperationType = Writing | Reading
+
+-- | Make an SQL query which returns some result.
+--
+-- Note: performing WRITING operations is PROHIBITED within this function,
+-- use 'modifyingQuery' for such purpose.
+query
+    :: (MonadIO m, FromRow row, ToRow params)
+    => Query -> params -> DBT t w m [row]
```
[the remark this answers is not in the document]
Martoon-00: Funny that `query` will execute only one statement, and if I pass `SELECT ... ; INSERT ...`, only the first operation will be executed.
So we can add a simple runtime check which allows only commands starting from a word in the given set (probably just `[insert]`).

educator/tests/Test/Dscp/DB/SQLite/Transactions.hs
Martoon-00: Right, let me try

Martoon-00: No, it doesn't :pepe_sad:

PR #227, opened 2018-09-14
[DSCP-260] Replace blockHeaderHash with blockHeader in publication

volhovm:
Previously, both PublicationTx and Publication had private block hashes inside. It is a design mistake and actually what should be there is a private block header itself.


core/src/Dscp/Core/Foundation/Witness.hs
```diff
@@ -173,14 +173,15 @@ instance Buildable TxWitnessed where
 
 -- | Transaction for private block publications.
 data PublicationTx = PublicationTx
-    { ptAuthor    :: Address
-    , ptPrevBlock :: Maybe PrivateHeaderHash
-    , ptBlock     :: PrivateHeaderHash
+    { ptAuthor     :: Address
+    , ptHeader     :: PrivateBlockHeader
+    , ptPrevHeader :: Maybe PrivateBlockHeader
```
Heimdell: What? Why not hash in `ptPrevHeader`? We already have that entity in our world state, why duplicate?

core/src/Dscp/Core/Foundation/Witness.hs
```diff
@@ -190,10 +191,9 @@ toPtxId = hash
 
 -- | Incoming publication message; transaction payload.
 data Publication = Publication
-    { pPrivateBlockHash  :: PrivateHeaderHash
-    , pPreviousBlockHash :: Maybe PrivateHeaderHash
-    }
-    deriving (Eq, Ord, Show, Generic)
+    { pPrivateHeader  :: PrivateBlockHeader
+    , pPreviousHeader :: Maybe PrivateBlockHeader
```
Heimdell: Same question. Why do we need to send previous block header, if we can send hash of previous block header?

core/src/Dscp/Core/Foundation/Witness.hs
volhovm: Oh yeah my god I'm kind of stupid. Give me a sec.

core/src/Dscp/Core/Foundation/Witness.hs
flyingleafe: Also, hash of previous block header is already contained in `PrivateBlockHeader` :man_shrugging:

PR #229, opened 2018-09-14
[DSCP-228] Make it possible to specify all node parameters in config file

flyingleafe:
What should be done:

The following executables should be able to launch with no CLI params but path to config file (and config key):

- [x] Witness node
- [x] Educator node
- [x] Faucet

For following executables overriding of separate config options via CLI params is possible:
- [x] Witness node
- [x] Educator node
- [x] Faucet

Also:
- [x] Config file structure in sane and well-documented
- [x] All launch scripts work in the same manner they did before


core/package.yaml
```diff
@@ -52,6 +52,7 @@ library:
     - time
     - unix
     - unliftio
+    - vinyl
```
Martoon-00: Oh finally, now we can use `++` from vynil instead of self-written `+++`

core/src/Dscp/Config/Util.hs
```diff
@@ -105,6 +110,15 @@ buildConfig ConfigParams{..} filler =
             finalise fileConfigFilled
         pure config
 
+-- | Utility function for filling up a config reusing existing function
+-- for a subconfig.
```
Martoon-00: I cannot parse this sentence :pape_sad:

core/src/Dscp/Util/Aeson.hs
```diff
@@ -110,3 +111,9 @@ instance ToJSON SemVer.Version where
     toJSON = String . SemVer.toText
 instance FromJSON SemVer.Version where
     parseJSON = withText "version" $ leftToFail . SemVer.fromText
+
+instance KnownRatName unit => ToJSON (Time unit) where
+    toJSON = String . toText . unitsF
+instance KnownRatName unit => FromJSON (Time unit) where
+    parseJSON = withText "time duration" $
+        nothingToFail ("Invalid time format" :: String) . unitsP . toString
```
Martoon-00: Cool, didn't know about `unitsF` and `unitsP`

educator/src/Dscp/Educator/Config.hs
```diff
@@ -1,30 +1,48 @@
...
 
-
-type EducatorConfig = CoreConfig
+type EducatorConfig = WitnessConfig +++
+    '[ "educator" ::<
+       '[ "db" ::: SQLiteParams
+        , "keys" ::: EducatorKeyParams
+        , "api" ::: EducatorWebParams
+        ]
+     ]
```
Martoon-00: Total configuration occurs to be ambiguous as soon as some parameters are present both in `witnessConfig` and `educatorConfig`.
Could we make `EducatorConfig` contain no fields of `WitnessConfig`; then parse `WitnessConfig ++ EducatorConfig` and `rcast` it appropriately when passing to `withWitnessConfig` and `withEducatorConfig`?

Martoon-00 (commented):
LGTM, although I wonder how config would look like now.

educator/src/Dscp/Educator/Config.hs
flyingleafe: @Martoon-00 Well, I think you're wrong, because currently configuration doesn't contain any duplicated data (similar `api` and `keys` fields in `witness` and `educator` subsections are about different keys and different APIs). Though your suggestion is legit anyway; I even wanted to separate config further, having parts like `LoggingConfig`, `NetworkConfig`, etc.

lukateras (approved):

PR #233, opened 2018-09-18
[DSCP-263] Add helper for testing with curl

Martoon-00:
I think those should be easily available for everyone.


Martoon-00:
DSCP-263 seem to be resolved at our side, but to check that I had to make `curl` request to backend (which by the way is accepted successfully by `makeSubmission` endpoint).
And generating authentication header manually is tiresome. Anyone who needs to test server without involving frontend will find this function useful while we have our current authentication.

Martoon-00:
Added doc entry.

PR #234, opened 2018-09-19
[DSCP-252] Generalised fee policy

Martoon-00:
Allow not only linear policies.
Since we have several kinds of transactions, I used GADTs to make it possible to apply some policies only to specific kind of transactions, maybe I overengeneered.

todo:
- [x] Wait for #225 to be merged (to avoid conflicts)
- [x] Wait for #229 to be merged (to avoid conflicts)
- [ ] Update configuration file

### DIFF
--- a/config-full-sample.yaml
+++ b/config-full-sample.yaml
@@ -6,19 +6,27 @@ sample:
     # in a slot
     slotDuration: 10000
     
-    # Transaction fee parameters.
+    # Transaction fee parameters
     fee:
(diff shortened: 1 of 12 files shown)


Martoon-00:
I wanted to but it will produce conflicts with your work on config.
Will do tomorrow's noon anyway.

Martoon-00:
Wow, so cool that all those PR's are just merged finally, I was expecting to waste a day on it.

PR #237, opened 2018-09-19
[DSCP-253] Show estimated tx fee in UI

goodsoft:


wallet/src/Dscp/Wallet/Backend.hs
```diff
@@ -22,12 +22,14 @@ createWalletFace ::
 createWalletFace serverAddress sendEvent = buildComponent_ "Wallet" $ do
     sendStateUpdateEvent sendEvent
     wc <- createWitnessClient serverAddress
+    (dummySK, _) <- keyGen
```
Martoon-00: I can imagine the day when estimated fees would vary across restarts depending on the moon phase.
(I do not insist, but we have `detGen` for deterministic generation)

wallet/src/Dscp/Wallet/Backend.hs
```diff
@@ -96,6 +98,27 @@ sendTx wc sendEvent eSecretKey mPassPhrase (toList -> outs) = do
...
+    gTx = GMoneyTxWitnessed . fixFees feeCoefficients $ \fees ->
+        let inAcc   = TxInAcc { tiaNonce = nonce, tiaAddr = address }
+            inValue = Coin (sum $ map (unCoin . txOutValue) outs) `unsafeAddCoin` unFees fees
+            tx      = Tx { txInAcc = inAcc, txInValue = inValue, txOuts = outs }
+
+            signature   = sign dummySK (toTxId tx, publicKey, ())
+            witness     = TxWitness   { txwSig = signature, txwPk = publicKey }
+            txWitnessed = TxWitnessed { twTx   = tx, twWitness = witness }
+        in txWitnessed
+
```
Martoon-00: I think it should be in `Dscp.Core.Fees` module

flyingleafe: And actually this should not require a secret key. All witnesses have the same size, we can just have hardcoded "sample" witness for this purpose.
I disagree with @Martoon-00's "moon phase" argument; I claim that fee size will not depend on moon phase (or the identity of a sender) in the nearest future, I see no possible reason for it to depend on it and I doubt that we should have suboptimal code "just in case". If we'll find it necessary, we'll introduce such a dependency separately anyway.

goodsoft: What is the preferred way of _hardcoding witness_? Coercing some hardcoded stuff to `Signature` and `PublicKey`?

flyingleafe:
@goodsoft bump, whatsup with that?

goodsoft:
Waiting for Ariadne part to get reviewed and merged, hopefully that will happen today.

core/src/Dscp/Core/Fees.hs
```diff
@@ -102,3 +103,20 @@ fixLinearFees coeffs buildTx = go (Fees $ fcMinimal coeffs)
...
 fixFees feePolicy = case feePolicy of
     LinearFeePolicy coeffs -> fixLinearFees coeffs
+
+-- | Estimate fees for given outputs.
+-- Only works for fixed nonce, key and signature sizes.
+estimateFees :: FeeConfig -> [TxOut] -> Fees
+estimateFees feeConfig outs = calcFeeG feeConfig gTx
+  where
+    publicKey = leftToPanic $ fromByteArray ("patakbardaq_skovoroda_pvaforever" :: ByteString)
+    signature = leftToPanic $ fromByteArray ("patakbardaq_skovoroda_pvaforever_and_thirty_two_more_bytes_mkay?" :: ByteString)
```
Martoon-00: For both you can use `detGen 42 arbitrary` if you care about size which may change.

Martoon-00: well yes, it doesn't differ conceptually

PR #238, opened 2018-09-19
[DSCP-271] Add description: how to contruct witness for student API

Martoon-00:


docs/api-types.md
```diff
@@ -0,0 +1,100 @@
...
+
+Let's assume we have following data (all entities are given in hex for convenience, algorithm itself operates with raw bytes):
+
+```javascript
+var studentSecretKey = "6dbb1eeeb3e8e81547f307ba9344dae6286a9319458e9596f2ffc43670d5c648"
+var contentsHash = "01a7cc7de66f6e398ccc5bfdeb2db81646aed6e43f90f7333d9a06ba205b7f74"
+var assignmentHash = "219bfb1e71af4f888fce709900862b4467f45b42b0603a56e9428cb3e8eb8940"
+```
+
+Then algorithm proceedes as follows:
```
flyingleafe: `proceeds`

docs/api-types.md
```diff
@@ -0,0 +1,100 @@
+# API types documentation
+
+This document describes format of objects used in various API's across the project.
```
flyingleafe: why do you need a tick in `API's`?

docs/api-types.md
```diff
@@ -0,0 +1,100 @@
...
+For other details, see swagger documentation of [educator](/specs/disciplina/educator/api/educator.yaml), [student](/specs/disciplina/educator/api/student.yaml) and [witness](/specs/disciplina/witness/api/witness.yaml) API.
+
+## Common objects
+
+Some types of values are always represented in the same format, among them:
```
flyingleafe: `s/among them/for example/` maybe?

docs/api-types.md
flyingleafe: I have a brilliant idea: let's introduce a requirement that documentation should be passed through grammarly.com before submitting :thinking:

Martoon-00: That sounds like a really good idea. Although I'm not sure how can we submit a document on check quickly.

docs/api-types.md
Martoon-00: Dunno, it seems more readable than `APIs`.
I guess you don't think so, I'll remove.

docs/api-types.md
Martoon-00: I meant to say that I enlist all such types. Maybe `they are`?

flyingleafe: `Among them` really says quite the opposite. You should say `The following types are always represented in the same format` then.

Martoon-00:
@flyingleafe if you are in general satisfied with the format of this document, I would ask @romkaspb to review javascript code :slightly_smiling_face:

PR #240, opened 2018-09-20
[DSCP-263] Make address encoding correspond CBOR spec

Martoon-00:
Let compatibility break one more time!
Tests pass locally.


Martoon-00:
We need to regenerate all address in config, which is sad.
I prefer to rework our serialisation ASAP and include these changes there to break compatibility just one more time

PR #241, opened 2018-09-20
Re-add GHC to shell.nix

yegortimoshenko:
I've wiped .stack-work clean and currently rebuilding it back inside `nix-shell`.

### DIFF
--- a/shell.nix
+++ b/shell.nix
@@ -1,20 +1,28 @@
 { pkgs ? import ./closure.nix }: with pkgs;
 
-stdenv.mkDerivation {
+stdenv.mkDerivation rec {
   name = "disciplina";
 
   nativeBuildInputs = [
+    binutils
+    git
+    haskell.compiler.ghc822
     haskellPackages.tw
     hlint
+    nix
+    pkgconfig
+    stack
   ];
 
   buildInputs = [
-    binutils
-    git
     gmp
     openssl
     rocksdb
     zeromq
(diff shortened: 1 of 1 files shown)


flyingleafe:
No, not immediately. For some reason, at first it seemingly installed `nix-2.1.1` (I guess because my system nix version differs from version in `serokell-closure`), and then failed with aforementioned error when trying to initialize ghc package database. When I tried to run `stack build` again, it actually started building (from scracth, all 272 packages).

flyingleafe:
@yegortimoshenko my system `nix` is 2.1.1 too, though, and https://github.com/serokell/nixpkgs seems pretty fresh, so I have no idea why it started to download and build `nix` itself.

yegortimoshenko:
@flyingleafe The reason is that our Nix contains patches that enable some `fetchGit` operations that mainline Nix would normally not support at the moment. That's expected.

I wonder if `mv .stack-work .stack-work.bak; nix-shell --run 'stack build'` would work.

flyingleafe:
@yegortimoshenko My build failed on 171/272 complaining that `pkg-config` cannot be found, apparently, `pkgconfig` should be added to `nativeBuildInputs`

yegortimoshenko:
@flyingleafe I've completed the build, and pushed updated `shell.nix` with `pkgconfig` and `zlib` flag.

PR #243, opened 2018-09-21
Push Disciplina with runtime deps to Cachix

lukateras:
One can use cache by running:

```
nix-env -f '<nixpkgs>' -iA cachix
sudo cachix use disciplina
````

### DIFF
--- a/secrets/clusterCi/faucet.key
+++ b/secrets/clusterCi/faucet.key
@@ -1,1 +1,0 @@
-{"content":{"secret":"UIyfCtFDZrJOWikvglSImTNQSRARGSzGuLbe2HvCMIf7G1ggHmQmmaQu4yt9IQk+k7cBRO9G4xMHNfIZs/aBEmtHE/Y="},"version":"1.0.0"}

### CODE REVIEW

File: default.nix [hunk 1]
```diff
@@ -27,11 +27,12 @@ let
         "scripts/.*"
         "secrets/.*"
         "specs/.*"
+        ".*\\.nix"
```
yorickvP: This should happen automatically, since these are not whitelisted

release.nix
```diff
@@ -15,6 +15,20 @@ let
 in
 
 rec {
+  disciplina-haddock = with lib;
+    let
+      drvs = filterAttrs (const (hasAttr "doc")) project;
+      docs = mapAttrs (const (getOutput "doc")) drvs;
+      globs = map (doc: "${doc}/share/doc/*") (attrValues docs);
+    in
+    runCommand "disciplina-haddock" {} ''
```
yorickvP: [pkgs.linkFarm](https://github.com/NixOS/nixpkgs/blob/master/pkgs/build-support/trivial-builders.nix#L146) "disciplina-haddock" ?

shell.nix
```diff
@@ -5,6 +5,7 @@ stdenv.mkDerivation rec {
 
   nativeBuildInputs = [
     binutils
+    cachix
```
yorickvP: should this be a build input for `stack --nix`?

yegortimoshenko: See DisciplinaOU/disciplina-nixops@627e9ba7981b94e9b8bc63d8e96bda93be1fbc01.

release.nix
yegortimoshenko: Current approach looks much better in Buildkite artifacts output.

default.nix
yegortimoshenko: I'll check, but I think I added it because I had rebuilds on *.nix changes.

yegortimoshenko: Nah, I removed it, immediately caused a rebuild.

PR #244, opened 2018-09-23
[DSCP-263] Fix educator secret

Martoon-00:
1. Keep educator secret key in `run/educator.key` rather than `run/tmp/educator.key` which is deleted upon each launch
2. Keep unencrypted secret in key file (password is long forgotten)
3. Update authentication doc to tell how to deal with Educator API's authentication


Martoon-00:
Sounds really good. Let's remain this branch as a temporary workaround for QA and frontend team, I'll add that executable in `tools`. This will especially facilitate writing scripts on python for @ArgentumHeart, as he won't need to hardcode tokens generated in ghci anymore.

PR #250, opened 2018-09-24
[DSCP-271] Complete API doc

Martoon-00:
The format of `Address` serialization changed, need to catch up.
Result: https://github.com/DisciplinaOU/disciplina/blob/martoon%2Fdscp271-complete-api-doc/docs/api-types.md#L22

Also, I'm starting to promote `httpie` as it was discussed with John, since it is easy to install, requires less boilerplate and provides a much more detailed output on what's going on.

### DIFF
--- a/docs/api-types.md
+++ b/docs/api-types.md
@@ -1,7 +1,7 @@
 # API types documentation
 
-This document describes format of objects used in various APIs across the project.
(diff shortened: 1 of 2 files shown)


Martoon-00:
Thanks. I'll wait for the review of @romkaspb, if he does not respond I'll merge tomorrow

PR #251, opened 2018-09-24
[DSCP-275] [DSCP-282] Make Educator and Student APIs authentication optional

Martoon-00:
In order to avoid keeping two API types - with and without authentication - I pass that option via context along with educator secret key.

- [x] Clean up commit history

### DIFF
--- a/.stylish-haskell.yaml
+++ b/.stylish-haskell.yaml
@@ -225,6 +225,7 @@ language_extensions:
   - DefaultSignatures
   - DeriveDataTypeable
   - DeriveGeneric
+  - ExplicitNamespaces
   - FlexibleContexts
   - FlexibleInstances
   - FunctionalDependencies
--- a/config-full-sample.yaml
+++ b/config-full-sample.yaml
@@ -5,7 +5,7 @@ sample:
     # Slot duration in milliseconds. A block is issued once
     # in a slot
     slotDuration: 10000
-    
+
     # Transaction fee parameters
     fee:
       # Fee parameters for money transactions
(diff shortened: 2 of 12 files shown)


hithroc:
After very quickly glancing through the PR, why not just use raw `EducatorAPI` instead of `ProtectedEducatorAPI` based on if authentication disabled or not?

I'll look into PR more in detail a little bit later.

Martoon-00:
As soon as `EducatorAPI` and `ProtectedEducatorAPI` are different _types_, this would require me writing a huge `case` over [this](https://github.com/DisciplinaOU/disciplina/blob/master/educator/src/Dscp/Educator/Web/Server.hs#L114-L119) and [that](https://github.com/DisciplinaOU/disciplina/blob/master/educator/src/Dscp/Educator/Web/Server.hs#L46-L56) and duplicating those lines.

hithroc:
Okay, yes, I see the problem here. Having authentication in the type and just replacing it with a stub just doesn't feel right, but I'll admit that I have no objective reasoning here. I'll think about it on my way home, and when I get home I'll review the PR more thoroughly.

Martoon-00:
There is one another way I was thinking about: servant allows to specify several authentication methods and _any_ of them passing is sufficient, we could add a second authentication method which always passes if given flag is `True`.
Shortcoming if it is that it only allows to disable all authentication methods at once, but probably that's exactly we want. And it's still term-level approach, not type-level.

Martoon-00:
Ok, here goes another try.

Martoon-00:
@hithroc If you will manage to come up with a better decision, ping me to discuss. For now, I merge.

Martoon-00:
:+1:

PR #252, opened 2018-09-24
[DSCP-255] Produce `instance HasLens` boilerplate with TH

Martoon-00:
Add functions to automatically traverse fields of a datatype and add `HasLens` instances.

Probably I will move all TH code to lootbox once you tell whether the approach is sensible.

### DIFF
--- a/core/src/Dscp/Util/HasLens.hs
+++ b/core/src/Dscp/Util/HasLens.hs
@@ -1,0 +1,118 @@
+-- | Utilities to work with 'HasLens'.
+
+module Dscp.Util.HasLens
+    ( deriveHasLens
+    , deriveHasLensPath
+    , deriveHasLensPathWith
+    , deriveHasLensDirect
+    , deriveHasLensDirectWith
+    ) where
+
+import Control.Lens (DefName (TopName), LensRules, lensField, lensRules, to, traversed)
+import qualified Data.List as L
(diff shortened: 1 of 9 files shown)


core/src/Dscp/Util/HasLens.hs
```diff
@@ -0,0 +1,117 @@
...
+deriveHasLensExt deriveWay preLensQ ctxName tyName = do
+    ty <- TH.reify tyName
+    decs <- case ty of
+        TH.TyConI (TH.DataD _ _ _ _ [TH.RecC _ fields] _) -> do
+            let fieldNames = fields ^.. traversed . _1 . to show . to toUnqualifiedName
+            forM fields $ \(quaFieldName, _, fieldTy) -> do
+                let fieldName = toUnqualifiedName (show quaFieldName)
+                let mkLensName rules = toLensName rules fieldNames fieldName
+                mkInstance fieldTy mkLensName
+        _ -> fail "Expected name of a datatype with exactly one constructor"
```
flyingleafe: This error will also be shown on attempt to apply the function to the datatype which has exactly one constructor, but is not a record (as long as you match only `RecC`). I agree that support for `NormalC` would be cumbersome, but you might make better error message.

Martoon-00: Good point.

PR #255, opened 2018-09-25
[DSCP-274] Keygen

Martoon-00:
See [readme](https://github.com/DisciplinaOU/disciplina/blob/martoon%2Fdscp274-keygen/tools/README.md)

This is generally can be used for any imaginary operations with a secret key.


core/src/Dscp/Util.hs
```diff
@@ -172,6 +174,9 @@ nothingToFail e = maybe (fail $ toString e) pure
 nothingToPanic :: Text -> Maybe a -> a
 nothingToPanic e = fromMaybe (error e)
 
+maybeToEither :: e -> Maybe a -> Either e a
+maybeToEither e = maybe (Left e) Right
```
flyingleafe: This is `maybeToRight` from Universum.

tools/README.md
```diff
@@ -32,3 +40,93 @@ Contents of the config file must be in the following format:
...
+
+## keygen
+
+This is a small tool to manipulate with secret key. Along with generating new one, it allows to produce various deriviative data.
+
+### Usage
+
+Main usage pattern is
+```bash
+echo "<commands>" | dscp-keygen <base-secret-option>
```
flyingleafe: This is nice, but why not use `-c|--command` option instead of reading from stdin? It would be much clearer, and those who would like to supply command from input stream for some reason (e. g. redirecting from some script) could use `xargs` anyway.

If anything is worth reading from stdin, it's the _input_ of the command - which could be a seed, secret or a keyfile. Then the usage of the keygen would be something very similar to `sed`, like:
```
cat ./secret.key | dscp-keygen --keyfile --command "address"   # display the address
```
or 
```
echo "abcde123" | dscp-keygen --seed --command "keyfile:qwerty123:pretty"
```
Even `--random` option might be omitted, because you can just do
```
take -c 10 /dev/urandom | dscp-keygen --seed --command "secret:base64"
```
I think such approach is more straightforward and Unix-like.

tools/src/keygen/KeygenOptions.hs
```diff
@@ -0,0 +1,99 @@
...
+          help "Use given encrypted secret key as base."
+      pp <- optional . option passphraseReadM $
+          long "password" <>
+          metavar "TEXT" <>
+          help "Password from given key. If no password is specified, secret \
+               \key is assumed to have no password."
+      return $ runRIO (noLogging @IO) $ do
+          KeyResources{..} <- readStore file (pp ?: emptyPassPhrase)
+          return _krSecretKey
+    ]
```
flyingleafe: This is awful, copypaste and formatting is incomprehensible, also doing some actual logic _right inside options parser_ is something I cannot really justify unless you're on the programming competition and don't have extra 10 minutes to come up with some `SecretInputType` ADT. Please make it decent.

tools/README.md
Martoon-00: Right, this makes perfect sense.

Martoon-00:
I hope I address all comments, ready for rereview.

PR #262, opened 2018-09-26
[DSCP-284][DSCP-203] Include public key into JWS authentication header

flyingleafe:
### Description

This PR changes the authentication scheme of Student API in the following way:
- JOSE header of JWS passed in the `Authorization` header now must include public JWK in `jwk` field (https://tools.ietf.org/html/rfc7515#section-4.1.3)
- This allows to determine the student which made the query without traversing some hardcoded set of allowed keys in `O(n)`.
- Which allows for actually checking if the student exists in the database (if the bot is disabled and we don't want to allow everybody to use our API).

PR is WIP, because docs are not updated and not all desired tests are implemented.

### YT issue

https://issues.serokell.io/DSCP-284
https://issues.serokell.io/DSCP-203

### Introduced changes

- [x] New feature
- [ ] Bugfix 
- [x] Refactoring
- [ ] New tests (on functionality not fixed/introduced in this PR)
- [ ] New docs (on functionality not fixed/introduced in this PR)
- [ ] Infrastructure changes 

### Checklist

If I added new functionality, I
- [x] added tests covering it
- [x] explained it using haddock in the source code

If I fixed a bug, I
- [ ] added a regression test to prevent the bug from silently reappearing again

If I changed public API structure or/and data serialization, I made sure that
- [ ] those changes are reflected in [Swagger API specs](/specs/disciplina)
- [ ] and also in [related](/docs/api-types.md) [documentation](/docs/authentication.md).

If I changed config structure or/and CLI interface of any executable, I made sure that
- [ ] [sample config](/docs/config-full-sample.yaml) is updated accordingly
- [ ] [launch scripts](/scripts/launch) are updated accordingly and work as expected

If I changed the procedure of building a project and/or running any executable or helper script, I
- [ ] updated [README.md](/README.md) accordingly
- [ ] as well as subproject READMEs for all related executables

Also,
- [x] my commit history is clean, descriptive and do not contain merge or revert commits or commits like "WIP"
- [x] my code complies with the [style guide](/docs/code-style.md)
- [x] my documentation is checked with a spell checker (like [Grammarly](https://app.grammarly.com))

### DIFF
--- a/core/src/Dscp/Crypto/Util.hs
+++ b/core/src/Dscp/Crypto/Util.hs
@@ -1,0 +1,18 @@
+-- | Small utility functions
+
+module Dscp.Crypto.Util
+       ( constTimeEq
+       ) where
+
+import qualified Crypto.Util as CU (constTimeEq)
+import Data.ByteArray (ByteArrayAccess, convert)
+
+-- | Checks two bytestrings for equality without breaches for
+-- timing attacks.
+--
+-- Adapts 'Crypto.Util.constTimeEq' for any 'ByteArrayAccess' instance.
+-- See <http://codahale.com/a-lesson-in-timing-attacks/> for more info.
+constTimeEq
+    :: (ByteArrayAccess b1, ByteArrayAccess b2)
+    => b1 -> b2 -> Bool
+constTimeEq b1 b2 = CU.constTimeEq (convert b1) (convert b2)
--- a/educator/src/Dscp/Educator/Web/Auth.hs
+++ b/educator/src/Dscp/Educator/Web/Auth.hs
@@ -1,10 +1,12 @@
 {-# LANGUAGE TypeInType #-}
 
 -- | Necessary types and implementation for web server authenthication
+-- TODO: move all authentication stuff not directly related to Educator
+-- to a separate module.
 module Dscp.Educator.Web.Auth
        ( Auth'
        , AuthData (..)
-       , checkAuthData
+       , checkAuthBasic
 
        , NoAuth
        , NoAuthData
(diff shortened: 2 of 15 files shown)


core/src/Dscp/Crypto/Util.hs
```diff
@@ -0,0 +1,18 @@
...
+
+-- | Checks two bytestrings for equality without breaches for
+-- timing attacks.
+--
+-- Adapts 'Crypto.Util.constTimeEq' for any 'ByteArrayAccess' instance.
+-- See <http://codahale.com/a-lesson-in-timing-attacks/> for more info.
+constTimeEq
+    :: (ByteArrayAccess b1, ByteArrayAccess b2)
+    => b1 -> b2 -> Bool
+constTimeEq b1 b2 = CU.constTimeEq (convert b1) (convert b2)
```
Martoon-00: Lool, I wonder how many places are missing this function? For instance, our hashes are compared on a simple `==` inside.

educator/src/Dscp/Educator/Web/Auth.hs
```diff
@@ -69,21 +74,50 @@ instance ( HasServer api ctxs
...
 ---------------------------------------------------------------------------
 
+-- | If request has header "Authorization: Bearer <token>", get the
+-- "<token>" part.
+authBearerToken :: Request -> Maybe ByteString
+authBearerToken request = do
+    hdr <- L.lookup "Authorization" $ requestHeaders request
+    let bearer = "Bearer "
+        (bearer', rest) = BS.splitAt (BS.length bearer) hdr
+    guard (bearer == bearer')
```
Martoon-00: This is `rest <- BS.stripPrefix "Bearer " hdr`

core/src/Dscp/Crypto/Util.hs
flyingleafe: It's not that we need to use this everywhere - using it in inner logic of block validation, for example, makes no sense. We might just define `Eq` instances for all data which might be sensitive using this function, but then we might experience some detectable performance overhead without a good reason. Unfortunately, we need to _think_ when we compare some data with some another in order to decide whether or not to use this function.

educator/src/Dscp/Educator/Web/Server.hs
```diff
@@ -68,19 +69,24 @@ mkStudentApiServer nat botSwitch = do
...
-    return . GetStudentsAction $ atomically . readTVar $ tvr
+-- | Create an action which checks whether or not the
+-- student is a valid API user.
+-- If bot is enabled, all students are allowed to use API.
+createStudentCheckAction
+    :: forall ctx m. EducatorWorkMode ctx m
+    => EducatorBotSwitch
+    -> m StudentCheckAction
+createStudentCheckAction (EducatorBotOn _) =
+    return . StudentCheckAction . const $ pure True
```
Martoon-00: Maybe we can add student here instead of modifying each endpoint to do that?

Martoon-00 (approved):
LGTM

educator/src/Dscp/Educator/Web/Server.hs
flyingleafe: good point...

flyingleafe: Ah, no, I'd better not do it, it will break down all the tests on Educator.

educator/src/Dscp/Educator/Web/Auth.hs
```diff
@@ -69,21 +74,47 @@ instance ( HasServer api ctxs
...
+    request <- ask
+    token <- maybe mempty pure $ authBearerToken request
+    either (const mempty) pure $ do
+        jWitness <- first (show @Text @Error) $
+            decodeCompact $ LBS.fromStrict token
+        verifyJWitness jWitness
+
+-- | Authentication header timeout in seconds.
+-- TODO: 1) make it configurable via config file
+--       2) isn't 5 minutes too long?
```
hithroc: It was set to 5 minutes to account for the fact that the client might have their system clock behind by a couple of minutes.

specs/disciplina/educator/api/student.yaml
```diff
@@ -242,21 +242,30 @@ securityDefinitions:
     type: apiKey
     name: Authorization
     in: header
-    description: |
-      Authentication system requires a signed JWT in Authorization header in `Bearer <JWT>` format. Example header:
+    description: |
```
hithroc: Trailing whitespace

educator/src/Dscp/Educator/Web/Server.hs
Martoon-00: But we have only 3 tests with bot :thinking:

flyingleafe: Let's not risk bot correctness without necessity until we ship release. Task for the future: https://issues.serokell.io/issue/DSCP-289

PR #281, opened 2018-10-01
[DSCP-298] Implement private block publishing worker

Martoon-00:
### Description

Make educator create and publish private blocks periodically.

### YT issue

https://issues.serokell.io/DSCP-298

### Introduced changes

- [x] New feature
- [ ] Bugfix
- [ ] Refactoring
- [ ] New tests (on functionality not fixed/introduced in this PR)
- [ ] New docs (on functionality not fixed/introduced in this PR)
- [ ] Infrastructure changes

### Checklist

If I added new functionality, I
- [ ] added tests covering it
- [x] explained it using haddock in the source code

If I fixed a bug, I
- [ ] added a regression test to prevent the bug from silently reappearing again

If I changed public API structure or/and data serialization, I made sure that
- [ ] those changes are reflected in [Swagger API specs](/specs/disciplina)
- [ ] and also in [related](/docs/api-types.md) [documentation](/docs/authentication.md).

If I changed config structure or/and CLI interface of any executable, I made sure that
- [x] [sample config](/docs/config-full-sample.yaml) is updated accordingly
- [x] [launch scripts](/scripts/launch) are updated accordingly and work as expected

If I changed the procedure of building a project and/or running any executable or helper script, I
- [ ] updated [README.md](/README.md) accordingly
- [ ] as well as subproject READMEs for all related executables

Also,
- [x] my commit history is clean, descriptive and do not contain merge or revert commits or commits like "WIP"
- [x] my code complies with the [style guide](/docs/code-style.md)
- [ ] my documentation is checked with a spell checker (like [Grammarly](https://app.grammarly.com))

### DIFF
--- a/core/src/Dscp/Core/Foundation/Educator.hs
+++ b/core/src/Dscp/Core/Foundation/Educator.hs
@@ -69,6 +69,7 @@ module Dscp.Core.Foundation.Educator
 
     -- * Activity Type Graph
     , ATGDelta (..)
+    , isEmptyATGDelta
     , ATGNode (..)
     , atgnSubjectId
     , atgnChildren
@@ -80,7 +81,7 @@ module Dscp.Core.Foundation.Educator
 
 import Control.Lens (Getter, makeLenses, to)
 import Data.Time.Clock (UTCTime)
-import Fmt (build, mapF, (+|), (|+), genericF)
+import Fmt (build, genericF, mapF, (+|), (|+))
 
 import Dscp.Core.Foundation.Address (Address (..))
 import Dscp.Crypto
@@ -240,6 +241,10 @@ newtype ATGDelta = ATGDelta
 instance Buildable ATGDelta where
     build (ATGDelta d) = "ATGDelta { " +| mapF d |+ " }"
 
+-- | Whether does 'ATGDelta' carries no changes actually.
+isEmptyATGDelta :: ATGDelta -> Bool
+isEmptyATGDelta  = null . getATGDelta
+
 instance Buildable Course where
     build Course{..} = build getCourseId
 
@@ -248,6 +253,7 @@ instance Buildable () where
 
 instance Buildable DocumentType where
     build = genericF
+
 ----------------------------------------------------------------------------
 -- Transactions
 ----------------------------------------------------------------------------
--- a/core/src/Dscp/Util/Timing.hs
+++ b/core/src/Dscp/Util/Timing.hs
@@ -1,0 +1,59 @@
+module Dscp.Util.Timing
+       ( countingTime
+       , foreverAlive
+       , periodically
+       ) where
+
+import qualified Control.Concurrent as C
+import Data.Time.Clock.POSIX (getPOSIXTime)
+import Fmt ((+|), (+||), (|+), (||+))
+import Loot.Log (MonadLogging, logError)
+import Time (KnownRat, KnownRatName, Second, Time, threadDelay, toNum)
+import UnliftIO (MonadUnliftIO)
+
+import Dscp.Util.TimeLimit (logWarningWaitOnce)
+
+-- | Evaluate how much time did action take.
+countingTime :: MonadIO m => m a -> m (Double, a)
+countingTime m = do
+    t <- liftIO $ getPOSIXTime
+    a <- m
+    t' <- liftIO $ getPOSIXTime
+    let diff :: Double
+        diff = fromRational . toRational $ t' - t
+    return (diff, a)
+
+-- | Continuously execute the action, recovering from failures.
+foreverAlive
+    :: (MonadIO m, MonadCatch m, MonadLogging m,
+        KnownRat unit, KnownRatName unit)
+    => Text -> Time unit -> m () -> m a
+foreverAlive name recovery action =
+    forever action `catchAny` \e -> do
+        printNecrologue e
+        threadDelay recovery
+        foreverAlive name recovery action
+  where
+    printNecrologue e =
+        logError $ name |+ " died (" +|| e ||+ "); \
+                   \ressurecting in " +|| recovery ||+ ""
+
+-- | Execute an action periodically till the end of node's days.
+periodically
+    :: (MonadUnliftIO m, MonadCatch m, MonadLogging m, KnownRat unit)
+    => Text -> Time unit -> m () -> m a
+periodically desc period action = loop
+  where
+    periodSec = toNum @Second period
+    loop = do
+        start <- liftIO getPOSIXTime
+        safeAction
+        end <- liftIO getPOSIXTime
+        let remaining = periodSec - (end - start)
+        liftIO $ C.threadDelay (round $ remaining * 1000000)
+        loop
+    safeAction =
+        logWarningWaitOnce period desc $
+        action `catchAny` reportFailure
+    reportFailure e =
+        logError $ "Periodic action '" +| desc |+ "' failed: " +|| e ||+ ""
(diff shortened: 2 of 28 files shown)


Martoon-00:
Not WIP anymore

core/src/Dscp/Util/Timing.hs
```diff
@@ -0,0 +1,43 @@
...
+    loop = do
+        start <- liftIO getPOSIXTime
+        safeAction
+        end <- liftIO getPOSIXTime
+        let remaining = periodSec - (end - start)
+        liftIO $ threadDelay (round $ remaining * 1000000)
+        loop
+    safeAction =
+        logWarningWaitOnce period desc $
+        action `catchAny` reportFailure
```
flyingleafe: I don't think periodic action should stop being repeated on first exception. Use `foreverAlive` here instead, or better - the version of `foreverAlive` which makes sure that recovery delay after catching an exception is such that next action attempt will be performed in the beginning of next period.

docs/config-full-sample.yaml
```diff
@@ -179,6 +179,10 @@ sample:
         enabled: true
         data: []
 
+    # Private blocks publishing params
+    publishing:
+      period: 15s
```
flyingleafe: I know this is just a sample, but it really doesn't make sense to put publishing period less than a slot lenght even in sample.
By the way, displaying some kind of warning if publishing period is less than `slotDuration` seems like a good idea (because it never makes sense).

tools/README.md
```diff
@@ -128,7 +129,17 @@ echo '9JsCkzFH/W7SbgvgokeJTsckJs/uoAPWWwC40Y6UfFo=' | dscp-keygen --secret --com
...
-* `-H "Authorization: Bearer <token>"` for `curl`
-* `Authorization:Bearer\ <token>` to [`http`](https://httpie.org/).
+1. `-H "Authorization: Bearer <token>"` for `curl`
+2. `Authorization:Bearer\ <token>` for [`http`](https://httpie.org/).
+
+* Submit a new submission from student - owner of given secret key:
+
+```base
+echo 456 \
+    | dscp-keygen --seed --command student-submission:3656234 \
```
flyingleafe: Person reading the README will not get what is this number. Point out that this is the random seed.

witness/src/Dscp/CommonCLI.hs
```diff
@@ -94,13 +92,8 @@ appDirParamParser = AppDirectorySpecific <$>
...
 -- | Parses time with specified unit of measurement, e.g. @10s@.
-timeReadM :: KnownRat unit => ReadM (Time unit)
-timeReadM = asum
-    [ toUnit @_ @Second <$> auto
-    , toUnit @_ @Millisecond <$> auto
-    , toUnit @_ @Microsecond <$> auto
-    , toUnit @_ @Minute <$> auto
-    ]
+timeReadM :: KnownRatName unit => ReadM (Time unit)
+timeReadM = nothingToFail "Cannot parse time unit" . unitsP =<< str
```
flyingleafe: http://hackage.haskell.org/package/optparse-applicative-0.14.2.0/docs/Options-Applicative.html#v:maybeReader
But I already used it in #284 :stuck_out_tongue:

Martoon-00: Yeah now I see it, your version is better ))

core/src/Dscp/Util/Timing.hs
Martoon-00: Mm, actually I tried to merge `periodically` and `foreverAlive`, and got something unpleasant. These functions are too different: exceptions are handles in distinct ways, and one waits for a given time disregard possible exceptions while another executes the action at max rate.

And current `periodically` function tolerates exceptions.

educator/src/Dscp/Educator/CLI.hs
```diff
@@ -88,10 +89,23 @@ educatorKeyParamsParser :: Parser EducatorKeyParams
 educatorKeyParamsParser =
     EducatorKeyParams <$> baseKeyParamsParser "educator"
 
+publishingPeriodParser :: Parser (Time Second)
+publishingPeriodParser = option timeReadM $
+    long "publish-period" <>
```
Heimdell: Maybe, `publication-period`? Now it sound and if you're publishing a period.

educator/src/Dscp/Educator/Workers.hs
```diff
@@ -0,0 +1,74 @@
...
+        toUnit . ms $ fromIntegral $ unSlotDuration $ giveL @WitnessConfig
+
+    bootstrap =
+        unless (period > slotDuration) $
+            logWarning $ "Private block publishing period is not greater than \
+                         \witness slot duration ("
+                         +|| period ||+ " <= " +|| slotDuration ||+ ")"
+
+    work = periodically "Private block publisher" period $ do
+        mblock <- transactW $ runMaybeT (createPrivateBlock Nothing)
```
Heimdell: Hmm, is `MaybeT` wrapper only applied to this concrete function?

educator/tests/Test/Dscp/DB/SQLite/Queries.hs
```diff
@@ -302,7 +302,7 @@ spec_Instances = do
                         ptId <- DB.createTransaction trans
                         return ptId
 
-                    DB.createBlock Nothing
+                    _ <- runMaybeT $ DB.createPrivateBlock Nothing
```
Heimdell: "Fail silently" is not the best behaviour, I believe.

educator/src/Dscp/DB/SQLite/Queries.hs
```diff
@@ -422,9 +424,15 @@ createBlock delta = do
 
         hdr = PrivateBlockHeader prev root trueDelta
 
+    let isNullBlock = and
+            [ nullATGDelta trueDelta
+            , null txs
+            ]
+    guard (not isNullBlock)
```
Heimdell: Why don't you just throw an exception out of this function?
Later the output is just mostly ignored.
This is an inconsistency of this function behaviour, which can mess up things later.

I'm proposing throwing `NoNeedToCreateANewBlock` and `catch`ing it on the call sites instead, to show that this thing might happen.

Martoon-00: For me semantics of `m (Maybe a)` is "function performs an action and returns or does not return something", while `MaybeT m a` means "action may be performed or not" which is exactly what we want here.

educator/tests/Test/Dscp/DB/SQLite/Queries.hs
Martoon-00: Right, makes sense

educator/src/Dscp/Educator/Workers.hs
Martoon-00: What do you mean?

core/src/Dscp/Core/Foundation/Educator.hs
```diff
@@ -240,6 +241,10 @@ newtype ATGDelta = ATGDelta
 instance Buildable ATGDelta where
     build (ATGDelta d) = "ATGDelta { " +| mapF d |+ " }"
 
+-- | Whether does 'ATGDelta' carries no changes actually.
+nullATGDelta :: ATGDelta -> Bool
+nullATGDelta  = null . getATGDelta
```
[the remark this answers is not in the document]
Martoon-00: It is an exact analogy of `null` function, but ok, if at least one person finds it looking unpleasant it's worth changing.

Martoon-00:
Since there are no further comments, I merge.
If you still have objections, I can comply them in upcoming related PRs

PR #284, opened 2018-10-02
[DSCP-295] Merge multiple configs properly + fix CLI implicit overrides.

flyingleafe:
### Description

The following is done:
- Multiple YAML configs are now merged as `Value`s before parsing as `ConfigRec`s. This allows for overriding separate fields in records.
- There are no CLI parsers which always succeed anymore (those which did this are changed and now fail if no relevant parameters are passed). This behavior together with overriding config sections via CLI params led to very confusing bugs where some config parameters were seemingly ignored. Now default values for those parameters are provided via default config records, which do not override Yaml configs.

### YT issue

https://issues.serokell.io/issue/DSCP-295

### Checklist

- [x] I have checked the [sample config](/../../tree/master/docs/config-full-sample.yaml) and [launch scripts](/../../tree/master/scripts/launch) and updated them if it was necessary (especially if executables or CLIs were changed).
- [x] I have checked READMEs and changed them accordingly if it's necessary (the main [README.md](/README.md) as well as subproject READMEs for all related executables).

Stylistic (obligatory):

### DIFF
--- a/witness/src/Dscp/CommonCLI.hs
+++ b/witness/src/Dscp/CommonCLI.hs
@@ -20,12 +20,11 @@ module Dscp.CommonCLI
 import Data.Char (toLower)
 import Data.Version (showVersion)
 import qualified Loot.Log as Log
-import Options.Applicative (Parser, ReadM, auto, eitherReader, help, infoOption, long, metavar,
-                            option, str, strOption, switch)
+import Options.Applicative (Parser, ReadM, auto, eitherReader, flag', help, infoOption, long,
+                            maybeReader, metavar, option, str, strOption, switch)
 import Servant.Client (BaseUrl (..), parseBaseUrl)
 import Text.InterpolatedString.Perl6 (qc)
-import Time.Rational (KnownRat)
-import Time.Units (Microsecond, Millisecond, Minute, Second, Time, toUnit)
+import Time (KnownRatName, Time, unitsP)
 
 import Dscp.Core.Foundation.Address
 import Dscp.Core.Foundation.Witness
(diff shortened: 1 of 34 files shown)


mkaito:
Does this allow for intuitive recursive merge of multiple YAML files, or should I fall back on `yt` for merges?

flyingleafe:
@mkaito yes it does allow for recursive merges, I meant this in the 1 paragraph

witness/src/Dscp/CommonCLI.hs
```diff
@@ -38,11 +38,25 @@ import Dscp.Util
...
+default config values are provided in respective `*.Config` modules.
+-}
+
 logParamsParser :: Log.Name -> Parser LoggingParams
 logParamsParser lpDefaultName = do
     lpDebug <- logDebugParser
     lpConfigPath <- logConfigParser
-    lpDirectory <- logDirParser
+    -- [Note default-cli-params]
+    lpDirectory <- Just <$> logDirParser
```
Martoon-00: I don't think this makes sense. Now you have to provide both `--log-config` and `--log-dir` in order to make _any_ of these options take place, and there is no way to specify no file for logs.

witness/src/Dscp/CommonCLI.hs
```diff
@@ -84,14 +99,15 @@ baseKeyParamsParser who = do
...
-                    pure AppDirectoryOS
+appDirParamParser = fmap AppDirectorySpecific . strOption $
+     long "appdir" <>
+     metavar "FILEPATH" <>
+     help "Path to application folder. If not specified, \
+          \OS-dependent folder for applications will be \
+          \used, for instance '%APPDIR%/Disciplina'."
+     -- Removed default 'AppDirectoryOS'
+     -- [Note default-cli-params]
+
```
Martoon-00: This leaves no way to set `AppDirectoryOS` overriding config.
(Actually, this parser should've never been optional, one should've explicitly passed some `--os-appdir` option to make this parser return `AppDirectoryOS`)

witness/src/Dscp/Witness/Launcher/Resource.hs
```diff
@@ -52,8 +52,8 @@ getWitnessKeyResources
     => WitnessKeyParams
     -> AppDir
     -> m (KeyResources WitnessNode)
-getWitnessKeyResources (Basic bkp) appDir = linkStore bkp appDir
-getWitnessKeyResources (Committee cp) _   = genStore (Just cp)
+getWitnessKeyResources (Basic bkp) appDir = logInfo "MDA BASIC" >> linkStore bkp appDir
+getWitnessKeyResources (Committee cp) _   = logInfo "KAEF COMM" >> genStore (Just cp)
```
Martoon-00: Lool

Martoon-00 (changes requested):
There are a couple of changes I disagree with.

witness/src/Dscp/CommonCLI.hs
Martoon-00: Okay. I was thinking that leaving `logDirParser` optional would be better because `logConfigParser` should parse successfully to make whole `logParamsParser` take place, but it resulting behaviour may be even more confusing.
At least fix help in `logDirParser`.

witness/src/Dscp/CommonCLI.hs
```diff
@@ -38,11 +38,25 @@ import Dscp.Util
 import Dscp.Web (NetworkAddress (..), ServerParams (..), parseNetAddr)
 import Paths_disciplina_witness (version)
 
+{- |
```
Martoon-00: Minor: I also want to start using such notes and checked which format are you going to keep them in.

And I couldn't ensure whether is such haddock syntax valid, and it definitely is, however, it looks like
screenshot from 2018-10-03 18-40-13
when I insert this block above `traceShow` with its comment. So probably `|` would be an extra here.

flyingleafe:
@Martoon-00 bump, re-review pls

Martoon-00 (approved):
Thanks, LGTM now

PR #286, opened 2018-10-03
[DSCP-314] Prohibit empty CommitteeSecrets

nasportivnoy:
### Description

Prohibit empty `CommitteeSecret`s (throw an error when trying to decode empty string)
Add smart constructor to CommitteeSecret
Change exports and imports

### YT issue

https://issues.serokell.io/issue/DSCP-314

### Checklist

- [ ] I have checked the [sample config](/../../tree/master/docs/config-full-sample.yaml) and [launch scripts](/../../tree/master/scripts/launch) and updated them if it was necessary (especially if executables or CLIs were changed).
- [ ] I have checked READMEs and changed them accordingly if it's necessary (the main [README.md](/README.md) as well as subproject READMEs for all related executables).

Stylistic (obligatory):

### DIFF
--- a/core/src/Dscp/Core/Aeson.hs
+++ b/core/src/Dscp/Core/Aeson.hs
@@ -111,7 +111,7 @@ instance ToJSONKey Address
 deriving instance FromJSON GenAddressMap
 deriving instance ToJSON GenAddressMap
 instance FromJSON CommitteeSecret where
-    parseJSON = withText "CommitteeSecret" $ pure . CommitteeSecret . encodeUtf8
+    parseJSON = withText "CommitteeSecret" $ leftToFail . mkCommitteeSecret . encodeUtf8
 
 deriving instance ToJSON GenesisDistribution
 deriving instance FromJSON GenesisDistribution
--- a/tools/src/keygen/KeygenSecret.hs
+++ b/tools/src/keygen/KeygenSecret.hs
@@ -45,7 +45,7 @@ parseInputWithSecret = \case
         content <- decode $ LBS.fromStrict input
         fromKeyfileContent pp content
     CommSecret n -> \input ->
-        pure $ committeeDerive (CommitteeSecret input) n
+         flip committeeDerive n <$> eitherToMaybe (mkCommitteeSecret input)
     SecretFromSeed -> \input -> asum
         [ do
             -- we need this clause as soon as many code uses
--- a/witness/src/Dscp/Witness/CLI.hs
+++ b/witness/src/Dscp/Witness/CLI.hs
@@ -19,13 +19,14 @@ import Options.Applicative (Parser, auto, eitherReader, help, long, metavar, opt
 
 import Dscp.CommonCLI (appDirParamParser, baseKeyParamsParser, logParamsParser,
                        networkAddressParser, serverParamsParser)
-import Dscp.Core.Governance (CommitteeSecret (..))
+import Dscp.Core.Governance (mkCommitteeSecret)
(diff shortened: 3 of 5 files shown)


flyingleafe:
@nasportivnoy you apparently don't run `stack test` before pushing, don't you :peka:

core/src/Dscp/Core/Aeson.hs
```diff
@@ -111,7 +111,7 @@ instance ToJSONKey Address
 deriving instance FromJSON GenAddressMap
 deriving instance ToJSON GenAddressMap
 instance FromJSON CommitteeSecret where
-    parseJSON = withText "CommitteeSecret" $ pure . CommitteeSecret . encodeUtf8
+    parseJSON = withText "CommitteeSecret" $ pure . mkCommitteeSecret . encodeUtf8
```
flyingleafe: This will lead not to graceful parser failure, but to a failure which would propagate all the way up to the IO monad. Same with other parsers. Make `mkCommitteeSecret` return `Either` instead and fail gracefully - this is a way to go, esp. given that the only places `CommitteeSecret` is constructed are some kinds of parsers.

flyingleafe:
@nasportivnoy also, where are your ticks in checkboxes?

witness/tests/Test/Dscp/Witness/Mode.hs
```diff
@@ -77,7 +77,7 @@ testCommittee :: Committee
 testCommittee =
     CommitteeOpen
     { commN = 2
-    , commSecret = detGen 12 (CommitteeSecret <$> arbitrary)
```
nasportivnoy: `detGen 12` produces nullary bytestring

tools/src/keygen/KeygenSecret.hs
```diff
@@ -45,7 +45,7 @@ parseInputWithSecret = \case
         content <- decode $ LBS.fromStrict input
         fromKeyfileContent pp content
     CommSecret n -> \input ->
-        pure $ committeeDerive (CommitteeSecret input) n
+        pure $ committeeDerive (leftToPanic $ mkCommitteeSecret input) n
```
flyingleafe: why `leftToPanic`? this is a method which returns `Maybe` exactly to be able to fail gracefully, you should use `rightToMaybe` here.

witness/src/Dscp/Witness/CLI.hs
```diff
@@ -104,7 +105,7 @@ committeeParamsParser =
          help "Committee participant index. In event of secret key file \
               \generation, will be used to derive the secret.")
 
-    commSecretParser = CommitteeSecret <$> strOption
+    commSecretParser = leftToPanic . mkCommitteeSecret <$> strOption
```
flyingleafe: Again, why panic? this is a parser which can fail (with `empty`), so if an empty string is encountered, it should fail properly.

tools/src/keygen/KeygenSecret.hs
```diff
@@ -45,7 +45,7 @@ parseInputWithSecret = \case
         content <- decode $ LBS.fromStrict input
         fromKeyfileContent pp content
     CommSecret n -> \input ->
-        pure $ committeeDerive (CommitteeSecret input) n
+         liftA2 committeeDerive (eitherToMaybe $ mkCommitteeSecret input) $ pure n
```
flyingleafe: Wow such constructions, `flip committeeDerive n <$> eitherToMaybe (mkCommitteeSecret input)` seems much better.

PR #289, opened 2018-10-05
[DSCP-302] Add endpoint to get publications

Martoon-00:
### Description

1. Separated endpoints to see money and publication txs;
2. Add endpoint to see all publications of single educator.

### YT issue

https://issues.serokell.io/issue/DSCP-302

### Checklist

- [x] I have checked the [sample config](/../../tree/master/docs/config-full-sample.yaml) and [launch scripts](/../../tree/master/scripts/launch) and updated them if it was necessary (especially if executables or CLIs were changed).
- [x] I have checked READMEs and changed them accordingly if it's necessary (the main [README.md](/README.md) as well as subproject READMEs for all related executables).

Stylistic (obligatory):

### DIFF
--- a/specs/disciplina/witness/api/witness.yaml
+++ b/specs/disciplina/witness/api/witness.yaml
@@ -79,7 +79,7 @@ paths:
           schema:
             type: object
             properties:
-              transactions:
+              blocks:
                 type: array
                 items:
                   $ref: "#/definitions/BlockInfo"
@@ -178,15 +178,6 @@ paths:
         type: integer
         format: int32
       - in: query
-        name: txType
-        description: Filter on transaction type.
-        default: all
-        type: string
-        enum:
-          - all
-          - money
-          - publications
-      - in: query
         name: account
         description: Remain only transactions affecting given account.
         type: string
@@ -250,7 +241,7 @@ paths:
           description: Successful operation
           schema:
             allOf:
-            - $ref: "#/definitions/TransactionInfo"
+            - $ref: "#/definitions/GeneralizedTransactionInfo"
             - $ref: "#/definitions/AndBlockInfo"
         400:
           description: Invalid input
@@ -316,7 +307,7 @@ paths:
                 type: array
                 items:
                   allOf:
-                  - $ref: "#/definitions/PrivateBlockInfo"
+                  - $ref: "#/definitions/PublicationInfo"
                   - $ref: "#/definitions/AndBlockInfo"
               nextId:
                 $ref: "#/definitions/Hash"
@@ -394,7 +385,7 @@ definitions:
       currentNonce:
         type: integer
         format: int32
-  TransactionInfo:
+  GeneralizedTransactionInfo:
     type: object
     discriminator: txType
     properties:
@@ -403,6 +394,17 @@ definitions:
         enum:
         - money
         - publication
+      txId:
+        $ref: "#/definitions/Hash"
+      outValue:
+        $ref: "#/definitions/Coin"
+      money:
+        $ref: "#/definitions/TransactionInfo"
+      publication:
+        $ref: "#/definitions/PublicationInfo"
+  TransactionInfo:
+    type: object
+    properties:
       txId:
         $ref: "#/definitions/Hash"
       outValue:
@@ -429,15 +431,20 @@ definitions:
                   $ref: "#/definitions/Address"
                 value:
                   $ref: "#/definitions/Coin"
+  PublicationInfo:
+    type: object
+    properties:
+      txId:
+        $ref: "#/definitions/Hash"
       publication:
         type: object
         properties:
           author:
             $ref: "#/definitions/Address"
-          block:
-            $ref: "#/definitions/Hash"
-          prevBlock:
-            $ref: "#/definitions/Hash"
+          feesAmount:
+            $ref: "#/definitions/Coin"
+          header:
+            $ref: "#/definitions/PrivateBlockHeader"
   AndBlockInfo:
     type: object
     properties:
@@ -452,13 +459,22 @@ definitions:
           allOf:
           - $ref: "#/definitions/TransactionInfo"
           - $ref: "#/definitions/AndBlockInfo"
+  GeneralizedTransactionList:
+    type: object
+    properties:
+      transactions:
+        type: array
+        items:
+          allOf:
+          - $ref: "#/definitions/GeneralizedTransactionInfo"
+          - $ref: "#/definitions/AndBlockInfo"
   BlockTransactionList:
     type: object
     properties:
       transactions:
         type: array
         items:
-          $ref: "#/definitions/TransactionInfo"
+          $ref: "#/definitions/GeneralizedTransactionInfo"
   TxInAcc:
     type: object
     description: Transaction source account info.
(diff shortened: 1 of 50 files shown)


specs/disciplina/witness/api/witness.yaml
```diff
@@ -216,6 +216,84 @@ paths:
           schema:
             $ref: '#/definitions/TxValidationError'
 
+  /transactions/money:
```
flyingleafe: It should be either a filter parameter to `/transactions` endpoint (like `type=(money|pub|all)`, with default value `all`), if we don't want to separate publications and transactions completely, or it should be two endpoints `/transactions` (existing one) and `/publications`, otherwise. In the latter case, a single type `TransactionInfo` should be split into two types representing info about money transaction and info about publication.

Your current approach is unnecessarily verbose.

specs/disciplina/witness/api/witness.yaml
```diff
@@ -264,6 +342,50 @@ paths:
           schema:
             $ref: "#/definitions/Hash"
 
+  /publications/educator/{educator}:
```
flyingleafe: Why not just add `includePubs` parameter, similarly to `includeTxs` one, to `/account/{address}` endpoint? Every Educator is an account and every account can be an Educator (in principle).
If you'd like to fetch transaction histories via separate endpoints, this makes sense (because, well, current approach with `includeTxs` is actually bullshit, because it doesn't provide any limiting/pagination). But then the most clean solution would be adding an `account` filter parameter to `/transaction` endpoint (or, if you'd like, `/transactions` and `/publications` endpoints).

witness/src/Dscp/Witness/Web/Logic.hs
```diff
@@ -6,18 +6,26 @@ module Dscp.Witness.Web.Logic
...
+       , getEducatorPublications
        , getHashType
        ) where
 
 import Codec.Serialise (serialise)
 import qualified Data.ByteString.Lazy as BS
+import Data.Conduit ((.|))
+import qualified Data.Conduit as C
+import qualified Data.Conduit.Combinators as C
+import qualified Data.Conduit.Lift as C
```
flyingleafe: my gosh finally :heart:

specs/disciplina/witness/api/witness.yaml
Martoon-00: Fair, adding a filter to `/transactions` should work.

specs/disciplina/witness/api/witness.yaml
Martoon-00: Let's assume I prefer a cleaner solution.
Just adding `account` filter parameter to `/transactions` is a bit difficult, because `/transactions` returns transactions with minimal info (`author`/`block`/`prevBlock` in `TxPublication` case), while `/publications` returns more elaborate private blocks (`fee`/`merkleRoot`/`transactions number`/`ATG`). Maybe this is a very bad and confusing difference.

flyingleafe: @Martoon-00 well, fair enough, that just means that because we don't have a proper way to encode sum types in JSON, the `/transactions` and `/publications` endpoints should be separate, and each of them should have the `account` filter parameter.

flyingleafe: @Martoon-00 just do for `/transactions` the same what `includeTxs` flag does in `/account` endpoint, but with limits.

witness/src/Dscp/Witness/Logic/Getters.hs
```diff
@@ -138,20 +134,27 @@ getAccountTxs address = runSdMRead $ loadTxs >>= mapM getTx
...
     maybe (SD.throwLocalError $ LETxAbsent $
               "Can't get transaction with id " <> pretty gTxId)
           pure
           tM
+
+----------------------------------------------------------------------------
+-- Publication getters
+----------------------------------------------------------------------------
+
+getPrivateTipHash :: HasWitnessConfig => Address -> SdM_ chgacc (Maybe PrivateHeaderHash)
```
volhovm: Why `chgacc` is abstracted here?

witness/src/Dscp/Witness/Web/Client/Logic.hs
```diff
@@ -58,4 +59,4 @@ instance AllocResource WitnessClient where
     allocResource params =
         buildComponentR "witness client"
             (createWitnessClient params)
-            (\_ -> pass)  -- items clean on itself on GC, kaef
+            (\_ -> pass)  -- things clean up on themselves on GC, kaef
```
volhovm: by themselves :peka:

witness/src/Dscp/Witness/Logic/Getters.hs
Martoon-00: It allows running an action with both `runSdRead` and `runSdMempoolRead`.

witness/src/Dscp/Witness/Web/Client/Logic.hs
Martoon-00: :woman_facepalming: :woman_facepalming:

Martoon-00:
Ok, I think I'm finally done here.
Here is how `/publications` response looks like: https://pastebin.com/k8QFmi51
And `/transactions`: https://pastebin.com/qimkr5uH

witness/src/Dscp/Witness/Web/Types.hs
```diff
@@ -119,35 +160,72 @@ deriveJSON defaultOptions ''BlocksOrMempool
...
-                , "txType" .= ("publication" :: Text)
-                , "publication" .= pTx
-                ]
-
-instance FromJSON TxInfo where
-    parseJSON = withObject "tx info" $ \o -> do
+
+instance ToJSON (Detailed a) => ToJSON (WithBlockInfo a) where
+    toJSON WithBlockInfo{..} = mergeObjects
+        (object $ maybe [] (\block -> ["block" .= block]) wbiBlockInfo)
```
flyingleafe: Why do you need this function `mergeObjects` here, if you just try to add a `block` field to `Detailed a` serialization? Why just not add it manually? Seems like an overkill.

witness/tests/Test/Dscp/Witness/Explorer/ExplorerSpec.hs
```diff
@@ -0,0 +1,143 @@
...
+    :: (WitnessWorkMode ctx m, WithinWriteSDLock)
+    => Gen SecretKey -> PropertyM m Tx
+createAndSubmitTx genSecret = do
+    sk <- pick $ mkSecretKeyData <$> genSecret
+    outs <- pick $ genSafeTxOuts 100 (choose (1, 5))
+    account <- lift $ fromMaybe def <$> getMempoolAccountMaybe (skAddress sk)
+
+    let txw = createTxw (fcMoney feeConfig) sk (aNonce account) outs
+    isNew <- lift $ addTxToMempool (GMoneyTxWitnessed txw)
+    unless isNew $ error "Duplicated transaction???"
```
flyingleafe: Why `error` and not `fail`?

witness/tests/Test/Dscp/Witness/Explorer/ExplorerSpec.hs
```diff
@@ -0,0 +1,143 @@
...
+            , _pbhAtgDelta = mempty
+            }
+        tx = PublicationTx
+            { ptAuthor = skAddress sk
+            , ptFeesAmount = unFees $ calcFeePub (fcPublication feeConfig) ptHeader
+            , ptHeader
+            }
+    let txw = signPubTx sk tx
+    isNew <- lift $ addTxToMempool (GPublicationTxWitnessed txw)
+    unless isNew $ error "Duplicated transaction???"
```
flyingleafe: Same

witness/src/Dscp/Witness/Web/Types.hs
Martoon-00: What do you mean under "manually"?
I can't prove compiler that `Detailed a` serializes into an object, so can't add a field in a fully safe way.

PR #300, opened 2018-10-19
[DSCP-345] syslog backend for logging

pasqu4le:
### Description

This adds syslog backend for logging.

NOTE: it is labeled WIP because it relies on a [yet unmerged branch](https://github.com/serokell/lootbox/tree/pasqu4le/lbt42-syslog-backend-for-loot-log) of `loot-box` made for issue [LTB-42](https://issues.serokell.io/issue/LTB-42)

There is now a new configuration option: `log-type` that defaults to `Syslog` but can be set to `Warper`. A new logging configuration file, that mimics the still-present warper one as much as possible, has been added: `run/syslog-config.yaml`.

### YT issue

https://issues.serokell.io/issue/DSCP-345

### Checklist

- [X] I added tests if required, in case they are:
- [X] I have checked the [sample config](../tree/master/docs/config-full-sample.yaml) and [launch scripts](../tree/master/scripts/launch) and updated them if it was necessary (especially if executables or CLIs were changed).
- [X] I have checked READMEs and changed them accordingly if it's necessary (the main [README.md](../tree/master/README.md) as well as subproject READMEs for all related executables).
- [X] If a public API structure or/and data serialization was changed, I made sure that these changes are reflected in the:
- [X] If any public documentation was written, I have spellchecked it.

Stylistic (obligatory):
- [X] My commit history is clean, descriptive and do not contain merge or revert commits or commits like "WIP".
- [X] My code complies with the [style guide](../tree/master/docs/code-style.md).
- [X] My code is documented (haddock) and these docs are decent (full enough and spellchecked).

### DIFF
--- a/core/src/Dscp/Resource/Logging.hs
+++ b/core/src/Dscp/Resource/Logging.hs
@@ -4,18 +4,21 @@
 
 module Dscp.Resource.Logging
     ( LoggingParams(..)
+    , LoggingType (..)
     , allocLogging
     , noLogging
     ) where
 
 import Control.Monad.Component (ComponentM, buildComponent)
-import Data.Aeson (FromJSON (..), encode)
+import Data.Aeson (FromJSON (..), ToJSON, encode, withText)
 import Data.Aeson.Options (defaultOptions)
 import Data.Aeson.TH (deriveFromJSON)
 import Fmt ((+|), (|+))
-import Loot.Log (Logging (..), Name, NameSelector (CallstackName, GivenName), logInfo,
-                 modifyLogName)
+import Loot.Log (Logging (..), Name, NameSelector (CallstackName, GivenName),
+                 logInfo, logWarning, modifyLogName, Level(Debug))
 import Loot.Log.Rio (LoggingIO)
+import Loot.Log.Syslog (SyslogConfig(..), loggerName, withPERROR, minLevel,
+    stopLoggers, syslogConfigFromFile, defaultLoggerConfig, prepareSyslog)
 import Loot.Log.Warper (LoggerConfig, prepareLogWarper)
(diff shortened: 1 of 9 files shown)


core/src/Dscp/Resource/Logging.hs
```diff
@@ -56,22 +62,55 @@ getRealLoggerConfig LoggingParams{..} = do
...
-    printCfg logging finalConfig =
+    selectedType = fromMaybe Syslog $ lpLoggingType params -- Syslog is default
+
+    pre = case selectedType of
+        Warper -> do
+            config <- getWarperLoggerConfig params
+            (config', logging) <-
+                prepareLogWarper config (GivenName $ lpDefaultName params)
+            printCfg logging config Nothing
+            printCfg logging config' Nothing
```
flyingleafe: This will print the same message twice with slightly different configs. Why? I find this kind of misleading and redundant.

core/src/Dscp/Resource/Logging.hs
```diff
@@ -80,6 +119,12 @@ allocLogging params = buildComponent "logging" pre fin
 instance FromJSON Name where
     parseJSON = fmap fromString . parseJSON
 
+instance FromJSON LoggingType where
+    parseJSON = withText "LoggingType" $ \case
+        "Syslog" -> pure Syslog
+        "Warper" -> pure Warper
+        _ -> fail "Unknown logging type. Available options: Sysylog or Warper"
```
flyingleafe: Typo: `Sys_y_log`

core/src/Dscp/Resource/Logging.hs
pasqu4le: I don't know why, I left it (for `log-warper` only) because it was there before, should I print just the second one?

flyingleafe: Ah, missed the fact it was here before. Yes, I think, printing the second one is sufficient.

PR #311, opened 2018-11-05
DSCP-359 Standard logging configuration for wallet

nasportivnoy:
### YT issue

https://issues.serokell.io/issue/DSCP-359

### Checklist

- [ ] I have checked the [sample config](../tree/master/docs/config-full-sample.yaml) and [launch scripts](../tree/master/scripts/launch) and updated them if it was necessary (especially if executables or CLIs were changed).

Stylistic (obligatory):
- [ ] My code complies with the [style guide](../tree/master/docs/code-style.md).
- [ ] My code is documented (haddock) and these docs are decent (full enough and spellchecked).

### DIFF
--- a/wallet/src/Dscp/Wallet/CLI.hs
+++ b/wallet/src/Dscp/Wallet/CLI.hs
@@ -1,31 +1,46 @@
 module Dscp.Wallet.CLI
-       ( WalletCLIParams (..)
-       , getWalletCLIParams
+       ( getWalletConfig
+       , getCLIWalletConfig
        ) where
 
 import Options.Applicative (execParser, fullDesc, help, helper, info, long, metavar, optional,
-                            progDesc, strOption)
+                            progDesc, strOption, Parser)
 
+import Loot.Config (OptModParser, (.::), (.:<))
 import Dscp.CommonCLI
 import Dscp.Config
-import Dscp.Web
+import Dscp.Wallet.Config
 
-data WalletCLIParams = WalletCLIParams
-    { wpWitness      :: BaseUrl
-    , wpKnitCommand  :: Maybe Text
-    , wpConfigParams :: ConfigParams
-    }
+knitCommandParser :: Parser (Maybe Text)
+knitCommandParser = optional $ strOption $
+    long "knit" <>
+    metavar "COMMAND" <>
+    help "Execute provided knit command and exit."
 
-getWalletCLIParams :: IO WalletCLIParams
-getWalletCLIParams = do
-    let parser = do
-            wpWitness <- clientAddressParser "witness" "Address of a witness node to communicate with"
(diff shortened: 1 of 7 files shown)


wallet/src/Dscp/Wallet/Config.hs
```diff
@@ -0,0 +1,68 @@
...
+import Dscp.Config
+import Dscp.Core.Config
+import Dscp.Resource.Logging (LoggingParams (..), basicLoggingParams)
+import Dscp.Web
+
+type WalletConfig = CoreConfig ++
+    '[ "wallet" ::<
+       '[ "logging" ::: LoggingParams
+        , "witness" ::: BaseUrl
+        , "knitCommand" ::: Maybe Text
```
flyingleafe: Arguably, `knitCommand` should not actually be in config, it makes no sense to put it into config so that a wallet always runs a single command on start by default. Probably it's better to parse this separately only from CLI and pass `Maybe Text` value directly to `initializeEverything` as an argument. Especially given that it's only used for `dscp-wallet-cli`, but not for TUI `dscp-wallet`.

flyingleafe (changes requested):
Please, address my comment.
Regarding what to do about actual logging:
1) Inside `createWalletFace`, a _context_ should be created first (which should include `SendEvent` and `LoggingIO`, obtained from `allocLogging`
2) All functions which are included into `WalletFace` should then run inside RIO with this context instead of just simple IO with passed arguments
3*) Helpers `walletLogInfo`, `walletLogWarning` and `walletLogError` should be created which do _both_ `sendLogEvent` and corresponding logging action from `loot-log`
4) After `co-log` logging backend by @pasqu4le is approved and used, we might drop (3*) and instead create a custom `LogAction` which does `sendLogEvent` and combine it with logging backend, so that `logInfo`, `logWarning` etc. invoke `sendLogEvent` by default. But this can be done later.

wallet/src/Dscp/Wallet/CLI.hs
```diff
@@ -1,31 +1,46 @@
...
+getCLIWalletConfig = do
+    let parser = (,,) <$> configParamsParser <*>
+                          walletConfigParser <*>
+                          knitCommandParser
+    (configParams, cliConfigMod, knitCommand) <- execParser $
+        info (helper <*> versionOption <*> parser) $
+        fullDesc <> progDesc "Ariadne cli wallet for Disciplina"
+    let wrapConfig cfg = cliConfigMod $ defaultWalletConfig <> cfg
+    configRec <- buildConfig configParams $
+        fmap wrapConfig . fillWalletConfig
```
flyingleafe: This copy-paste between `getWalletConfig` and `getCLIWalletConfig` can be for sure reduced.

flyingleafe (changes requested):

PR #313, opened 2018-11-05
[DSCP-285] Fix `loop in publication` test case

Martoon-00:
### Description

**Should work after DSCP-335 is merged.**

*Problem:* in rare cases two educators can create identical private block headers
(e.g. if they are empty and the first ones in the chain, then all fields will match).
This leads to data with the same keys being put into storage.

*Solution:* add 'author' field to the key along with the 'headerHash' field.

*Possible alternative solutions:*
* Forbid empty private blocks in validation. In tests generate all different merkle
signatures, assuming that all `PrivateTx`s are different (and they are since they
contain timestamp).

### YT issue

https://issues.serokell.io/issue/DSCP-285

### Checklist

- [ ] I have checked the [sample config](../tree/master/docs/config-full-sample.yaml) and [launch scripts](../tree/master/scripts/launch) and updated them if it was necessary (especially if executables or CLIs were changed).

Stylistic (obligatory):

### DIFF
--- a/educator/tests/Test/Dscp/Educator/Web/Educator/Queries.hs
+++ b/educator/tests/Test/Dscp/Educator/Web/Educator/Queries.hs
@@ -61,7 +61,7 @@ spec_EducatorApiQueries = describe "Basic database operations" $ do
                               \(CourseEducatorInfo courseId desc subjs) ->
                                   (Just courseId, desc, subjs)
             return $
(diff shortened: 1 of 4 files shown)


Martoon-00:
We decided to, probably, make genesis hash of each private chain unique instead.

PR #317, opened 2018-11-12
[DSCP-193] Fixed accidentally broken config

Heimdell:
### Description

Fixing config I broke when doing DSCP-193

### YT issue

https://issues.serokell.io/issue/DSCP-354

### Checklist

Stylistic (obligatory):

### DIFF
--- a/config.yaml
+++ b/config.yaml
@@ -36,11 +36,6 @@ demo: &demo
           - # Test key #2, secret: fOhZOFr6SSnYv1/xfKPPMHdWLdQDboysJ4Q794Gezmw=
             - 3BAyX5pSiFerJgLsJZ9XyoBDMHPtq3NLqm5AGPcBpba6W644zSiSWy3vVr
             - 1000
-  witness:
-    db:
-      path: "/var/db/dscp-witness"
-      clean: false
-
     logging:
       defaultName: "witness"
       config:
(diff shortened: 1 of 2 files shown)


config.yaml
```diff
@@ -36,11 +36,6 @@ demo: &demo
           - # Test key #2, secret: fOhZOFr6SSnYv1/xfKPPMHdWLdQDboysJ4Q794Gezmw=
             - 3BAyX5pSiFerJgLsJZ9XyoBDMHPtq3NLqm5AGPcBpba6W644zSiSWy3vVr
             - 1000
-  witness:
-    db:
-      path: "/var/db/dscp-witness"
-      clean: false
-
     logging:
```
pasqu4le: I don't understand this change. This `logging` here should be a child of `witness` (see the educator component below), but with this it becomes a child of `core`.

config.yaml
```diff
@@ -70,6 +65,11 @@ demo: &demo
               log:
                 severity: Debug+
 
+  witness:
+    db:
+      path: "/var/db/dscp-witness"
```
mkaito: This needs to be overridden for production, so not much point in putting it here.

docs/config-full-sample.yaml
```diff
@@ -97,6 +97,7 @@ sample:
     db:
       # Path to the database folder
       path: "/var/db/dscp-witness"
+      clean: False
```
mkaito: So what does this do?

flyingleafe:
@mkaito @Heimdell
It seems to me that the problem has been resolved by other means for a while and therefore there's no point in keeping this PR.

PR #326, opened 2018-11-22
[DSCP-301] Add endpoint showing educator status

Martoon-00:
### Description

Add a way to fetch educator's info directly via Educator API.

### YT issue

https://issues.serokell.io/issue/DSCP-301

### Checklist

Stylistic (obligatory):

### DIFF
--- a/educator/src/Dscp/Educator/Web/Logic.hs
+++ b/educator/src/Dscp/Educator/Web/Logic.hs
@@ -2,11 +2,22 @@
 
 module Dscp.Educator.Web.Logic
     ( commonGetProofs
+    , getEducatorStatus
     ) where
 
+import Data.Default (def)
+import Dscp.Core
 import Dscp.DB.SQLite
+import Dscp.Educator.Launcher.Mode
+import Dscp.Educator.Web.Educator.Types
 import Dscp.Educator.Web.Types
+import Dscp.Resource.Keys
+import Dscp.Snowdrop.Mode
+import Dscp.Snowdrop.Types
 import Dscp.Util.Aeson
+import Dscp.Witness.Logic.Getters
+import Dscp.Witness.SDLock
+import Dscp.Witness.Web.Types
 
 commonGetProofs
     :: MonadEducatorWebQuery m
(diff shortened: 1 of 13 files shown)


educator/src/Dscp/Educator/Web/Logic.hs
```diff
@@ -19,3 +31,22 @@ commonGetProofs filters = do
...
+    -- TODO [DSCP-367]: to replace with one-liner
+    accounts <- readingSDLock $ do
+        blockAccount <- runSdReadM $ getAccountMaybe address
+        poolAccount  <- runSdReadM $ getMempoolAccountMaybe address
+        return $ fromMaybe def <$>
+                 BlocksOrMempool{ bmConfirmed = blockAccount, bmTotal = poolAccount }
+
+    return EducatorInfo
+        { eiAddress = address
+        , eiBalances = leftToPanic . coinFromInteger . aBalance <$> accounts
```
flyingleafe: Either use `unsafeMkCoin`, or throw an error inside a monad instead.
Another question is: why the heck `aBalance` in `Account` is not a `Coin` already?...

witness/src/Dscp/Witness/Web/Logic.hs
```diff
@@ -193,7 +193,7 @@ getHashType someHash = fmap (fromMaybe HashIsUnknown) . runMaybeT . asum $
     isAddress = is
         (const HashIsAddress)
         addrFromText
-        (runSdReadMLocked @'ChainAndMempool . getMempoolAccountMaybe)
+        (\x -> runSdReadMLocked @'ChainAndMempool $ getMempoolAccountMaybe x)
```
flyingleafe: Why?

witness/src/Dscp/Witness/Logic/Getters.hs
```diff
@@ -122,16 +122,16 @@ resolveNext = SD.queryOne . NextBlockOf . headerHash >=> pure . map unNextBlock
...
-    lift . runStateSdMLocked . SD.queryOne . AccountId
+    lift . runStateSdM . SD.queryOne . AccountId
 
 -- | Safely get an account taking mempool into consideration.
 getMempoolAccountMaybe
-    :: WitnessWorkMode ctx m
+    :: (WitnessWorkMode ctx m, WithinReadSDLock)
     => Address -> SdReadM 'ChainAndMempool m (Maybe Account)
-getMempoolAccountMaybe addr = readingSDLock $ liftSdM $ SD.queryOne (AccountId addr)
+getMempoolAccountMaybe addr = liftSdM $ SD.queryOne (AccountId addr)
```
flyingleafe: Do I understand correctly that every time those operations were called in existing code, the lock has been already taken, and therefore taking it again were obsolete?

educator/src/Dscp/Educator/Web/Logic.hs
Martoon-00: I guess we don't want `Account` to keep `Coin` because `< maxBound` checks are completely unnecessary in expansion and validation where `Account` is primarily used. Actually, this check is an extra here too, so `unsafeMkCoin` looks like a nice option.

Although, what do you think about this way of resolving the inconsistency (bounded in some places, and not bounded in others) in `Coin` semantics?
https://issues.serokell.io/issue/DSCP-363#comment=96-11517

witness/src/Dscp/Witness/Web/Logic.hs
Martoon-00: `(.)` is not able of passing `WithinReadSDLock` constraint from `runSdReadMLocked` to `getMempoolAccountMaybe` :pepe_sad:

witness/src/Dscp/Witness/Logic/Getters.hs
Martoon-00: Not that far ago every small operation took lock itself locally, including this one.
When I modified Explorer API recently, I mostly replaced local locking with endpoint-level locking, but yes, I missed this function, it took the reading lock again for nothing until this PR.

Martoon-00:
Ready for rereview.

PR #328, opened 2018-11-24
[DSCP-261] Validate block's slotId and difficulty

Martoon-00:
### Description

Adding missing checks required me to add one more type of transaction - block meta, which is responsible for block-level stuff expansion and validation. It is not a true transaction though, rather just a helper, so it is not signed, it does not appear in network messages and is not visible in explorer API.

### YT issue

https://issues.serokell.io/issue/DSCP-261

### Checklist

Stylistic (obligatory):

### DIFF
--- a/core/src/Dscp/Util.hs
+++ b/core/src/Dscp/Util.hs
@@ -12,7 +12,6 @@ module Dscp.Util
          -- * Exceptions processing
        , wrapRethrow
        , wrapRethrowIO
-       , onAnException
 
          -- * Error handling
        , assert
@@ -144,10 +143,6 @@ wrapRethrowIO
     :: (Exception e1, Exception e2, MonadCatch m, MonadIO m)
     => (e1 -> e2) -> IO a -> m a
 wrapRethrowIO wrap action = wrapRethrow wrap (liftIO action)
-
--- | Similar to 'onException', but provides exception itself.
-onAnException :: (MonadCatch m, Exception e) => m a -> (e -> m ()) -> m a
-onAnException action onExc = action `catch` \e -> onExc e >> throwM e
 
 -----------------------------------------------------------
 -- Do-or-throw error handlers
--- a/witness/src/Dscp/Snowdrop/Expanders.hs
+++ b/witness/src/Dscp/Snowdrop/Expanders.hs
@@ -10,6 +10,8 @@ import Data.Default (Default (..))
 import qualified Data.List as List
 import qualified Data.Map.Strict as Map
 import qualified Data.Set as Set
+import Fmt ((+|), (|+))
+import Serokell.Util (enumerate)
 
 import qualified Snowdrop.Block as SD
 import Snowdrop.Core (ChgAccum, ChgAccumCtx, ERoComp, Expander (..), SeqExpanders (..),
@@ -19,7 +21,8 @@ import Snowdrop.Util
 
 import Dscp.Core
 import Dscp.Crypto as Dscp
-import Dscp.Snowdrop.Configuration hiding (PublicationTxWitness)
+import Dscp.Snowdrop.AccountValidation
+import Dscp.Snowdrop.Configuration
 import qualified Dscp.Snowdrop.Configuration as Conf (Proofs (..))
 import Dscp.Snowdrop.Storage.Types
 import Dscp.Snowdrop.Types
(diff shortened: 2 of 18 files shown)


core/src/Dscp/Util.hs
```diff
@@ -146,6 +146,7 @@ wrapRethrowIO
 wrapRethrowIO wrap action = wrapRethrow wrap (liftIO action)
 
 -- | Similar to 'onException', but provides exception itself.
+-- TODO: remove in favor of 'UnliftIO.withException'?
```
flyingleafe: Yes, why not.

witness/src/Dscp/Snowdrop/Expanders.hs
```diff
@@ -38,10 +41,25 @@ expandBlock ::
...
+       , HasLens ctx (ChgAccumCtx ctx)
+       , HasCoreConfig
+       )
+    => BlockMetaTx -> ERoComp Exceptions Ids Values ctx SStateTx
+expandBlockMetaTx meta =
+    let stateTx = StateTxType $ getId (Proxy @TxIds) BlockMetaTxTypeId
+        proof = BlockMetaTxWitnessProof
+    in expandUnionRawTxs (\_ -> (stateTx, proof, seqExpandersBlockMetaTx)) [meta] >>= \case
+           [expanded] -> return expanded
+           _          -> error "expandBlockMetaTx: did not expand 1 meta into 1 sd tx"
```
flyingleafe: Doesn't SD provide a function for expanding _one_ transaction? This looks ugly.

witness/src/Dscp/Snowdrop/Expanders.hs
```diff
@@ -306,6 +325,94 @@ seqExpandersBalanceTx feesReceiverAddr (Fees minimalFees) =
...
+        let block@Block{ bHeader = header, bBody = body } = bmtBlock
+        let issuerAddr = mkAddr (hIssuer header)
+        let isGenesis = header == bHeader genesisBlock
+
+        {- Validation -}
+
+        when isGenesis $
+            -- further we assume that our genesis block has the least possible difficulty
+            unless (hDifficulty header == 0) $
+                error "Genesis block difficulty must be 0"
```
flyingleafe: And why should we just crash on this occasion? Maybe better to throw a decent exception instead?

witness/src/Dscp/Snowdrop/Expanders.hs
```diff
@@ -306,6 +325,94 @@ seqExpandersBalanceTx feesReceiverAddr (Fees minimalFees) =
...
+                (inj . NextBlockOf $ hPrevHash header)
+                (New . NextBlockOfVal . NextBlock $ headerHash header)
+
+        -- no keys collision is possible sinse we have validated transactions
+        -- before this function call, thus every transaction is unique
+        let txBlockRefsChange = Map.fromList $
+                enumerate (bbTxs (bBody block)) <&> \(idx, gTx) ->
+                    ( inj . TxBlockRefId . toGTxId . unGTxWitnessed $ gTx
+                    , New . TxBlockVal $ TxBlockRef (headerHash block) idx
+                    )
```
flyingleafe: Can we collect those changes in transactions expanders? E. g. by passing header hash to `expandGTxs` and further to `seqExpandersGTx`? I'm uncomfortable with the fact that this is the only reason why `BlockMetaTx` contains the whole `Block` and not only its header is this piece of code.

witness/src/Dscp/Snowdrop/Expanders.hs
Martoon-00: There is definitely no one :snowdrop:

witness/src/Dscp/Snowdrop/Expanders.hs
Martoon-00: Well, it's more about programmer's mistake than about handling normal-in-the-wild case, `error` seems to be more justified (?)

witness/src/Dscp/Snowdrop/Expanders.hs
Martoon-00: Good point :+1:

Martoon-00: Hmm, actually also I moved signature verification to the expander, and block body participates in it.
I also find carrying the whole block around ugly, but ways to make it less ugly doesn't seem to be worth the effort :thinking:

witness/src/Dscp/Snowdrop/Expanders.hs
flyingleafe: @Martoon-00 AFAIU this code is used when we are validating the block which comes from the _outside_, so it might be just anything. For me, it seems like this line of code makes any malformed genesis block with difficulty > 1 a remote killing spell for witness node.

witness/src/Dscp/Snowdrop/Expanders.hs
flyingleafe: @Martoon-00 Just found out that the block body is hashed _entirely_ instead of building a Merkle tree over it...
WTF is really going on, my god, I really need to dive into our public chain code, the more I look into it, the more ashamed and confused I become.

flyingleafe: And actually, there shouldn't be even a Merkle tree hash, there should be just a _final AVL root_ in the block header which should be signed, and this will already give us all the guarantees. Where is this final AVL hash stored in the block anyway? I don't see any.

flyingleafe: @Martoon-00 nevermind, this totally should not be addressed in scope of this PR.

witness/src/Dscp/Snowdrop/Expanders.hs
flyingleafe: Ah, well, indeed, the block header should be exactly the same as the genesis header from configuration. Well yes, then it's impossible to fire malformed genesis spells, having `error` here is fine.

witness/src/Dscp/Snowdrop/Expanders.hs
Martoon-00: Nice, I'll set a todo to keep just the header in `TxBlockMeta`

Martoon-00: in the corresponding YT ticket

PR #330, opened 2018-11-26
[DSCP-390] Mock time

Martoon-00:
### Description

For the sake of #328 get passing tests in presence of slot verification I had to mock slotting.
However, mocking the time entirely occurred to be even simpler in some sense.

### YT issue

https://issues.serokell.io/issue/DSCP-390

### Checklist

Stylistic (obligatory):

### DIFF
--- a/witness/tests/Test/Dscp/Witness/Explorer/ExplorerSpec.hs
+++ b/witness/tests/Test/Dscp/Witness/Explorer/ExplorerSpec.hs
@@ -1,9 +1,11 @@
 module Test.Dscp.Witness.Explorer.ExplorerSpec where
 
 import Data.Default (def)
+import Loot.Base.HasLens (lensOf)
 
 import Dscp.Core
 import Dscp.Crypto
+import Dscp.Resource.Keys
 import Dscp.Snowdrop.Configuration
 import Dscp.Snowdrop.Mode
 import Dscp.Snowdrop.Types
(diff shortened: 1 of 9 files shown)


witness/tests/Test/Dscp/Witness/Explorer/ExplorerSpec.hs
```diff
@@ -51,12 +53,16 @@ createAndSubmitPub genSecret = do
...
-    => SlotId -> m HeaderHash
-dumpBlock slotId = do
-    block <- createBlock runSdM slotId
-    void $ applyBlock block
-    return (headerHash block)
+    :: (TestWitnessWorkMode ctx m, WithinWriteSDLock)
+    => m HeaderHash
+dumpBlock = do
+    slotId <- rewindToNextSlot
+    let issuerKey = KeyResources . mkSecretKeyData $ testFindSlotOwner slotId
```
Heimdell: This operation is a bit uncontrollable. I don't think anything can happen so we might want to cancel this, though.

Martoon-00: Well, that's true. I'd like to come up with a better thing here, but with interface provided by our `witness <-> slot` mapping logic it's probably the only normal way.

Heimdell: No, I'm about `createBlock` after `sleep`.

Martoon-00: Lool, that wasn't obvious for me.
Okay, I'll just probably tell in the comment that it's part of `rewindToNextSlot`'s semantics to return _new_ slot. If we cannot guarantee this simple invariant for `rewindToNextSlot`, then the whole mocking doesn't look sound.

PR #331, opened 2018-11-26
[DSCP-367] Merge `runSdM` and `runStateSdM`

Martoon-00:
### Description

Previously we had separate processing for data stored in AVL and all other data.
They both were considered simultaneously only when necessary, i.e. in block application
and mempool modification.

Problem: this does not provide a reasonable level of abstraction, allowing for subtle bugs.

Solution: quick observation of blockchain-util's code has shown that composition of
actions on storages deliver an about-zero overhead (O(log (prefixes number))),
thus it makes sense to provide an interface which hides exact storage being accessed.

### YT issue

https://issues.serokell.io/issue/DSCP-367

### Checklist

Stylistic (obligatory):

### DIFF
--- a/witness/src/Dscp/Snowdrop/Actions.hs
+++ b/witness/src/Dscp/Snowdrop/Actions.hs
@@ -3,16 +3,19 @@
 module Dscp.Snowdrop.Actions
     ( SDVars
     , SDActionsM (..)
+    , AVLPlusBlockComposition
     , initSDActions
     , sdActionsComposition
     , initAccounts
     ) where
 
 import qualified Data.Map as Map
+import Data.Reflection (Reifies (..))
 import Data.Time.Clock (UTCTime (..))
 import UnliftIO (MonadUnliftIO)
 
 import qualified Snowdrop.Block as SD
+import qualified Snowdrop.Core as SD
 import qualified Snowdrop.Execution as SD (CompositeChgAccum, DbAccessActions (..),
                                            DbModifyActions (..), SumChangeSet,
                                            constructCompositeActions)
@@ -21,7 +24,7 @@ import qualified Snowdrop.Util as SD (ExecM, RIO (..))
 import Dscp.Core
 import Dscp.Crypto
 import qualified Dscp.DB.CanProvideDB as DB
-import Dscp.Snowdrop.Configuration (BlockPlusAVLComposition, Ids (..), Values (..))
+import Dscp.Snowdrop.Configuration (Ids (..), Values (..), accountPrefix)
 import Dscp.Snowdrop.Serialise ()
 import Dscp.Snowdrop.Storage.Avlp
 import Dscp.Snowdrop.Storage.PluginBased
@@ -36,6 +39,10 @@ deriving instance MonadUnliftIO (SD.RIO ctx)
 
 type SDDataActions  m = SD.DbModifyActions (AVLChgAccum AvlHash Ids Values) Ids Values m AvlProof
 type SDBlockActions m = SD.DbModifyActions (SD.SumChangeSet Ids Values) Ids Values m ()
+
+data AVLPlusBlockComposition
+instance Reifies AVLPlusBlockComposition (Set SD.Prefix) where
+    reflect _ = one accountPrefix
 
 -- Parameter m will be instantiated with RIO Context when the context is defined.
 data SDActionsM m = SDActionsM
@@ -80,14 +87,14 @@ sdActionsComposition
   => SDActionsM m
   -> SD.DbAccessActions
       (SD.CompositeChgAccum
-          (SD.SumChangeSet Ids Values)
           (AVLChgAccum AvlHash Ids Values)
-          BlockPlusAVLComposition)
+          (SD.SumChangeSet Ids Values)
+          AVLPlusBlockComposition)
       Ids
       Values
       m
 sdActionsComposition SDActionsM{..} =
     SD.constructCompositeActions
-        @BlockPlusAVLComposition
-        (SD.dmaAccessActions nsBlockDBActions)
+        @AVLPlusBlockComposition
         (SD.dmaAccessActions nsStateDBActions)
+        (SD.dmaAccessActions nsBlockDBActions)
--- a/witness/src/Dscp/Snowdrop/ReadMode.hs
+++ b/witness/src/Dscp/Snowdrop/ReadMode.hs
@@ -1,0 +1,53 @@
+-- | SD extended reading mode.
+
+module Dscp.Snowdrop.ReadMode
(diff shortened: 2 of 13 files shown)


witness/src/Dscp/Snowdrop/Actions.hs
```diff
@@ -37,6 +40,10 @@ deriving instance MonadUnliftIO (SD.RIO ctx)
 type SDDataActions  m = SD.DbModifyActions (AVLChgAccum AvlHash Ids Values) Ids Values m AvlProof
 type SDBlockActions m = SD.DbModifyActions (SD.SumChangeSet Ids Values) Ids Values m ()
 
+data AVLPlusBlockComposition
```
flyingleafe: How does it differ from `BlockPlusAVLComposition` imported from `Dscp.Snowdrop.Configuration`, except the order of the... arguments to `SD.CompositeChgAccum` in `sdActionsComposition`?
What's going on, generally?

witness/src/Dscp/Witness/Logic/Getters.hs
```diff
@@ -36,21 +37,33 @@ import Dscp.Witness.Launcher.Context
...
+    , bmTotal = runSdReadM @'ChainAndMempool action
+    }
 
 ----------------------------------------------------------------------------
 -- Block/Header/Tip getters
 ----------------------------------------------------------------------------
 
 -- | Retrieves current tip.
-getTipHash :: HasWitnessConfig => SdM_ chgacc HeaderHash
+getTipHash :: HasWitnessConfig => SdM HeaderHash
```
flyingleafe: How much sense does it make to restrict all those methods only to one possible `chgacc`?

witness/src/Dscp/Witness/Logic/Getters.hs
```diff
@@ -36,21 +37,33 @@ import Dscp.Witness.Launcher.Context
...
+-- | Run an action both in chain-only and chain+mempool modes.
+runSdDual
+    :: WitnessWorkMode ctx m
+    => (forall mode. KnownSdReadMode mode => SdReadM mode m a)
+    -> m (BlocksOrMempool a)
+runSdDual action =
+    sequence BlocksOrMempool
+    { bmConfirmed = runSdReadM @'ChainOnly action
+    , bmTotal = runSdReadM @'ChainAndMempool action
+    }
```
flyingleafe: Seems like a very bad helper which should never be used (as it would cause double query to blockchain)

Martoon-00: Mm yeah, it's probably worth removing this function.
Although doing two requests is what snowdrop assumes as to do at the moment.

witness/src/Dscp/Witness/Logic/Getters.hs
Martoon-00: It seems to me that now when we support both chain-only and chain+mempool modes, requirements to `chgacc` should not change anymore.
They were actually restricted to one `chgacc` yet just a couple of weeks ago (before mempool PR), if anyone ever needs polymorphic `chgacc` back, it's quick to return.

witness/src/Dscp/Snowdrop/Actions.hs
Martoon-00: I actually moved `BlockPlusAVLComposition` here and renamed it.
`Composition A B` isn't symmetric - we have to provide the set of prefixes `A` needs access to, all other prefixes are assumed to relate to `B`. The overhead of the composition scales with the logarithm of the size of that set. To decrease its size I swapped `A` and `B`, thus `BlockPlusAVLComposition` should've been also renamed.

witness/src/Dscp/Witness/Logic/Getters.hs
flyingleafe: @Martoon-00 you claim that we cannot query _just_ the mempool in Snowdrop and there is no built-in way of obtaining such a dual result via a single query?

Martoon-00: Well, AFAIS snowdrop's way of accessing data allows only one exact source of data (either only chain or chain with mempool) per `runERoComp` (similar to our `runSdM`) run. 
Need to ask Gosha about that.

Martoon-00: We discussed this with @Heimdell and came to conclusion, that this dual db access is probably ok due to caching, need benchmarks to judge about it. And snowdrop indeed does not allow to do one db access from the code point of view.

witness/src/Dscp/Snowdrop/ReadMode.hs
```diff
@@ -0,0 +1,53 @@
+-- | SD extended reading mode.
+
+module Dscp.Snowdrop.ReadMode
```
Heimdell: Why do we call it "mode", anyway?

witness/src/Dscp/Witness/Logic/Getters.hs
Heimdell: Confirming. At least different states will be in different AVL+ trees. Also, rocksdb has a very good cache and is cheap to read.

witness/src/Dscp/Witness/Logic/Getters.hs
```diff
@@ -112,26 +125,18 @@ resolvePrevious o = do
...
-
 -- | Safely get an account taking mempool into consideration.
-getMempoolAccountMaybe
-    :: WitnessWorkMode ctx m
-    => Address -> SdReadM 'ChainAndMempool m (Maybe Account)
-getMempoolAccountMaybe addr = readingSDLock $ liftSdM $ SD.queryOne (AccountId addr)
+getAccountMaybe
+    :: (WitnessWorkMode ctx m, KnownSdReadMode mode, WithinReadSDLock)
+    => Address -> SdReadM mode m (Maybe Account)
+getAccountMaybe addr = liftSdM $ SD.queryOne (AccountId addr)
```
Heimdell: So, there were "get account' and "get account from mempool", both doing `SD.queryOne`. What does the survivor do?

Martoon-00: Asks for an account in the mode requested by the caller of this method.

witness/src/Dscp/Snowdrop/ReadMode.hs
Martoon-00: It provides two modes for read-only actions: chain-only and chain+mempool. Do you have alternatives?

witness/src/Dscp/Witness/Logic/Getters.hs
flyingleafe: Cool then!

PR #335, opened 2018-12-07
[DSCP-406] Remove DBT

Martoon-00:
### Description

Use directly a monad which is instance of `MonadBeam`.
This allows us not to implement `MonadBeam` with undefined methods, not without
plenty of other workarounds though.

### YT issue

https://issues.serokell.io/issue/DSCP-406

### Checklist

Hint: a perfect PR has all the checkmarks set.

Possible related changes (conditional):
- [x] I added tests if required, in case they are:
  - Covering new introduced functionality (feature).
  - Regression tests preventing the bug from silently reappearing again (bugfix).
- [x] I have checked the [sample config](../tree/master/docs/config-full-sample.yaml) and [launch scripts](../tree/master/scripts/launch) and updated them if it was necessary (especially if executables or CLIs were changed).
- [x] I have checked READMEs and changed them accordingly if it's necessary (the main [README.md](../tree/master/README.md) as well as subproject READMEs for all related executables).
- [x] If a public API structure or/and data serialization was changed, I made sure that these changes are reflected in the:
  - [Swagger](../tree/master/specs/disciplina) API specs.
  - Any [related documentation](../tree/master/docs/api-types.md).
- [x] If any public documentation was written, I have spellchecked it.

Stylistic (obligatory):
- [x] My commit history is clean, descriptive and do not contain merge or revert commits or commits like "WIP".
- [x] My code complies with the [style guide](../tree/master/docs/code-style.md).
- [x] My code is documented (haddock) and these docs are decent (full enough and spellchecked).

PR #340, opened 2018-12-15
Revert "[DSCP-406] Remove DBT"

Martoon-00:
This is an ~exact revert of DisciplinaOU/disciplina#335.

Motivation:
* I need `instance MonadCatch` in database endpoints, and that's problematic to achieve in `Pg` monad which is `MonadBeam` implementation in Postgres case.
* `Pg`'s `>>=` does not execute SQL requests, rather collects them for execution in `Pg` runner. That's an inconvenient property for database endpoint monad, it complicates debugging.
* Previously mentioned reasons, like the fact that we carry context explicitly now (and we always carry it explicitly in tests).


flyingleafe:
@Martoon-00 so what, how are we supposed to deal with `MonadCatch` inside `PG` monad, if we use `DBT` wrapper?

Martoon-00:
The point is to lift `Pg` to `DBT m` on per `select/insert` basis rather than on db-endpoint basis and then use `MonadCatch` for `DBT m`. `DBT` will be still needed to carry `Connection` in which the transaction happens.

PR #348, opened 2018-12-19
[DSCP-364] Move validation to expander, part 1: accounts

Martoon-00:
### Description

Split account expansion into parts (corresponding functions) and perform all the validation there.
To the best of my knowledge, I performed only move and thoughtful remove operations on code, the logic behind this tx machinery should be exactly the same as it was before; after that several controversial points were uncovered which you can see in `TODO`s.

I'm not sure my way to organize the expansion code is really good, although it seems better than having 4-screens-height wall of code which it would otherwise become, I'm open to suggestions here.

Next part is ready where I move the logic of publication transactions, will publish it once this PR is merged.

### YT issue

https://issues.serokell.io/issue/DSCP-364

### Checklist

Stylistic (obligatory):


witness/tests/Test/Dscp/Witness/Tx/MoneyTx.hs
```diff
@@ -106,26 +106,28 @@ spec_Money_tx_application = do
...
+                applyTx tx
 
         it "Having no money is not fine" $ witnessProperty $ do
             txData' <- pick genSafeTxData
             secret <- pick $ arbitrary `suchThat` (`notElem` testGenesisSecrets)
             let txData = txData'{ tdSecret = secret }
 
             let tx = makeTx properSteps txData
-            lift $ throwsPrism (_AccountError . _BalanceCannotBecomeNegative) $
+            lift $ throwsPrism (_AccountError . _InsufficientBalance) $
```
Heimdell: What does `throwsPrism` do? Repacks exception with a prism? Or does it checks that only exception, covered by prism, is thrown? Why not `assertThrowsPrism`?

Heimdell: Also, is it possible to remove all these `lift`s? We have nice test logic here, its a pity that all this plumbing is sticking out of the wall.

Heimdell (commented):
Ugh.

Yeah, SD guys removed validators in their newest version, but
1) its questionable we ever migrate to it, because reasons
2) this removal actually has made things kinda worse.

At least I hope we have less code now. The combined expanders/validators are still scary and look like some legacy, however.

specs/disciplina/witness/api/witness.yaml
```diff
@@ -716,25 +716,20 @@ components:
...
-            - NotASingletonSelfUpdate
-            - NonceMustBeIncremented
-            - PaymentMustBePositive
-            - ReceiverMustIncreaseBalance
-            - SumMustBeNonNegative
+            - NonceMismatch
+            - CannotAffordOutputs
             - CannotAffordFees
-            - InsufficientBalance
+            - WitnessMismatchesInput
```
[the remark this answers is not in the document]
Martoon-00: I actually renamed `InsufficientBalance` into `CannotAffordOutputs`, because otherwise existing `CannotAffordFees` is confusing.

witness/src/Dscp/Snowdrop/Expanders/BlockMeta.hs
```diff
@@ -0,0 +1,122 @@
...
+        let totalChange = mconcat [diffChange, forwardBlockChange, txBlockRefsChange]
+        pure $ mkDiffCS (totalChange :: Map.Map Ids (ValueOp Values))
+  where
+    -- Account prefixes are used during the computation to access current balance
+    inP  = Set.fromList []
+    -- Expander returns account changes only
+    outP = Set.fromList [blockIdxPrefix, txBlockPrefix, nextBlockPrefix]
+
+    noPrevBlock (h :: HeaderHash) =
+        BlockMetaInternalError $ "Can't resolve previous block " +| h |+ ""
```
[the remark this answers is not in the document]
Martoon-00: > _portyankas_
:joy: 

Could you expand on your statement about AVL, how does replace need in this code?

This is exactly a movement of existing code actually, it checks all things related to block semantics which are not checked anywhere else.

witness/tests/Test/Dscp/Witness/Tx/MoneyTx.hs
Martoon-00: > Or does it checks that only exception, covered by prism, is thrown? 

Yep. Since the verb ends with `s`, I thought this function looks like a predicate.
I would rename, but this function has been used in too many places at now.


> Also, is it possible to remove all these lifts? We have nice test logic here, its a pity that all this plumbing is sticking out of the wall.

I tried really hard to avoid these `lift`s, and it looks like the best we can live with.

The approach you used in `sqliteProperty`, where arbitrarily-generated parameter was passed to a test case, scales not very well: I cannot use custom generators (like `genSafeTxData` above) without some half-working hacks, and writing preconditions (like with `pre` function) makes us use `if/then/else` with the corresponding indentation madness.

Thus I use `PropertyM` mostly everywhere for testing monadic code, but implementing `MonadUnliftIO` for such Cont-based monad seems impossible, so `lift`s.

PR #351, opened 2018-12-24
[DSCP-142] [DSCP-459] Add sorting to API

Martoon-00:
### Description

This PR consists of 3 parts:
1. `SortingParams` Servant API combinator which makes an endpoint optionally accept sorting specification and provide the corresponding object of `SortingSpec` type to an endpoint implementation. See `Dscp/Util/Servant.hs`.
2. `SortingSpec` mapping to an ordering thing understandable by Beam. See `Dscp/DB/SQLite/ApiUtil.hs`.
3. Introduce this all to Student API.

I'm not sure which kind of tests would be appropriate here, probably we'd better create a ticket to QA team? Or, we can wait till #353 is merged.

**For future work:**
* DSCP-425: allow to specify nulls-first / nulls-last order - but not earlier than we move to Postgres.
* Do a similar thing for filters, so that all filters would be passed as a single parameter (no more countless `sGetAssignments Nothing Nothing Nothing` in tests, no more need to define `EducatorGetAssugnmentsFilters` datatypes, generic filters (`==`, comparisons, `like` statements e.t.c)). DSCP-434.

**Potential problems:**
* Not clear how pagination will couple with sorting. Simple `offset/limit` pagination remains easy to implement, but we will probably want to use keyset (`nextId/limit`) pagination in most cases instead and here sorting breaks things unless special care is taken.

### YT issue

https://issues.serokell.io/issue/DSCP-142

### Checklist

Stylistic (obligatory):


yorickvP:
HList has a really old `Setup.lhs` file, that is not compatible with cabal 2. Stack ignores the `Setup.lhs` file and just runs `cabal configure`.

default.nix:
```nix
stackToNix {
# ...
overrides = self: super: {
  HList = pkgs.haskell.lib.overrideCabal super.HList (o: {
    preCompileBuildDriver = ''
      rm -f Setup.lhs Setup.hs
      cat << EOF > Setup.hs
      import Distribution.Simple
      main = defaultMain
      EOF
    '';
  });
};
}
```

Martoon-00:
Thanks!

Martoon-00:
Created DSCP-459 for extraction to a library.

> Can we do it in a more lightweight way?
:thinking: :thinking: :thinking: 
Megaparsec allows to easily achieve decent parsing errors, not saying about readable code. And actually, I counted on that some paring library will be used for other purposes as well sooner or later.
On the other side, if I extract this code to another library, it would be very weird to have such a dependency.
Maybe let me rewrite this code after I do DSCP-459 so that I don't litter Disciplina project with tests?

flyingleafe:
@Martoon-00 let's just do it then, and this PR will contain only the import of the lib and its application, together with removal of `Dscp.Util.Servant`

Martoon-00:
:tada: Celebrating the hoursewarming :tada:
https://github.com/serokell/servant-util

PR #353, opened 2018-12-28
[DSCP-432] Authentication for API's clients

Martoon-00:
### Description

- [x] Allows making authenticated requests to Student API.
- [x] Extracts `ClientApiError` datatype (used in all APIs) to a common location.
- [x] Adds first server-client tests.

### YT issue

https://issues.serokell.io/issue/DSCP-432

### Checklist

Stylistic (obligatory):


educator/src/Dscp/Educator/Web/Auth.hs
```diff
@@ -80,6 +94,130 @@ instance ApiHasArgClass (Auth' auths res) where
...
+
+runAuthClientM :: MakeAuthHeaders -> AuthClientM m a -> m a
+runAuthClientM ctx action = runReaderT (unAuthClientM action) ctx
+
+-- | Attaches authentication header to given request.
+runAuthRequest :: (Cli.Request -> m a) -> Cli.Request -> AuthClientM m a
+runAuthRequest runReq req =
+    AuthClientM . ReaderT $ \(MakeAuthHeaders makeAuthHeaders) -> do
+        let authHeaders = makeAuthHeaders req
+        let totalAuthHeader = BS.intercalate ", " $ map unAuthHeader authHeaders
```
flyingleafe: What is the semantics of it? AFAIK, passing a comma-separated list of different authentication tokens in not the valid or standard way of doing authentication; our custom server code does not support this too. So what is the idea behind this?

educator/src/Dscp/Educator/Web/Auth.hs
```diff
@@ -122,13 +264,48 @@ checkAuthBasic = do
...
+-- | Extract endpoint name from built request.
+clientRequestEndpoint :: Cli.Request -> Text
+clientRequestEndpoint = decodeUtf8 . toLazyByteString . Cli.requestPath
+
+-- | Make authentication signature.
+signRequestBasic :: SecretKey -> Cli.Request -> AuthToken
+signRequestBasic sk req =
+    let endpoint = decodeUtf8 . toLazyByteString $ Cli.requestPath req
+    in AuthToken $ "Bearer " <> makeTestAuthToken sk endpoint
+        ---- ^ TODO: provide not a test auth token, but that requires IO :(
```
flyingleafe: So what?...

educator/src/Dscp/Educator/Web/Auth.hs
Martoon-00: Some people say that it's fine to pass several authentication headers like this
https://stackoverflow.com/questions/29282578/multiple-http-authorization-headers.
Probably it should cause the server to accept the first working authentication method in the list.

Well, it actually doesn't seem canonical, so I'd better allow only one auth credentials.

educator/src/Dscp/Educator/Web/Auth.hs
Martoon-00: Right. Since this is going to be used in production, I will add some `unsafePerformIO`s.

flyingleafe: @Martoon-00 well, at first you can make `provideAuth` monadic and see how it goes. This would be a valid change anyway, because construction of auth token often requires something from the outside world (timestamp, random, whatever)

Martoon-00: Yeah, I want to try to propagate this `IO` addition to `instance HasServer`, maybe I won't even need unsafe things at all.

Martoon-00:
Addressed both comments, should look better now.

PR #358, opened 2019-01-08
[DSCP-415] Rewrite scripts on shelly, part 1

Martoon-00:
### Description

Rewrite bash script `node.sh`

### YT issue

https://issues.serokell.io/issue/DSCP-415

### Checklist

Stylistic (obligatory):


yorickvP:
Hi! It's sad, but stack-to-nix doesn't support custom snapshots right now. Looks like it would be possible, (add a loop in https://github.com/serokell/stack-to-nix/blob/master/build.nix#L14), but it would be some work.

As an alternative, maybe you can build the scripts from stack.yaml instead?

Looking at DSCP-415, it was aiming for code integration with disciplina, so we would have compile-time checks to see if it's used correctly. I see no value in just porting the shell script to haskell. It makes it harder to understand how this demo uses disciplina, as well, and doesn't prevent any breakage.

Martoon-00:
> As an alternative, maybe you can build the scripts from stack.yaml instead?

Yep, now I'd like to try harder and come to using project's configuration.

> Looking at DSCP-415, it was aiming for code integration with disciplina, so we would have compile-time checks to see if it's used correctly. I see no value in just porting the shell script to haskell. It makes it harder to understand how this demo uses disciplina, as well, and doesn't prevent any breakage.

That makes sense actually :sweat_smile:.
Moreover, using a custom snapshot as script resolver seems not to allow to depend on the codebase, so probably that's an entirely wrong approach.

PR #360, opened 2019-01-11
[DSCP-438] Checkcv-PDF endpoint

Heimdell:
### Description

- [x] `checkcv-pdf` endpoint
- [x] `JSON`-injector (as `ByteString` in `base64` form)
- [x] `JSON`-extractor (as `ByteString`)

### YT issue

https://issues.serokell.io/issue/DSCP-438

### Checklist

Stylistic (obligatory):
- [ ] My code complies with the [style guide](../tree/master/docs/code-style.md).
- [ ] My code is documented (haddock) and these docs are decent (full enough and spellchecked).

### DIFF
--- a/pdfs/app/Main.hs
+++ b/pdfs/app/Main.hs
@@ -1,0 +1,20 @@
+
+import Data.ByteString as LBS
+import Data.Text as Text
+
+import Pdf.Scanner
+
+main :: IO ()
+main = do
+    [pdf, json, out] <- getArgs
+
+    pdfText  <- LBS.readFile pdf
+    jsonText <- LBS.readFile json
+
+    LBS.length pdfText `seq`
+        case inject (MaxSearchLength (Just 2048)) jsonText (PDFBody pdfText) of
+            Nothing -> do
+                error . Text.pack $ "Cannot inject <" <> json <> "> into <" <> pdf <> ">."
+
+            Just (PDFBody it) -> do
+                LBS.writeFile out it
--- a/pdfs/app/Main.hs
+++ b/pdfs/app/Main.hs
@@ -6,15 +6,17 @@ import Pdf.Scanner
 
 main :: IO ()
 main = do
-    [pdf, json, out] <- getArgs
+    getArgs >>= \case
+        [pdf, json, out] -> do
+            pdfText  <- LBS.readFile pdf
+            jsonText <- LBS.readFile json
 
-    pdfText  <- LBS.readFile pdf
-    jsonText <- LBS.readFile json
+            LBS.length pdfText `seq`
+                case inject (MaxSearchLength (Just 2048)) jsonText (PDFBody pdfText) of
+                    Nothing -> do
+                        error . Text.pack $ "Cannot inject <" <> json <> "> into <" <> pdf <> ">."
 
-    LBS.length pdfText `seq`
-        case inject (MaxSearchLength (Just 2048)) jsonText (PDFBody pdfText) of
-            Nothing -> do
-                error . Text.pack $ "Cannot inject <" <> json <> "> into <" <> pdf <> ">."
-
-            Just (PDFBody it) -> do
-                LBS.writeFile out it
+                    Just (PDFBody it) -> do
+                        LBS.writeFile out it
+        _ -> do
+            error "USAGE: dscp-inject-json-into-pdf <pdf-in> <json> <pdf-out>"
--- a/pdfs/src/Pdf/Scanner.hs
+++ b/pdfs/src/Pdf/Scanner.hs
@@ -38,15 +38,15 @@ findFromEnd :: Int -> LBS.ByteString -> LBS.ByteString -> Maybe Int
 findFromEnd quota what text = do
     guard $ not $ LBS.null what
     guard $ not $ LBS.null text
-    go quota (LBS.length text - LBS.length what) id
+    go quota (LBS.length text - LBS.length what)
   where
-    go 0 _ k         = k Nothing
-    go _ i k | i < 0 = k Nothing
-    go q i k = do
+    go 0 _         = Nothing
+    go _ i | i < 0 = Nothing
+    go q i = do
         if LBS.index text i == LBS.head what
             && equalsFrom i what text
-        then k (Just i)
-        else go (q - 1) (i - 1) k
+        then Just i
+        else go (q - 1) (i - 1)
 
 equalsFrom :: Int -> LBS.ByteString -> LBS.ByteString -> Bool
 equalsFrom offset what text = do
--- a/pdfs/src/Pdf/Scanner.hs
+++ b/pdfs/src/Pdf/Scanner.hs
@@ -16,7 +16,7 @@ inject
     (PDFBody text)
   =
     let base64    = LBS.encode thing
-        commented = "\n" <> fairCVStartMark <> "{" <> base64 <> "}%"
+        commented = "\n%" <> fairCVStartMark <> "{" <> base64 <> "}%"
     in  do
         place <- findFromEnd quota insertionMark text
         return $ PDFBody $ insertAt place commented text
--- a/pdfs/src/Pdf/Scanner.hs
+++ b/pdfs/src/Pdf/Scanner.hs
@@ -1,5 +1,5 @@
 
-module Pdf.Scanner where
+module Pdf.Scanner (inject, project, PDFBody (..), MaxSearchLength (..)) where
 
 import Data.ByteString.Char8  as LBS
 import Data.ByteString.Base64 as LBS
--- a/pdfs/src/Pdf/Scanner.hs
+++ b/pdfs/src/Pdf/Scanner.hs
@@ -1,5 +1,5 @@
 
-module Pdf.Scanner (inject, project, PDFBody (..), MaxSearchLength (..)) where
+module Pdf.Scanner (inject, project, insertionMark, PDFBody (..), MaxSearchLength (..)) where
 
 import Data.ByteString.Char8  as LBS
 import Data.ByteString.Base64 as LBS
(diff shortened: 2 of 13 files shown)


pdfs/src/Pdf/Scanner.hs
```diff
@@ -0,0 +1,66 @@
...
+equalsFrom offset what text = do
+    what == LBS.take (LBS.length what) (LBS.drop offset text)
+
+insertAt :: Int -> LBS.ByteString -> LBS.ByteString -> LBS.ByteString
+insertAt place what text =
+    let (before, after) = LBS.splitAt place text
+    in   before <> what <> after
+
+insertionMark :: LBS.ByteString
+insertionMark = "\ntrailer"
```
Martoon-00: Funny, I opened my diplomas made in latex (ru and eng versions), and they both have no "trailer" section.

Heimdell: Can you grep for it?
I definitely can't put stuff before `xref` section, b/c there is somewhere a "count of bytes from start to `xref`" field. I can put comment right after xref, though.

Martoon-00: I tried to grep and failed. That's weird though, any pdf format tutorial seems to say that the trailer section must be present.

Heimdell: Its like `<html>` and `<body>` tags, I think. We can't really know how manglable PDF is, because its not the standard is used to show PDFs, but concrete PDF-clients.

pdfs/tests/Main.hs
```diff
@@ -0,0 +1,50 @@
...
+main :: IO ()
+main = hspec $ do
+    let inf  = MaxSearchLength Nothing
+        zero = MaxSearchLength (Just 0)
+
+    describe "injection/projection works" $ do
+        it "project _ x (inject _ x text) == x" . property $
+            \(before, after, piece) ->
+                let
+                    text = before <> insertionMark <> after
```
flyingleafe: I don't think there's anything wrong with having `let` definitions on the same line as `let` itself...

pdfs/src/Pdf/Scanner.hs
```diff
@@ -0,0 +1,66 @@
...
+
+newtype PDFBody = PDFBody { getPDFBody :: LBS.ByteString }
+
+newtype MaxSearchLength = MaxSearchLength { getMaxSearchLength :: Maybe Int }
+
+inject :: MaxSearchLength -> LBS.ByteString -> PDFBody -> Maybe PDFBody
+inject
+    (fromMaybe maxBound . getMaxSearchLength -> quota)
+     thing
+    (PDFBody text)
```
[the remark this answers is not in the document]
Heimdell: We have this thing the style guide, actually. Are we short on screen space (or paper to print the code)?

pdfs/tests/Main.hs
Heimdell: Yeah, its just looked as a wall of text in the moment.

pdfs/app/Main.hs
```diff
@@ -0,0 +1,20 @@
+
+import Data.ByteString as LBS
+import Data.Text as Text
+
+import Pdf.Scanner
+
+main :: IO ()
+main = do
+    [pdf, json, out] <- getArgs
```
Martoon-00: I think it's worth at least printing a help message if a wrong number of arguments is supplied.

witness/src/Dscp/Witness/Web/Logic.hs
```diff
@@ -234,3 +240,16 @@ checkFairCV =
...
+    let maybeFairCV = do
+            base64json    <- Pdf.project (Pdf.MaxSearchLength (Just 2048)) (Pdf.PDFBody pdf)
+            fairCVencoded <- toMaybe $ BS.decode base64json
+            Aeson.decode $ BS.fromStrict fairCVencoded
+
+    fairCV    <- maybe (throwM InvalidFormat) pure maybeFairCV
+    checkeRes <- checkFairCV fairCV
+    return (checkeRes, fairCV)
+  where
+    toMaybe = either (const Nothing) pure
```
Martoon-00: There is `rightToMaybe`

witness/src/Dscp/Witness/Web/Logic.hs
```diff
@@ -234,3 +240,16 @@ checkFairCV =
         { fairCVCheckResults = results
         , fairCVFullyValid = all and results
         }
+
+checkFairCVPDF :: forall ctx m. WitnessWorkMode ctx m => ByteString -> m (FairCVCheckResult, FairCV)
+checkFairCVPDF pdf = do
+    let maybeFairCV = do
+            base64json    <- Pdf.project (Pdf.MaxSearchLength (Just 2048)) (Pdf.PDFBody pdf)
+            fairCVencoded <- toMaybe $ BS.decode base64json
```
Martoon-00: Minor: `BS.decode` is not very telling name for `Base??` :/
(same for `Scanner` module).

Also, `project` already seems to apply `LBS.decode`, is it fine?

witness/src/Dscp/Witness/Web/API.hs
```diff
@@ -82,6 +82,11 @@ data WitnessEndpoints route = WitnessEndpoints
         :- "checkcv"
         :> ReqBody '[JSON] FairCV
         :> Verb 'PUT 200 '[DSON] FairCVCheckResult
+
+    , wCheckFairCVPDF :: route
+        :- "checkcv"
+        :> ReqBody '[OctetStream] ByteString
+        :> Verb 'PUT 200 '[DSON] (FairCVCheckResult, FairCV)
```
Martoon-00: This response would look like a JSON-list from frontend's point of view, and it's supposed canonical to always return JSON-objects. Better create a datatype for this thing (we do so in every API).

witness/src/Dscp/Witness/Web/Types.hs
```diff
@@ -107,6 +108,9 @@ data HashIs
 deriving instance (Eq (Id a), Eq a) => Eq (PaginatedList d a)
 deriving instance (Show (Id a), Show a) => Show (PaginatedList d a)
 
+instance Buildable ByteString where
+    build = (build :: Text -> Builder) . decodeUtf8 . Base64.encode
```
Martoon-00: Maybe we do not always want to see them in base64? I think there is a reason why `instance Buildable ByteString` is not defined in the original library. 
Your data extracted from PDF seems to have a quite specific format, would be nice to see a newtype carried around.

witness/src/Dscp/Witness/Web/Logic.hs
Heimdell: No, it isn't. Thanks for notifying, you probably save me some debug time.

witness/src/Dscp/Witness/Web/Types.hs
Heimdell: Actually, that's the PDF itself. I'm thinking about `build _ = "<binary-data>"`, if the output of this ever goes to any log.

Martoon-00: Ah, right.
We really do log it since any call of an endpoint is entirely logged.
I can imagine which trash our logs will turn into if we print the entire PDF :smile:

witness/src/Dscp/Witness/Web/Types.hs
```diff
@@ -107,6 +113,9 @@ data HashIs
 deriving instance (Eq (Id a), Eq a) => Eq (PaginatedList d a)
 deriving instance (Show (Id a), Show a) => Show (PaginatedList d a)
 
+instance Buildable ByteString where
+    build _ = "<binary data, omitted>"
+
```
Martoon-00: I mean, I still think it's important to create a dedicated newtype and use it `checkFairCVPDF` (maybe even `Pdf.PDFBody` itself).

witness/src/Dscp/Witness/Web/Logic.hs
```diff
@@ -234,3 +239,13 @@ checkFairCV =
...
         , fairCVFullyValid = all and results
         }
+
+checkFairCVPDF :: forall ctx m. WitnessWorkMode ctx m => ByteString -> m FairCVAndCheckResult
+checkFairCVPDF pdf = do
+    let maybeFairCV = do
+            fairCVencoded <- Pdf.project (Pdf.MaxSearchLength (Just 2048)) (Pdf.PDFBody pdf)
+            Aeson.decode $ BS.fromStrict fairCVencoded
+
+    fairCV   <- maybe (throwM InvalidFormat) pure maybeFairCV
```
Martoon-00: One more minor thing: if I use this endpoint and get this `InvalidFormat` which usually stands for wrong request body or query parameters format, that would be misleading because I did everything correctly and it's PDF which is wrong (and it is made not by me but by some another endpoint).
Maybe creating a separate exception would be better, something like "invalid certificate pdf format".

specs/disciplina/witness/api/witness.yaml
```diff
@@ -368,6 +368,22 @@ paths:
...
+        200:
+          description: >-
+            FairCV in a PDF was well-formed and has been successfully checked. The
+            result of checking is returned in a response along with FairCV itself.
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/FairCVAndCheckResult'
+      requestBody:
+        $ref: '#/components/requestBodies/ByteString'
```
flyingleafe: You forgot to add the request body specification to `#/components/requestBodies`, which makes the spec invalid (editor.swagger.io will complain if you try to upload it). Also, this is not generic `ByteString`, it should be called `PDF`, because request content type should be `application/pdf`, not `application/octet-stream`

Heimdell: > `#/components/requestBodies`

Is this a local address in the same yaml document?

flyingleafe: @Heimdell yes, the section where request bodies are defined is literally a couple of lines lower.

PR #361, opened 2019-01-11
[DSCP-431] use student API in faircv tool

pasqu4le:
### Description

This PR aims at using the student API in the tool that produces faircvs, to better test these interactions.

The WIP label has been removed because this PR also contains the fixes for DSCP-450, the issue tracking the problem that this PR itself revealed.

### YT issue

https://issues.serokell.io/issue/DSCP-431

### Checklist

Stylistic (obligatory):


scripts/launch/node.sh
```diff
@@ -139,25 +139,25 @@ fi
...
     if [ -z "$(which dscp-witness)" ]; then
-        stack exec "dscp-witness" -- $common_params $witness_params
+        stack exec "dscp-witness" --nix -- $common_params $witness_params
     else
         dscp-witness $common_params $witness_params
     fi
 elif [[ "$node" == "faucet" ]]; then
     if [ -z "$(which dscp-faucet)" ]; then
-        stack exec "dscp-faucet" -- $common_params $faucet_params
+        stack exec "dscp-faucet" --nix -- $common_params $faucet_params
```
Martoon-00: Please do not do this. Not everyone uses nix.
I believe that this can be set somehow globally on your machine, maybe in `.stack/global-project/stack.yaml`.

tools/src/faircv/Main.hs
```diff
@@ -76,15 +76,22 @@ main = do
...
     assLst <- sGetAssignments sClient Nothing Nothing Nothing False
     -- Make a new submission from the first one, then send it
-    mapM_ (sendSubmission sClient . infoToNewSub) $ take 1 assLst
+    subInfos <- mapM (sendSubmission sClient . infoToNewSub) $ take 1 assLst
 
-    -- Wait some time for the proofs to be available, then get them and make a faircv
+    -- Wait for the proofs to be available, then get them and make a faircv
     logInfo "Submissions sent, waiting for proofs"
-    threadDelay $ sec 20
+    waitForProofs subInfos
```
Martoon-00: :cool:

tools/src/faircv/Options.hs
```diff
@@ -0,0 +1,68 @@
...
+       ) where
+
+import Options.Applicative (Parser, auto, execParser, fullDesc, help, helper,
+                            info, long, metavar, option, progDesc, short,
+                            showDefault, strOption, value)
+import Time.Rational (RatioNat)
+
+
+data FaircvOptions = FaircvOptions
+    { witnessUrl    :: String
```
Martoon-00: What do you think about the following:
Make this parameter optional. In `node.sh` and `demo.sh` make `witness-listen` and `educator-listen` parameters have the same value - this will effectively make educator node serve both witness and educator APIs on the same port. And from this executable, you request `educatorUrl` when `witnessUrl` is unspecified.

tools/src/faircv/Main.hs
```diff
@@ -1,26 +1,31 @@
...
+
+-- This is a copy of dscp-keygen's parseInputWithSecret (for SecretFromSeed)
+secretFromSeed :: ByteString -> Maybe SecretKey
+secretFromSeed input = asum [
+      do
+        sd <- readMaybe @Word . toString @Text $ decodeUtf8 input
+        return $ withIntSeed (fromIntegral sd) genSecretKey
+    , do
+        return $ withSeed input genSecretKey
+    ]
```
Martoon-00: Would be very nice to move this (and the original function) somewhere to `Dscp.Crypto.?` with a note "this is how we generate secret keys from user-supplied seed".
I did this once, but then my change occurred to be redundant.

tools/src/faircv/Options.hs
```diff
@@ -0,0 +1,68 @@
+module Options
```
Martoon-00: Try not to use such general name, `stack ghci` goes mad when it sees two modules with the same name (except for `Main`).

scripts/launch/node.sh
pasqu4le: sorry that was not intended, I changed it for a quick run and forgot about it, I'll remove it

tools/src/faircv/Options.hs
pasqu4le: I liked the idea, but it seems we cannot do that. I just tried and it seems that communications using the `StudentApiClient` result all in a 404.

tools/src/faircv/Options.hs
pasqu4le: Done, I also renamed the Client module for the same reason.

tools/src/faircv/Main.hs
pasqu4le: Done, I moved the function in `Dscp.Crypto.Util` which seemed a fair place.

tools/src/faircv/Options.hs
Martoon-00: Weird, I have run the executable like `stack exec dscp-faircv localhost:8091 localhost:8090 ./run/tmp/educator.key` and that was working.
But ok, that's not significant.

tools/src/faircv/Main.hs
Martoon-00: Thanks :+1:

tools/src/faircv/Options.hs
pasqu4le: well, but `8090` and `8091` are different, maybe I didn't understand your suggestion

Martoon-00: Generally, you need no cluster of witnesses, having a single educator (which is also a witness) is enough. So you can replace `8091` with `8090` [here](https://github.com/DisciplinaOU/disciplina/blob/8032e15454ab099030dfc9d83c6e8e2c1f6b1be1/scripts/launch/node.sh#L54), and launch

```
> ./scripts/launch/node.sh e b
> stack exec dscp-faircv -- -w localhost:8090 -e localhost:8090 -k ./run/tmp/educator.key
``` 
Now `-w` and `-e` accept the same value and it works

Although you may have reasons to launch separate witnesses especially since that discovered a bug.

pasqu4le: Oh ok, I understood you then, but I have tried with a cluster (unsuccessfully).
I'll give it another try later.

educator/tests/Test/Dscp/Educator/Publishing.hs
```diff
@@ -15,8 +15,10 @@ import Test.Dscp.DB.SQL.Mode
 import Test.Dscp.Educator.Mode
 import Test.Dscp.Educator.Web.Scenarios
 
-spec_Private_blocks_publishing :: Spec
-spec_Private_blocks_publishing = specWithTempPostgresServer $ do
+-- | This was renamed from "spec_Private_blocks_publishing" to be ignored.
+-- Both the tests it contains do not pass anymore and a solution is not known yet.
```
Martoon-00: Please mention `DSCP-464` ticket here (I've just created it) so that we remember where this issue should be resolved.

tools/README.md
```diff
@@ -127,3 +127,27 @@ echo 456 \
...
+- `--educator-url` (or `-e`) to use the student API
+- `--key-file` (or `-k`)     to generate the educator keys
+
+There are other options, they can be listed by using `dscp-faircv --help`
+
+#### Example on a local cluster
+```bash
+dscp-faircv -w 127.0.0.1:8013 -e 127.0.0.1:8090 -k run/tmp/educator/educator.key
+```
+
```
Martoon-00: Cool :+1:

tools/src/faircv/Main.hs
Martoon-00: IMO it may worth waiting for a little bit more (fractions of seconds) after proofs are found, since your ask educator for proofs, but then verify FairCV in witness. It may take a little while for the educator to send its block to the witness.

tools/src/faircv/Main.hs
```diff
@@ -35,74 +31,70 @@ instance MonadLogging IO where
...
+    fcv <- unReadyFairCV <$> leftToFail (mergeFairCVList fcvrs)
+
+    -- Write the result in the specified file and show the validation result
+    logInfo "Writing result to file"
+    writeFile outputFile $ decodeUtf8 $ encode fcv
+
+    checkFairCV wClient fcv >>= \res ->
+        if fairCVFullyValid res
+        then logInfo "FairCV was successfully validated"
+        else logError "FairCV was not validated successfully"
```
Martoon-00: Now it looks like a serious executable :+1:

educator/src/Dscp/Educator/Logic/Publishing.hs
```diff
@@ -46,7 +46,7 @@ publishMissingBlocks =
         let pubTxs = createPublicationTxw feePolicy sk <$> blocks
 
         logInfo . fmt $ nameF "Adding publications to mempool" (listF $ toPtxId . ptwTx <$> pubTxs)
-        forM_ pubTxs $ addTxToMempool . GPublicationTxWitnessed
+        forM_ pubTxs $ relayTx . GPublicationTxWitnessed
```
Martoon-00: Actually, I'm still wrong here. It should be 
```
forM_ pubTxs $ \pubTx -> relayTx (GPublicationTxWitnessed pubTx) >>= wait @"tx in mempool"
```

Otherwise, since `updateMempoolWithPublications` is invoked every second, on high load when transactions which are put to the relay does not appear in mempool immediately, we will effectively fill the relay with many duplicates of our publications.

@pasqu4le do you mind if I make this change and force-push?

shell-native.nix
```diff
@@ -7,6 +7,7 @@ stdenv.mkDerivation rec {
     binutils
     haskell.compiler.ghc822
     pkgconfig
+    postgresql
```
Martoon-00: Sorry for this, I have seen this line in `default.nix` and was sure that was enough.

educator/src/Dscp/Educator/Logic/Publishing.hs
pasqu4le: @Martoon-00 no problem, please do.
Just please check that everything stay signed, because for some reason I seem to loose some signatures when rebasing or cherry-picking.
You could also add another commit on top of this, perhaps

shell-native.nix
pasqu4le: It's ok, also even with this `stack --nix` has some problems with tests (issue [DSCP-462](https://issues.serokell.io/issue/DSCP-462)).

P.s. I saw your comment on YouTrack, but haven't tried yet.

tools/src/faircv/FaircvOptions.hs
```diff
@@ -0,0 +1,99 @@
...
+outputFileParser = strOption $
+    long "file-output" <>
+    short 'f' <>
+    value "fairCV-example.json" <>
+    showDefault <>
+    metavar "PATH" <>
+    help "File that will contain the resulting FairCV"
+
+contentSeedParser :: Parser Text
+contentSeedParser = option auto $
```
Martoon-00: This is wrong btw, `auto` means that `read @Text` is used, and it expects strings like `"something"`.
I had to pass them like `-c "\"something\""` from shell.

I will fix this.

flyingleafe (approved):

educator/src/Dscp/Educator/Logic/Publishing.hs
flyingleafe: @Martoon-00 why didn't you proceed with this change?

PR #371, opened 2019-01-25
[DSCP-454] Certificate store

Martoon-00:
### Description

Reviewing on per-commit basis is recommended. 

With current approach where we keep certificate meta as JSON and `ToJSON`/`FromJSON` are generated automatically, access to fields of meta becomes fragile, I assume that we will cover that with tests to prevent breaks.
I do not add tests here though because that would be way easier and better to test endpoints instead.

### YT issue

https://issues.serokell.io/issue/DSCP-454

### Checklist

Stylistic (obligatory):


educator/database/schema.sql
```diff
@@ -161,3 +165,22 @@ create table if not exists block_txs (
...
+    hash  BYTEA  not null,
+    meta  JSONB  not null,
+    pdf   BYTEA  not null,
+
+    primary key (hash),
+
+);
+
+create index if not exists certificates_student_name
+    on certificates ((meta ->> 'studentName'));
```
flyingleafe: If we'd like to implement searching by diploma number too, we would also like to have `meta ->> 'number'` index. However, this is minor, since we'll do filtering on frontend anyway, at least initially.

educator/src/Dscp/Educator/DB/Instances.hs
```diff
@@ -48,6 +49,14 @@ instance FromField (TYPE) where \
     fromField field ty = \
         leftToPanic . fromByteArray @(TYPE) @ByteString <$> fromField field ty
 
+#define JsonInstanceEnc(TYPE) \
+instance IsPgValue (TYPE) where \
+    sqlValueSyntax = sqlValueSyntax . Aeson.encode
+
+#define JsonInstanceDec(TYPE) \
+instance FromField (TYPE) where \
+    fromField field ty = leftToPanic . Aeson.eitherDecode <$> fromField field ty
```
flyingleafe: Not sure that creating a macro for a single use case is a good thing to do, but whatever.

educator/src/Dscp/Educator/DB/Schema.hs
```diff
@@ -80,6 +80,12 @@ data BlockRowT f = BlockRow
     , brMerkleTree   :: C f (EmptyMerkleTree PrivateTx)
     } deriving (Generic)
 
+data CertificateRowT f = CertificateRow
+    { crHash :: C f (Hash CertificateMeta)
+    , crMeta :: C f CertificateMeta
+    , crPdf  :: C f Text  -- ^ TODO: Change it to whatever is used as PDF type
```
flyingleafe: It should be `LByteString`

educator/src/Dscp/DB/SQL/Util/Singleton.hs
```diff
@@ -0,0 +1,84 @@
...
+import Dscp.Util
+
+-- | Stores a single value.
+data SingletonT a (f :: * -> *) = SingletonUnsafe
+    { singletonId   :: Beam.C f Text
+      -- ^ Primary key, Beam cannot live without it.
+      -- Also it allows to easily extend the table to keep multiple elements.
+    , singletonItem :: Beam.C f a
+      -- ^ The stored item itself.
+    } deriving (Generic)
```
flyingleafe: Do I understand correctly that it is done in order to support JSON migrations for certificates? Not quite sure how exactly it would be used if the serialisation version is not included into JSON (using `Versioned`, for example`), and if we do store certificates in `Versioned` form, then I don't quite understand why do we need a global singleton value of currently supported version in a database (not in the code).

Martoon-00: Yes, this is for migrations. AFAIU we have to choices:
1. Attach a version to each certificate JSON, and migrate lazily on `GET` requests (otherwise, all versions would be always the same - the latest).
2. Keep the global version of all certificate JSONs, and migrate upon the node launch.

If we want to ever support certificate search then the first option becomes tricky: for instance, in a new version we add a new field and a search on this field, how will we perform this search if the field does not yet exist in old versions of certificate JSON?
Thus 2nd looks like the better option to me.

flyingleafe: @Martoon-00 Okay, so that means that we also need to have some constant version in the code, which represents the certificate version used by the code and should be matched with the value in DB on startup. Also, the migration logic itself should be provided somehow. When do you suggest to introduce all this?

Martoon-00:
Let it be https://issues.serokell.io/issue/DSCP-477

PR #374, opened 2019-01-30
[DSCP-468] Set timeout on resources teardown

Martoon-00:
### Description

This is the first part of the mentioned ticket where I'm trying to make our application
tolerant to freeze of its components.

In particular, all resources cleanup is now limited in time.
On timeout we print an error, and abandon that cleanup sending an exception to it.

**Why this relates to DSCP-468.**

How actually happens the bug with the node freezing on Postgres authentication failure:
* Networking resources are allocated;
* Attempt to allocate SQL resources is performed, including connections to the database
  and schema preparation (and all these using `componentm` library);
* Allocation fails (in particular, connection to the database fails) and error is properly
  thrown. Nothing is printed yet, because `componentm` tries to destroy all already
  allocated resources before printing full statistics on resources allocation & teardown
  and exceptions;
* Thus networking resources are destroyed, and here the node hangs.

This fix makes errors appear even if resource teardown code is unreliable (i.e. may hang).

### YT issue

https://issues.serokell.io/issue/DSCP-468

### Checklist

Stylistic (obligatory):

### DIFF
--- a/core/src/Dscp/Resource/Class.hs
+++ b/core/src/Dscp/Resource/Class.hs
@@ -14,11 +14,13 @@ module Dscp.Resource.Class
 
 import Control.Lens (makeLenses)
 import Control.Monad.Component (ComponentM, buildComponent)
-import Loot.Base.HasLens (HasLens (..))
+import Loot.Base.HasLens (HasLens (..), HasLens')
 import Loot.Log.Rio (LoggingIO)
+import Time (sec)
 
 import Dscp.Resource.Logging (LoggingParamsRec)
 import Dscp.Rio (RIO, runRIO)
+import Dscp.Util.TimeLimit
 
 -- | Contains parameters required for most of resources allocation.
 data InitParams = InitParams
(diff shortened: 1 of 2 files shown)


Martoon-00:
@kirelagin if freezes still bother you, you can rebase upon this PR once it is merged to master (and maybe even assist in that if you want).

core/src/Dscp/Resource/Class.hs
```diff
@@ -46,10 +48,15 @@ class AllocResource resource where
...
+    :: (HasLens' r LoggingIO)
+    => Text
     -> RIO r a
     -> (a -> RIO r ())
     -> ReaderT r ComponentM a
 buildComponentR desc allocate release =
     ReaderT $
-      \r -> buildComponent desc (runRIO r allocate) (runRIO r . release)
+      \r -> buildComponent desc
+                (runRIO r allocate)
```
flyingleafe: The original problem is the problem with DB authentication, which should arise during `allocate` - since `allocate` for Postgres includes schema preparation. It seems that if `boundTeardown` should be put around something, it more likely around `allocate`, not `release`.

kirelagin: > DB authentication is handled fine, attempt to use non-existing user or incorrect password produces an error at `openPostgreSQL`.

But it never showed up in logs 🤔. As @flyingleafe, I’m not really sure how this PR fixes the issue. The behaviour I saw was that on failed authentication no error was reported by educator and it froze. Does it report the error with this fix applied?
I guess it would be really helpful if commits we make contained a description of the _cause_ of the original issue and how it is addressed in the commit, e.g. in this particular case how adding a timeout on _cleanup_ helps get the error reported.

kirelagin: Oh, ok, I made it to the email in my inbox that has your YT comment and now everything is clear (I added my comment there). However, I’d like to reiterate that what you wrote in that YT comment should really have been in the commit message. (cc @gromakovsky)

Martoon-00: Right, this makes sense.

Martoon-00:
Thanks :tada: 

> On the other hand, one might say that this is called "code robustness"

Yeah, I see it in this quality, otherwise this PR is indeed a piece of trash.

PR #376, opened 2019-02-04
[DSCP-471] More specific `Arbitrary` instance for `CertificateMeta`

flyingleafe:
### Description

Changes `Arbitrary` instance so that dates make sense and text fields do not contain incomprehensible characters. Also, removes `DefPageSize` from pagination parameters.

### YT issue

https://issues.serokell.io/issue/DSCP-471

### Checklist

Stylistic (obligatory):

### DIFF
--- a/core/src/Dscp/Core/Arbitrary.hs
+++ b/core/src/Dscp/Core/Arbitrary.hs
@@ -41,7 +41,7 @@ module Dscp.Core.Arbitrary
 
 import qualified Data.Foldable
 import qualified Data.Text.Buildable
-import Data.Time.Calendar (toGregorian)
+import Data.Time.Calendar (fromGregorian, toGregorian)
 import Data.Time.Clock (UTCTime, getCurrentTime)
 import Fmt ((+||), (||+))
 import qualified GHC.Exts as Exts
@@ -132,9 +132,12 @@ instance Arbitrary CertificateMeta where
         cmEducationForm <- arbitrary
         -- Document number should be large enough to resemble
         -- actual diploma numbers
-        cmNumber <- arbitrary `suchThat` (> 1000000)
+        cmNumber <- (+ 1000000) <$> arbitrary
         -- Document should be issued in the year when studies are ended
-        cmIssueDate <- arbitrary `suchThat` (\issueDate -> getYear issueDate == cmEndYear)
+        let setYear year date =
+                let (_, m, d) = toGregorian date
+                in fromGregorian (toInteger year) m d
+        cmIssueDate <- setYear cmEndYear <$> arbitrary
         cmTitle <- arbitrary
         cmMajor <- arbitrary
         cmSpecialization <- arbitrary
--- a/core/src/Dscp/Core/Arbitrary.hs
+++ b/core/src/Dscp/Core/Arbitrary.hs
@@ -41,6 +41,7 @@ module Dscp.Core.Arbitrary
 
 import qualified Data.Foldable
 import qualified Data.Text.Buildable
+import Data.Time.Calendar (toGregorian)
 import Data.Time.Clock (UTCTime, getCurrentTime)
 import Fmt ((+||), (||+))
 import qualified GHC.Exts as Exts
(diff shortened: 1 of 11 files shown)


core/src/Dscp/Core/Arbitrary.hs
```diff
@@ -246,7 +265,8 @@ instance ArbitraryMixture BlockBody
 instance ArbitraryMixture Block
 
 instance Arbitrary ItemDesc where
-    arbitrary = arbitrary `suchThatMap` (rightToMaybe . toItemDesc)
+    arbitrary = (toBase64 @ByteString <$> arbitrary)
```
Martoon-00: I don't like it, tests should cover any possible scenarious and it would be better to create a separate generator for such smart thing.
I cannot image a case when this change causes us to miss a bug though.

flyingleafe: Well, type `ItemDesc` assumes that its contents should be somewhat readable, and set of `base64` encoded strings is still quite wide. 
Basically, this change might cause us to miss a bug if we processed the contents of `ItemDesc`s in some way which is more complex than taking a hash of it and putting it into the database, which we do not and I cannot imagine a plausible situation when we will.

core/src/Dscp/Core/Arbitrary.hs
```diff
@@ -118,8 +120,25 @@ instance Arbitrary EducationForm where
...
+            studentBirthYear = getYear cmStudentBirthDate
+        cmStartYear <- arbitrary `suchThat` (> studentBirthYear)
+        -- End year should be after start year
+        cmEndYear <- arbitrary `suchThat` (> cmStartYear)
+        cmEducationForm <- arbitrary
+        -- Document number should be large enough to resemble
+        -- actual diploma numbers
+        cmNumber <- arbitrary `suchThat` (> 1000000)
+        -- Document should be issued in the year when studies are ended
+        cmIssueDate <- arbitrary `suchThat` (\issueDate -> getYear issueDate == cmEndYear)
```
Martoon-00: Kek, requests to the endpoint hang. And something in this line makes me guess a suspect.

flyingleafe: @Martoon-00 resolved. But the removal of `DefPageSize` required some fixes in `servant-util`, check out: https://github.com/serokell/servant-util/pull/3

pdfs/src/Pdf/FromLatex.hs
```diff
@@ -35,20 +37,27 @@ import Dscp.Core.Foundation.Educator
 
 import Pdf.MkLatex
 
--- | Generate latex certificate from locate and data.
-generate :: Language -> CertificateFullInfo -> Text
-generate lang cert = do
-    Text.toStrict $ Text.toLazyText $ make (lang, cert)
+-- | Datatype containing information about Educator which is
+-- required in order to render a certificate.
+data EducatorInfo = EducatorInfo
```
Martoon-00: Oh kek, we already have one `EducatorInfo` [here](https://github.com/DisciplinaOU/disciplina/blob/martoon%2Fdscp457-bump/educator/src/Dscp/Educator/Web/Educator/Types.hs#L80).
Could you rename that to `EducatorStatusInfo` (to comply with endpoint name)?

shell-native.nix
```diff
@@ -18,7 +18,30 @@ stdenv.mkDerivation rec {
...
+        realscripts
+        datetime2
+        tracklang
+        etoolbox
+        everypage
+        xkeyval
+        xcolor
+        pgf
+        fontspec;
+    })
```
Martoon-00: Life is harsh

Martoon-00 (approved):
Welcome back to programming )

pdfs/src/Pdf/FromLatex.hs
flyingleafe: Which endpoint name, exactly?....
Better to make it `CertificateIssuerInfo` then, frankly.

Martoon-00: `api/educator/v1/status` endpoint, but do as you see fit.

PR #384, opened 2019-02-11
[DSCP-445] Authentication for certificate generator

pasqu4le:
### Description

This is a PR that modifies the certificate generator to use an AAA microservice.

### YT issue

https://issues.serokell.io/issue/DSCP-445

### Checklist

Stylistic (obligatory):


Martoon-00:
Could you explain in detail why you needed to use `AuthProtect`, and `Auth'` from `Dscp.Educator.Web.Auth` failed to suit purposes of the new authentication?
AFAIU `Auth'` thing is quite general (e.g. you are not tied to a specific JWT token format) and potentially supports multiple authentication methods.

(I'm worried about the fact we are going to use two different authentication "frameworks". But if that's the only way, then we can change the existing one someday).

flyingleafe (commented):
I have a couple of comments.

pasqu4le:
@Martoon-00 you were right, `Auth'` is more than general enough to be used in this case as well.
@flyingleafe I think I have addressed your comments as well (except the passphrase one that I don't know what to do for yet).

multi-educator/src/Dscp/MultiEducator/Launcher/Mode.hs
```diff
@@ -111,35 +111,36 @@ lookupEducator login = do
...
     let MultiEducatorResources{ _merEducatorData }
             = mctx ^. lensOf @MultiEducatorResources
-    fmap (\(EducatorContexts ctxs) -> M.lookup login ctxs) . atomically $ readTVar _merEducatorData
-
-loadEducator :: (MultiEducatorWorkMode ctx m) => Bool -> Text -> Maybe PassPhrase -> m Bool
-loadEducator _new login mpassphrase = do
+    EducatorContexts ctxs <- atomically $ readTVar _merEducatorData
+    case M.lookup login ctxs of
+        Just res -> return $ Just res
+        Nothing -> Just <$> loadEducator login Nothing
```
Martoon-00: It looks suspicious that we never write to `_merEducatorData`. Could you fix this now as well?

Martoon-00:
AFAIS unloading educator context is not handled here (and I doubt that it should be right now), don't you mind creating a YT ticket alongside DSCP-445 for that? And maybe leave your thoughts on this if you have any.

multi-educator/src/Dscp/MultiEducator/Launcher/Mode.hs
pasqu4le: Done, I forgot about this.

Anyway this couple of functions (especially `loadEducator`) should be modified, but I don't know how much it is necessary for this issue.
There are a couple of changes that could probably be done already (e.g. `lookupEducator` does not have to return a `Maybe` anymore), but I have not done those yet because I'm firstly trying to understand what exactly I should do here.

pasqu4le:
@Martoon-00 `_merEducatorData` allocation is badly described in `Dscp.MultiEducator.Launcher.Resource`, wouldn't it be enough to replace it and properly use `buildComponent` to make sure the teardown is handled correctly?

Martoon-00:
Oh, then this should be handled too )  (wherever you find easier).
What I meant is that the user may want to logout and then we need to close all connections to the database and unload the remaining educator context while multi-educator node continues its execution. But how to implement logout functionality is still vague afaiu.

pasqu4le:
Oh yes, right. Logout is still a bit of a mistery, but I'll create the issue later on, I want to put some thought into it so it can be detailed.

multi-educator/src/Dscp/MultiEducator/Launcher/Mode.hs
pasqu4le: @Martoon-00 I just remembered that this was not in `lookupEducator` because it is already handled at the end of `loadEducator`.

I suppose I could remove the `lookup` check here and revert the last commit.

Martoon-00: Oh lol, it's using `lensOf` to access contexts `TVar`, my bad.

pasqu4le: No problem, I searched `_merEducatorData` as well and though I forgot about the update, it's probably better if I make this visible as well.

multi-educator/src/Dscp/MultiEducator/Web/Educator/Auth.hs
```diff
@@ -37,21 +41,35 @@ deriveJSON defaultOptions ''EducatorAuthData
...
-    (JWTData login) <- jwtAuthCheck jwtSettings
-    liftIO (lookupEducator login) >>= \case
-        Just (key, db) -> return $ EducatorData key db
-        Nothing -> return $ EducatorNotFound
-
--}
+-- | Checks the signature of the JWT and returns an 'EducatorAuthData'
+multiEducatorAuthCheck :: MultiEducatorPublicKey -> AuthCheck EducatorAuthData
+multiEducatorAuthCheck (MultiEducatorPublicKey mpk) = do
+    (pk, payload) <- checkJWitness
```
flyingleafe: This only checks the JWS signature, but does not perform other checks typical for JWT, such as checking for expiry in `exp` claim. I really think that at least expiry check should be used. However, it's better to consult TMP team on which claims do they use for their own authorization processing, just to double-check.

flyingleafe (changes requested):
Found a bug.
@pasqu4le Regarding logout: you may temporarily forget about it. I don't think that a proper logout can be implemented without asking the AAA microservice directly "is this JWT still valid" during auth check. However, if AAA service sets up sufficiently small expiry times and if frontend processes logging out well, this should not be a huge problem.

multi-educator/src/Dscp/MultiEducator/Web/Educator/Auth.hs
```diff
@@ -37,21 +41,35 @@ deriveJSON defaultOptions ''EducatorAuthData
...
+    educatorAuthData <- maybe mempty pure $ decodeStrict payload
+    -- check that the JWS public key is the same one we have
+    guard (pk == mpk)
```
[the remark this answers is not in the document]
pasqu4le: Whoops, removed.

pasqu4le:
Update: Something was wrong and I ended up making a JWT to try the API's auth protection and I found that the JWS I made (with its `SecretKey`) were not passing the check (using its `PublicKey`).

Good news is: I figured the error was in the fact that I used (and read) the `publicKey` option as a `JWK` and not a directly `PublicKey`. This has now been fixed and the problem above is solved.
Bad news is: for some reason the provided AAA public key seems to be invalid and cannot be used as a `PublicKey`. I'll have to solve this starting tomorrow.

pasqu4le:
@flyingleafe I have looked into what is left to do and right now here's what is left:

- unloading educator context
  possibly needs some kind of logout and also now has it's own issue: DSCP-478
- there is a new TODO I left about probably needing a passphrase in `lookupEducator`
  you have commented on this already, there doesn't seem to be a valid thing to do yet
  maybe an issue should be opened?
- bot functionality is not implemented
  I tried to do this myself hoping that the reason it wasn't there was the lack of a DB in resources. Apparently there are missing constraints in `MultiEducatorWorkMode` necessary to use `initializeBot` (`HasLens' ctx SQL, HasLens' ctx (KeyResources EducatorNode), W.WitnessWorkMode ctx m`). To have these working `educatorEntry` needs to be modified to do what `withEducatorBackground` does and I'm not sure how to do that or if I can. I'd suggest a new issue for this as well.

Apart from these 3 things I can only try to fill as much of the PR checklist as possible.
P.s. should I squash all these commits?

flyingleafe (approved):
Let's merge that.

flyingleafe:
@pasqu4le
1) Thanks for creating an issue.
2) We'll not think about passphrases for a while, let's leave it as is.
3) Bot functionality for multi-educator is really not necessary.

pasqu4le:
@flyingleafe even though the bot functionality is not needed or used, I'd like to bring to your attention that the code regarding it is a bit funky.

What I mean is that `createStudentCheckAction` will ignore the bot config and behave as if it was active, the resulting `StudentCheckAction` will end up in the context.
On the other end `mkStudentApiServer'` will too ignore the bot config but behave as if it was off, and of course the result is part of the student API server.

Now I suppose this was written like this because they are respectively the easier behaviors to implement without a DB and I'm not sure this creates any problems right now, but it is probably something worth mentioning.

PR #385, opened 2019-02-11
[DSCP-475] Implement certificate generation logic

Martoon-00:
### Description

According to the plan in DSCP-475 description.
Some questionable moments are marked as `error`, I need a top-down decision to resolve them. \

I needed to make a bit of changes in `disciplina-pdfs` package to make PDF certificate generation and PDF injection operate with the same type of `ByteString` (made both lazy).

### YT issue

https://issues.serokell.io/issue/DSCP-475

### Checklist

Stylistic (obligatory):

### DIFF
--- a/core/src/Dscp/Core/FairCV.hs
+++ b/core/src/Dscp/Core/FairCV.hs
@@ -13,6 +13,7 @@ module Dscp.Core.FairCV
        , singletonFCV
        , mergeFairCVs
        , addProof
+       , privateBlockToFairCV
 
          -- * Fair CV check result
        , FairCVCheckResult (..)
@@ -99,6 +100,18 @@ addProof
 addProof educatorAddr blkHash proof fcv@(FairCV sAddr sName _) =
     mergeFairCVs fcv $
     singletonFCV sAddr sName educatorAddr blkHash proof
+
+-- | Make a FairCV from one private block.
+privateBlockToFairCV
+    :: PrivateBlockHeader
+    -> NonEmpty PrivateTx
+    -> Address
+    -> (Address, Text)
+    -> FairCVReady
+privateBlockToFairCV blkHeader txs educator (student, studentName) =
+    let blkHash = hash blkHeader
+        proof = readyProof $ merkleProofFromList txs
+    in singletonFCV student studentName educator blkHash proof
 
 ---------------------------------------------------------------------------
 -- Fair CV check result
--- a/core/src/Dscp/Core/Foundation/Educator.hs
+++ b/core/src/Dscp/Core/Foundation/Educator.hs
@@ -52,8 +52,6 @@ module Dscp.Core.Foundation.Educator
     , swSig
     , ssSubmission
     , ssWitness
-    , certificateStudentNameField
-    , certificateIssueDateField
 
     -- * Private transactions
     , PrivateTx (..)
@@ -345,22 +343,6 @@ data CertificateMeta = CertificateMeta
     , cmMajor            :: !ItemDesc
     , cmSpecialization   :: !(Maybe ItemDesc)
     } deriving (Show, Eq, Generic)
-
--- | Student name as it appears in JSON.
-certificateStudentNameField :: Text
-certificateStudentNameField = "studentName"
-
--- | Certificate issue date as it appears in JSON.
-certificateIssueDateField :: Text
-certificateIssueDateField = "issueDate"
-
--- TODO: I guess within DSCP-467 I will need to define
--- `instance {To,From}JSON CertificateMeta` via
--- https://github.com/agrafix/highjson, this will help us to
---
--- 1. Assign custom field names while preseving overall JSON instance derivation easy.
--- Then we can reuse those names.
--- 2. Provide a swagger description for each field.
 
 data EducationForm = Fulltime | Parttime | Fullpart
     deriving (Show, Eq, Generic, Enum, Bounded)
--- a/core/src/Dscp/Core/Foundation/Educator.hs
+++ b/core/src/Dscp/Core/Foundation/Educator.hs
@@ -31,6 +31,7 @@ module Dscp.Core.Foundation.Educator
     , CertificateFullInfo (..)
     , CertificateMeta (..)
     , CertificateGrade (..)
+    , CertificateIssuerInfo (..)
     , Certificate (..)
     , StudentInfo (..)
     , GradeInfo (..)
@@ -51,6 +52,8 @@ module Dscp.Core.Foundation.Educator
     , swSig
     , ssSubmission
     , ssWitness
+    , certificateStudentNameField
+    , certificateIssueDateField
 
     -- * Private transactions
     , PrivateTx (..)
@@ -343,6 +346,22 @@ data CertificateMeta = CertificateMeta
     , cmSpecialization   :: !(Maybe ItemDesc)
     } deriving (Show, Eq, Generic)
 
+-- | Student name as it appears in JSON.
+certificateStudentNameField :: Text
+certificateStudentNameField = "studentName"
+
+-- | Certificate issue date as it appears in JSON.
+certificateIssueDateField :: Text
+certificateIssueDateField = "issueDate"
+
+-- TODO: I guess within DSCP-467 I will need to define
+-- `instance {To,From}JSON CertificateMeta` via
+-- https://github.com/agrafix/highjson, this will help us to
+--
+-- 1. Assign custom field names while preseving overall JSON instance derivation easy.
+-- Then we can reuse those names.
+-- 2. Provide a swagger description for each field.
+
 data EducationForm = Fulltime | Parttime | Fullpart
     deriving (Show, Eq, Generic, Enum, Bounded)
 
@@ -357,8 +376,8 @@ data Certificate = Certificate
 data CertificateGrade = CertificateGrade
     { cgSubject :: ItemDesc
     , cgLang    :: Language
-    , cgHours   :: Int
-    , cgCredits :: Maybe Int
+    , cgHours   :: Word32
+    , cgCredits :: Maybe Word32
     , cgGrade   :: Grade
     } deriving (Show, Eq, Generic)
 
@@ -369,7 +388,7 @@ data Language = EN | RU
 -- datatype represents a request body for 'AddCertificate' endpoint.
 data CertificateFullInfo = CertificateFullInfo
     { cfiMeta   :: CertificateMeta
-    , cfiGrades :: [CertificateGrade]
+    , cfiGrades :: NonEmpty CertificateGrade
     } deriving (Show, Eq, Generic)
 
 data StudentInfo = StudentInfo
@@ -382,6 +401,13 @@ data GradeInfo = GradeInfo
     , giTimestamp      :: Timestamp
     , giHasProof       :: Bool
     } deriving (Show, Eq, Ord, Generic)
+
+-- | Datatype containing information about Educator which issued
+-- the certificate, required in order to render a certificate.
+data CertificateIssuerInfo = CertificateIssuerInfo
+    { ciiName :: ItemDesc
+    , ciiUrl  :: ItemDesc
+    } deriving (Show, Eq, Generic)
 
 instance Buildable (StudentInfo) where
     build (StudentInfo{..}) =
--- a/educator/src/Dscp/DB/SQL/Util/Json.hs
+++ b/educator/src/Dscp/DB/SQL/Util/Json.hs
@@ -1,0 +1,174 @@
+{- | Provides means of safe JSON fields access in a database.
+
+Primarily intended for product datatypes with automatically derived
+'Aeson.ToJSON' and 'Aeson.FromJSON' instances which appear in Postgres database
+schema.
+-}
+
+module Dscp.DB.SQL.Util.Json
+    ( JsonFieldType
+    , jsonFieldName
+    , HasSqlCastableJson
+    , (->$.)
+    , (->>$.)
+    ) where
+
+import qualified Data.Aeson as Aeson hiding (defaultOptions)
+import qualified Data.Aeson.Options as Aeson
+import Data.Time.Calendar (Day)
+import Database.Beam.Migrate (HasDefaultSqlDataType (..))
+import Database.Beam.Postgres (IsPgJSON, (->$), (->>$))
+import Database.Beam.Postgres.Syntax (PgDataTypeSyntax, PgExpressionSyntax)
+import Database.Beam.Query (QGenExpr, val_)
+import qualified GHC.Generics as G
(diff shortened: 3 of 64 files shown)


core/src/Dscp/Core/Foundation/Educator.hs
```diff
@@ -471,6 +472,14 @@ makeLenses ''PrivateTx
 makeLenses ''PrivateTxWitness
 makeLenses ''PrivateTxAux
 
+certGradeToPrivateTx :: Timestamp -> CertificateGrade -> PrivateTx
+certGradeToPrivateTx time CertificateGrade{..} =
+    PrivateTx
+    { _ptSignedSubmission = error "Mmm, what goes here???"
```
flyingleafe: As written in https://www.notion.so/serokell/Verifiable-PDF-certificates-6b49c78665134366934429be962f9880:
>As those certificates would not consist of actual submissions generated and sent to Educator by actual students, but rather of the submissions constructed by the certificate creator, all the submission signatures and student's address for such FairCV should be produced via dedicated secret key (e. g. consisting only from zeros/ones).

So, just hardcode some Student's private key and contents/assignments hashes for submissions. Another option is to use Educator's key to make signatures and call such grades "self-signed". However, this make it slightly harder for the users of certificate format (e.g. frontend) to distinguish between certificates produced with and without student's signatures.

educator/src/Dscp/Educator/DB/Queries.hs
```diff
@@ -354,6 +357,25 @@ createPrivateBlock delta = runMaybeT $ do
...
+    -> DBT 'WithinTx m (Maybe PrivateBlockHeader)
+dumpNonChainedTransactions atgDelta = do
+    txs <- getAllNonChainedTransactions
+    createPrivateBlock txs atgDelta
+
+createPrivateBlockFromCert
+    :: (DBM m, HasCtx ctx m '[KeyResources EducatorNode])
+    => CertificateFullInfo -> DBT 'WithinTx m (Maybe PrivateBlock)
+createPrivateBlockFromCert cert = do
+    time <- toTimestamp <$> liftIO getCurrentTime
```
flyingleafe: Semantically, grade timestamp should represent the time when the grade has been issued. Maybe it's better to use `issueDate` here (like, pretending that grades has appeared at 00:00 on the day of diploma issue)?

core/src/Dscp/Core/FairCV.hs
```diff
@@ -100,6 +102,16 @@ addProof educatorAddr blkHash proof fcv@(FairCV sAddr sName _) =
     mergeFairCVs fcv $
     singletonFCV sAddr sName educatorAddr blkHash proof
 
+-- | Make a FairCV from one private block.
+privateBlockToFairCV :: PrivateBlock -> Address -> (Address, Text) -> FairCVReady
+privateBlockToFairCV blk educator (student, studentName) =
+    let txs = blk ^. pbBody . pbbTxs
+        blkHash = hash (_pbHeader blk)
+        proof = mkMerkleProof (fromList txs) (fromList $ zipWith const [1..] txs)
```
flyingleafe: You would already build a Merkle tree from transactions list in `createPrivateBlock` at this point, and here you would do it again, while it's really unnecessary - because you create a proof for a _full_ tree, so it does not have any intermediate hashes, you only need to construct a bunch of empty `ProofBranch` nodes. 
This is not essential, but it would be nice to have such a function which creates a full `MerkleProof` from list. It will also get rid of `Maybe` and this ugly `error`. Making such a function will probably require generalizing `Dscp.Crypto.MerkleTree.nodeFromList` a bit.

educator/src/Dscp/Educator/Web/Educator/Queries.hs
```diff
@@ -248,3 +253,21 @@ educatorPostGrade subH grade = do
...
+educatorAddCertificate
+    :: MonadEducatorWeb ctx m
+    => CertificateFullInfo -> m ()
+educatorAddCertificate cert = do
+    mblk <- transact $ createPrivateBlockFromCert cert
+    -- This private block will be added to the public chain on itself shortly
+
+    eAddr <- ourAddress @EducatorNode
+    let blk = mblk ?: error "What if there were no grades provided?"
+    let sAddr = error "Where do we get it? Shouldn't it be in CertificateMeta?"
```
flyingleafe: As I noticed in the previous comment, we cannot meaningfully assign an address to a student if a certificate is created solely by Educator, so we should use some hardcoded address (or Educator's address, or something similar which makes it possible to distinguish between genuine Student's signatures and "fake" ones provided by Educator)

educator/src/Dscp/Educator/Web/Educator/Queries.hs
```diff
@@ -248,3 +253,21 @@ educatorPostGrade subH grade = do
...
+
+educatorAddCertificate
+    :: MonadEducatorWeb ctx m
+    => CertificateFullInfo -> m ()
+educatorAddCertificate cert = do
+    mblk <- transact $ createPrivateBlockFromCert cert
+    -- This private block will be added to the public chain on itself shortly
+
+    eAddr <- ourAddress @EducatorNode
+    let blk = mblk ?: error "What if there were no grades provided?"
```
flyingleafe: Good point. Diploma without grades is something absurd, so I guess we should actually make `cfiGrades :: NonEmpty CertificateGrade` (and also note in the endpoint description that empty list of grades will result in error).

pdfs/inject/Main.hs
```diff
@@ -1,6 +1,7 @@
 
-import qualified Data.ByteString as LBS
-import qualified Data.Text       as Text
+import qualified Data.ByteString as BS
+import qualified Data.ByteString.Lazy as LBS
```
flyingleafe: Haha lol, how did this happen :man_facepalming:

pdfs/src/Pdf/FromLatex.hs
```diff
@@ -146,7 +148,16 @@ produce loc ciInfo info (ResourcePath resources) =
             _ <- action
             _ <- action
 
-            LBS.readFile "texput.pdf"
+            pdf <- LBS.readFile "texput.pdf"
+            return $ PDFBody pdf
```
flyingleafe: Could just use `PDFBody <$>` =\

pdfs/src/Pdf/Scanner.hs
```diff
@@ -1,28 +1,32 @@
...
+    def = MaxSearchLength (Just 2048)
+
+inject :: MaxSearchLength -> ByteString -> PDFBody -> Maybe PDFBody
 inject
     (fromMaybe maxBound . getMaxSearchLength -> quota)
      thing
     (PDFBody text)
   =
-    let base64    = LBS.encode thing
+    let base64    = LBS.fromStrict $ Base64.encode thing
```
flyingleafe: I must have been very tired (or exceptionally lazy) when I reviewed the original code :man_facepalming:

witness/src/Dscp/Witness/Web/Logic.hs
```diff
@@ -241,10 +241,10 @@ checkFairCV =
         }
 
 checkFairCVPDF :: forall ctx m. WitnessWorkMode ctx m => ByteString -> m FairCVAndCheckResult
-checkFairCVPDF pdf = do
+checkFairCVPDF (LBS.fromStrict -> pdf) = do
```
flyingleafe: Frankly, why the endpoint accepts strict `ByteString` in the first place? :thinking:

witness/src/Dscp/Witness/Web/Logic.hs
```diff
@@ -241,10 +241,10 @@ checkFairCV =
...
     let maybeFairCV = do
             fairCVencoded <- Pdf.project (Pdf.MaxSearchLength (Just 2048)) (Pdf.PDFBody pdf)
-            Aeson.decode $ BS.fromStrict fairCVencoded
+            Aeson.decode $ LBS.fromStrict fairCVencoded
```
flyingleafe: And here we can use `Aeson.decodeStrict`

educator/src/Dscp/Educator/DB/Queries.hs
Martoon-00: Okay, this makes sense.

witness/src/Dscp/Witness/Web/Logic.hs
Martoon-00: Made endpoint accept `LByteString`

witness/src/Dscp/Witness/Web/Logic.hs
Martoon-00: Right, nice catch

core/src/Dscp/Core/FairCV.hs
Martoon-00: > You would already build a Merkle tree from transactions list in createPrivateBlock at this point, and here you would do it again, while it's really unnecessary - because you create a proof for a full tree, so it does not have any intermediate hashes, you only need to construct a bunch of empty ProofBranch nodes.

So instead I would better create a `MerkleProof` from transactions list directly?
Under empty `ProofBranch` nodes you mean ones with transactions, not just hashes?

> This is not essential, but it would be nice to have such a function which creates a full MerkleProof from list. It will also get rid of Maybe and this ugly error. Making such a function will probably require generalizing `Dscp.Crypto.MerkleTree.nodeFromList` a bit.

You mean, generalizing in a way which allows creating both `MerkleNode`s and `MerkleProof`s?

educator/src/Dscp/Educator/Web/Educator/Queries.hs
Martoon-00: Doing it, although I find it a bit weird, since if an educator knows a student name then he probably should know the student address as well.
And student address is the only viable identifier; would be funny to produce a certificate which applies to multiple people.

flyingleafe: @Martoon-00 well, there are _no_ real student's addresses in our use case; does ITMO know your Disciplina address? You don't even have one.

The certificate generator is for Educators willing to provide verifiable versions of diplomas for existing students. Those students do not have a Disciplina address, and Educators definitely do not currently want to ask their students to obtain a Disciplina address and sign those grades. They could generate the Disciplina keys for students themselves and send those keys to the students together with certificates, but this kinda misses the point of a _secret_ key - moreover, it's possible for some student to receive two different secret keys from two different educators this way.

So it's better to not require a distinguished secret key for every student at all for this usecase, but instead make it possible to easily distinguish between a FairCV created "properly" - via Student App and containing signed submissions from a student - and a FairCV created "improperly" - solely by Educator via certificate generator.

Martoon-00: Thanks, now it makes sense to me.

educator/src/Dscp/Educator/DB/Queries.hs
Martoon-00: I would probably set to 12:00, or guys from London will see grades being issued a day before the certificate issue date (which is fine, but to avoid inconsistency and extra questions in future...)

core/src/Dscp/Core/FairCV.hs
flyingleafe: >So instead I would better create a `MerkleProof` from transactions list directly?
Under empty `ProofBranch` nodes you mean ones with transactions, not just hashes?

`MerkleProof` has `ProofLeaf`s, which contain actual values of a list, and `ProofPruned`s, which contain only hashes. If a Merkle proof includes all the items in the original array, there are no `ProofPruned`s, only `ProofLeaf`s and `ProofBranch`es. And yes, it's easy to create such a proof from an original array, you won't even need to hash anything.

>You mean, generalizing in a way which allows creating both `MerkleNode`s and `MerkleProof`s?

Yes

pdfs/src/Pdf/FromLatex.hs
```diff
@@ -146,7 +149,15 @@ produce loc ciInfo info (ResourcePath resources) =
...
-            LBS.readFile "texput.pdf"
+            PDFBody <$> LBS.readFile "texput.pdf"
+
+-- | Generate a PDF-certificate with the default template and return it as
+-- a bytestring.
+compileCertificate
+    :: MonadIO m
+    => Language -> CertificateIssuerInfo -> CertificateFullInfo -> m PDFBody
+compileCertificate loc ciInfo info =
+    liftIO $ produce loc ciInfo info (ResourcePath "pdfs/template")
```
flyingleafe: This depends on the current process directory, and it makes it painful to run the node without errors - e. g. the test for certificate building already fails in CI complaining that resources directory cannot be found. I think that 
1) this path should be configurable via config so that it could be either absolute or relative to `AppDir`
2) we should make sure that resource files are included into the deployment distribution (I'm sure they are not now)

educator/tests/Test/Dscp/Educator/Certificates.hs
```diff
@@ -0,0 +1,10 @@
+module Test.Dscp.Educator.Certificates where
+
+import Dscp.Educator
+import Dscp.Util.Test
+
+spec_Educator_certificates :: Spec
+spec_Educator_certificates = do
+    it "Can build a full student certificate flawlessly" . property $
+        \cert faircv -> ioProperty $
+            total <$> buildCertificatePDF cert faircv
```
flyingleafe: This test fails on CI because template files are not found, see another comment.

pdfs/src/Pdf/FromLatex.hs
Martoon-00: Going to spend another day trying to make nix digest our tests :unamused:

Martoon-00:
What remains for the next week:
* Complete a test on sorting.
* Test endpoints manually one more time (I have observed some discrepancies, need to check are they normal or that's a bug).

In general, this is ready for one more review.

Martoon-00:
I guess I'm done here, would be nice to get a rereview.

educator/src/Dscp/DB/SQL/Util/Json.hs
```diff
@@ -0,0 +1,172 @@
...
+
+{- Note on optional fields:
+
+We do not actually care about whether 'Aeson.omitNothingFields' option is set
+or not, because Postgres treats "null" fields and absent fields equally - they
+are SQL nulls.
+
+For instance, this is good for our sorting, because fields which are "Maybe"
+can be justifiably sorted with nulls-first/nulls-last option specified.
+-}
```
flyingleafe: Better to move it to module description, so it appears in Haddock.

Martoon-00:
:heart:

educator/src/Dscp/DB/SQL/Util/Json.hs
Martoon-00: This comment is more about an implementation detail - whether to consider `omitNothingFields` Aeson option. I can move it next to the related typeclass method implementation.

PR #389, opened 2019-02-18
[DSCP-434] API filtering

Martoon-00:
### Description

Introduces new filtering to Student API where the user is allowed to specify different filtering
operations (`neq`, `in`, `cmp`, `like`). More about it [here](https://github.com/serokell/servant-util/tree/master/servant-util#filtering).

This also provides a neat interface for carrying and manually specifying filtering
specifications, no more need in specific datatypes per endpoint to gather all filters.

Updated swagger specification: [click](http://editor.swagger.io?url=https://gist.githubusercontent.com/Martoon-00/477c04e3d9ee9e2a910aef380ed4f381/raw/16735869b27165531d0df69a993a0a6445695d44/swagger.json).

I have renamed `timestamp` filtering parameter to `createdAt`; it's not used at frontends (educator or student) so no changed were required.

Auto swagger generation thing became general enough to be moved to `servant-util`, don't get scared of huge deleted pieces of code.

### YT issue

https://issues.serokell.io/issue/DSCP-434

### Checklist

Stylistic (obligatory):


educator/src/Dscp/Educator/Web/Types.hs
```diff
@@ -204,3 +206,40 @@ deriving instance FromHttpApiData IsGraded
...
+
+type instance SupportedFilters IsEnrolled = SupportedFilters Bool
+type instance SupportedFilters Course = IdentifierFilterTypes
+type instance SupportedFilters (Hash a) = SupportedFilters ByteString
+type instance SupportedFilters Address = DatetimeFilterTypes
+type instance SupportedFilters Timestamp = DatetimeFilterTypes
+type instance SupportedFilters ItemDesc = TextFilterTypes
+
+type instance FilteringParamTypesOf StudentInfo =
+    '[ "course" ?: 'ManualFilter Course ]
```
pasqu4le: Shouldn't this have an `isEnrolled` filter as well?

Martoon-00: I'm not sure how to fit this flag here. `StudentInfo` is returned by `GET students` endpoints, and basically, they accept a course against which we could check enrollment flag only like query parameter, and this "course" query parameter is not specified then the semantics of "isEnrolled" flag gets unclear.

pasqu4le: Isn't it possible to ignore `isEnrolled` when `course` is not specified?
Maybe also give it a default value so it is not necessary

Martoon-00: Which semantics would it have in couple with `course` query param?
If `isEnrolled = True` then only students enrolled to course `course` are returned, and if `isEnrolled = False` then such students are excluded?

pasqu4le: oh ok, now I get it. It doesn't make sense if the only filtering is on course enrollment anyway.

One question, because I don't remember if this is the case: is a student that completes a course still considered "enrolled" in it or not?
Because if it isn't then the this filter could have a use after all.

Martoon-00: That makes a point. Currently, a student is unrolled even if he completes the course, although I'm not sure whether does it make sense :)
When we polish the business logic, I will keep this filtering use case in mind.

educator/tests/Test/Dscp/Educator/Web/Educator/Queries.hs
```diff
@@ -204,34 +206,6 @@ spec_Educator_API_queries = specWithTempPostgresServer $ do
...
-
-            let submissions' =
-                  map (\(s, _) -> educatorLiftSubmission s Nothing) $
-                  applyFilterOn snd courseIdF $
-                  applyFilterOn (_sAssignmentHash . _ssSubmission . fst) assignHF $
-                  applyFilterOn (_sDocumentType . _ssSubmission . fst) docTypeF $
-                  sigSubs `zip` cycle courses
-
-            return $ sortOn siHash submissions === sortOn siHash submissions'
-
```
flyingleafe: What's wrong with this test?

flyingleafe (commented):
This seems really nice (though really unnecessary at this point of time, but hey)

educator/tests/Test/Dscp/Educator/Web/Educator/Queries.hs
Martoon-00: I think at the current level of automatization there is no need to test filtering as soon as many human mistakes are now impossible.
There is one test which covers filtering on `DocType`, and if there will be other complex filters I will write tests on them as well.

For me, it looks like we already have too many tests which are costly to maintain (each endpoint change requires fixing multiple test cases), so I'd like to avoid an unnecessary test code.

PR #392, opened 2019-02-20
[DSCP-480] Prepare multi-educator to deployment

flyingleafe:
### Description

Nothing to see here, just adding an executable to a distribution.

Actually, no, there seems to be a bug in `loadEducator` also!

### YT issue

https://issues.serokell.io/issue/DSCP-480

### Checklist

Stylistic (obligatory):

### DIFF
--- a/multi-educator/src/Dscp/MultiEducator/Launcher/Mode.hs
+++ b/multi-educator/src/Dscp/MultiEducator/Launcher/Mode.hs
@@ -28,7 +28,7 @@ import Loot.Config (option, sub)
 import Loot.Log (LoggingIO)
 import qualified Pdf.FromLatex as Pdf
 import System.Directory (createDirectoryIfMissing)
-import System.FilePath.Posix ((</>))
+import System.FilePath ((<.>), (</>))
 
 import Dscp.Config
 import Dscp.Crypto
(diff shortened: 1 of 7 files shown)


multi-educator/src/Dscp/MultiEducator/Launcher/Mode.hs
```diff
@@ -128,15 +128,15 @@ loadEducator login mpassphrase = do
...
         db = resources ^. lensOf @SQL
     -- set the DB schema name and create it if it's not ready
     setSchemaName db ("educator_" <> login)
     prepareEducatorSchema db
-    liftIO $ createDirectoryIfMissing True (appDir </> keyFile)
+    liftIO $ createDirectoryIfMissing True (appDir </> keyDir)
     -- read key from file and creates one if it does not exist yet
     let keyParams = finaliseDeferredUnsafe $ mempty
-            & option #path       ?~ Just keyFile
+            & option #path       ?~ Just (keyDir </> toString login <.> "key")
```
Martoon-00: :+1:

pasqu4le: As it is now there is the problem that a few lines above `createDirectoryIfMissing` is called on `appDir </> keyDir`, which is not the one used here in `keyParams`, and so that with the current `scripts/launch/node.sh` configuration the directory created would be `./run/run/educator.key` and the keyfile path that `linkStore` will try to use ends up being `run/educator.key/{educatorId}.key`.

If we want these keys to be in a `tmp` directory, I would suggest to:
- change in `scripts/launch/node.sh` the option `--educator-key-dir` to be just `$tmp_files`
- change the line above this one to use `createDirectoryIfMissing` on `keyDir`
- leave this line like this
This way the directory to create if necessary (`$tmp_files` => `run/tmp/multi-educator`) will contain the keyfiles with the path `run/tmp/multi-educator/{educatorId}.key`. 

If instead you want these to be persistent I would suggest to:
- change in `scripts/launch/node.sh` the option `--educator-key-dir` to be `$files/multieducator`
- change the line above this one to use `createDirectoryIfMissing` on `keyDir`
- leave this line like this
- add `run/multieducator/` in `.gitignore`
This way the keyfiles will be created as `./run/multieducator/{educatorId}.key` and the directory containing them (`./run/multieducator`) will be created if necessary.

pasqu4le (changes requested):
@flyingleafe you are right, there is a bug here, but your fix needs some adjustments, I left a comment explaining what is wrong and 2 options to deal with it (I tried both and they work).

Please keep in mind that I have not reviewed the 2 previous commits

pasqu4le:
Took a look at the 2 previous commits too, that I thought were bigger, and I have nothing to say on those.

flyingleafe:
@pasqu4le thanks a lot, you were right! I went with second option you suggested (we really need to preserve those keys)

pasqu4le (approved):

PR #393, opened 2019-02-20
[DSCP-474] Optional multi auth

Martoon-00:
### Description

Allows optional authentication AND enables logging (sorre, otherwise that would take too long). Only 2 commits.

I made multi-educator authentication return `EducatorAuthData` instead of `EducatorAuthToken` because the latter mainly contains (aside from `EducatorAuthData` itself) auth implementation details unneeded for endpoints implementation.

### YT issue

https://issues.serokell.io/issue/DSCP-474

### Checklist

Stylistic (obligatory):

### DIFF
--- a/multi-educator/src/Dscp/MultiEducator/Launcher/Mode.hs
+++ b/multi-educator/src/Dscp/MultiEducator/Launcher/Mode.hs
@@ -157,7 +157,8 @@ loadEducator login mpassphrase = do
             & sub #witness .~ multiEducatorConfig ^. sub #witness
             & sub #educator . sub #db .~ multiEducatorConfig ^. sub #educator . sub #db
             & sub #educator . sub #keys . sub #keyParams .~ keyParams
-            & sub #educator . sub #api .~ multiEducatorConfig ^. sub #educator . sub #api
+            & sub #educator . sub #api . option #educatorAPINoAuth .~
+                error "Do not touch, multi-educator has already run the server"
             & sub #educator . sub #publishing .~ multiEducatorConfig ^. sub #educator . sub #publishing
         educatorCtxWithCfg = E.withEducatorConfig newCfg $ EducatorCtxWithCfg educatorContext
     -- add to/update the educator context map with the newly made one
--- a/multi-educator/src/Dscp/MultiEducator/Web/Server.hs
+++ b/multi-educator/src/Dscp/MultiEducator/Web/Server.hs
@@ -31,8 +31,8 @@ import Dscp.MultiEducator.Config
 import Dscp.MultiEducator.Launcher.Mode (MultiCombinedWorkMode, MultiEducatorWorkMode,
                                          lookupEducator, normalToMulti)
 import Dscp.MultiEducator.Launcher.Params (MultiEducatorAAAConfigRec)
-import Dscp.MultiEducator.Web.Educator (MultiEducatorAPI, MultiStudentAPI,
-                                        multiEducatorAPI, multiStudentAPI)
+import Dscp.MultiEducator.Web.Educator (MultiEducatorAPI, MultiStudentAPI, multiEducatorAPI,
+                                        multiStudentAPI)
 import Dscp.MultiEducator.Web.Educator.Auth
 import Dscp.Web (serveWeb)
 import Dscp.Web.Metrics (responseTimeMetric)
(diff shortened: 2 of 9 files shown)


multi-educator/src/Dscp/MultiEducator/Launcher/Mode.hs
```diff
@@ -157,7 +157,6 @@ loadEducator login mpassphrase = do
             & sub #witness .~ multiEducatorConfig ^. sub #witness
             & sub #educator . sub #db .~ multiEducatorConfig ^. sub #educator . sub #db
             & sub #educator . sub #keys . sub #keyParams .~ keyParams
-            & sub #educator . sub #api .~ multiEducatorConfig ^. sub #educator . sub #api
```
pasqu4le: Even tho `newCfg` seems to be not really used (before DSCP-445 the value was `error ""`) the implementation I left actually works and produces a correct (finalized) Educator config because it substitutes **every** `sub`config with one from the (finalized) MultiEducator config we have.

Given that it starts from a `finaliseDeferredUnsafe mempty` if a `sub` substitution gets removed we'll not get a finalized Educator config anymore as our result.

One way to still keep this working would be to define something like:
```
meWebConfig = multiEducatorConfig ^. sub #educator . sub #api
toEducatorNoAuth = \case
    NoAuthOnContext _ -> NoAuthOnContext ()
    NoAuthOffContext -> NoAuthOffContext
```
and replace this removed line with:
```
& sub #educator . sub #api . sub #serverParams .~ meWebConfig ^. sub #serverParams
& sub #educator . sub #api . sub #botConfig .~ meWebConfig ^. sub #botConfig
& sub #educator . sub #api . option #educatorAPINoAuth .~ toEducatorNoAuth (meWebConfig ^. option #multiEducatorAPINoAuth)
& sub #educator . sub #api . option #studentAPINoAuth .~ meWebConfig ^. option #studentAPINoAuth
```
or something like this.

As I said I'm not really sure how important this is and maybe I should have put a comment about the fact that this is basically the creation of a finalized config from another finalized one.

Martoon-00: Well, that makes sense.
On the other hand, we know that this parameter cannot and shouldn't be used by educator because multi-educator code handles bringing educator API server up. Using `toEducatorNoAuth` would look unpleasant IMO because `NoAuthContext "educator"` and `NoAuthContext "multi-educator"` have pretty different semantics, which may express in us getting stuck if `NoAuthOnContext` for `"educator"` one day contains something smarter than `()`.

I guess I will set it as
`sub #educator . sub #api . option #educatorAPINoAuth .~ error "Do not touch, multi-educator has already run the server"`. So that `finaliseDeferredUnsafe mempty` is again justified (we can replace it with `finalise` with the same effect), but we know that `"educatorAPINoAuth"` config parameter shouldn't be used by the educator.

pasqu4le: Yeah, that's better!

multi-educator/src/Dscp/MultiEducator/Web/Server.hs
```diff
@@ -68,45 +66,25 @@ mkMultiEducatorApiServer _aaaConfig nat =
...
 createStudentCheckAction
     :: forall ctx m. MultiEducatorWorkMode ctx m
-    => EducatorBotConfigRec
-    -> m StudentCheckAction
-createStudentCheckAction _ =
+    => m StudentCheckAction
+createStudentCheckAction =
     return . StudentCheckAction . const $ pure True
 {-
 createStudentCheckAction EducatorBotOff = do
```
pasqu4le: You missed a piece of bot comments here

Martoon-00: Hm, they are not fully bot-related, probably this commented check should happen in real life but I do not enable it because it would make testing more difficult :smirk_cat: 
@flyingleafe what do you think?

Anyway, I will remove that `EducatorBotOff` argument, thanks for the notice.

Martoon-00:
I suppose no crucial points remained here, could I get an approve?
If anything is wrong in the related code, we can fix that later.

pasqu4le:
Sorry, I forgot, done now

Martoon-00:
:tada:

PR #403, opened 2019-02-26
[DSCP-496] Workers in multi-educator bug

Martoon-00:
### Description

A pack of crucial bugfixes.

**Review on per-commit basis is recommended.**

### YT issue

https://issues.serokell.io/issue/DSCP-496

### Checklist

Stylistic (obligatory):

### DIFF
--- a/multi-educator/src/Dscp/MultiEducator/Launcher/Resource.hs
+++ b/multi-educator/src/Dscp/MultiEducator/Launcher/Resource.hs
@@ -3,7 +3,8 @@
 -- | Resources used by Educator node
 
 module Dscp.MultiEducator.Launcher.Resource
-       ( EducatorContexts (..)
+       ( EducatorContexts
+       , EducatorContextsVar (..)
        , LoadedEducatorContext (..)
        , MaybeLoadedEducatorContext (..)
        , MultiEducatorResources (..)
@@ -11,17 +12,20 @@ module Dscp.MultiEducator.Launcher.Resource
        , merEducatorData
        ) where
 
-import Control.Lens (makeLenses, Wrapped (..))
+import Control.Lens (makeLenses)
+import Loot.Log (logWarning)
 import qualified Pdf.FromLatex as Pdf
-import UnliftIO.Async (Async)
+import Time (sec)
+import UnliftIO.Async (Async, forConcurrently_, cancel)
 
 import Dscp.Config
 import Dscp.DB.SQL (SQL)
 import qualified Dscp.Educator.Config as E
 import qualified Dscp.Educator.Launcher.Mode as E
 import Dscp.MultiEducator.Config
-import Dscp.Resource.Class (AllocResource (..))
+import Dscp.Resource.Class (AllocResource (..), buildComponentR)
 import Dscp.Resource.Network (NetServResources)
+import Dscp.Util.TimeLimit
 import Dscp.Util.HasLens
 import qualified Dscp.Witness.Launcher.Resource as Witness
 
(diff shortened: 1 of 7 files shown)


scripts/launch/node.sh
```diff
@@ -112,7 +112,7 @@ educator_params="
 --student-api-no-auth 3BAyX5pNpoFrLJcP5bZ2kXihBfmBVLprSyP1RhcPPddm6Dw42jzEPXZz22
 --publication-period 15s
 --pdf-resource-path ../pdfs/template
---cert-issuer-name "Grimpy Cat University"
+--cert-issuer-name \"University\"
 --cert-issuer-url example@gmail.com
```
pasqu4le: I forgot to comment on this, but AFAIK it is good practice to use `*@example.com` for these cases.
Also I took both this values from `certificateIssuerInfoEx` (in `Dscp.Core.Arbitrary`) so you may want to change those too.

stack.yaml
```diff
@@ -77,14 +76,20 @@ extra-deps:
 - git: https://github.com/serokell/temporary.git
   commit: 102ed4671dd78c5d271f09756681aac2a41b6717
 
+# Fixes missing resources cleanup in case of async exception.
```
pasqu4le: Very nice!

scripts/launch/node.sh
Martoon-00: Oh ok, didn't know about that practice. Suggest some e-mail/website then :)

I had to set `University` here because I don't know how to escape a quoted expression containing spaces in bash (i.e. `... \"Grimpy Cat University\"` failed to interpret).

pasqu4le: I don't know, `university@example.com`?
You are right about the escaping, I totally forgot about that.
On the same matter, should't `--sql-conn-str "postgresql:///$sql_db_name"`, a few lines above, be escaped as well?

Martoon-00: Yeah, I like it.

Nice point, I somehow forgot about escaping there even despite emacs highlights it differently. Truly surprised it worked :smile:

pasqu4le: :man_shrugging:

multi-educator/src/Dscp/MultiEducator/Launcher/Resource.hs
```diff
@@ -62,6 +65,26 @@ deriveHasLensDirect ''MultiEducatorResources
...
+    type Deps EducatorContextsVar = ()
+    allocResource () =
+        fmap EducatorContextsVar $
+        buildComponentR "Educator contexts var"
+            (newTVarIO mempty)
+            shutdownAllContexts
+      where
+        shutdownAllContexts ctxsVar = do
+            ctxs <- atomically $ readTVar ctxsVar
+            -- TODO [DSCP-494]: prevent further insertions to the context var
```
flyingleafe: I wonder if doing this in one transaction would suffice.

scripts/launch/node.sh
```diff
@@ -106,18 +106,18 @@ fi
...
 --pdf-resource-path ../pdfs/template
---cert-issuer-name "Grimpy Cat University"
---cert-issuer-url example@gmail.com
+--cert-issuer-name \"University\"
+--cert-issuer-url university@example.com
 "
 multi_educator_params="
 --educator-key-dir $files/multieducator
---sql-conn-str "postgresql:///$sql_db_name"
+--sql-conn-str postgresql:///$sql_db_name
```
flyingleafe: did this really matter?...

stack.yaml
flyingleafe: Yeah, do the PR - this is also a way of knowing whether or not there were some deep idea behind not handling async exceptions (seems really unlikely, but anyway).

multi-educator/src/Dscp/MultiEducator/Launcher/Resource.hs
Martoon-00: What do you mean?

scripts/launch/node.sh
Martoon-00: Eventually no ofc, but I do not understand how previous construction worked (despite it's mine) :magic:
Since @pasqu4le's noticed, why not to fix.

multi-educator/src/Dscp/MultiEducator/Launcher/Resource.hs
flyingleafe: I mean, reading TVar and shutting down contexts in one STM transaction.

Martoon-00: Ah right, I'm already doing so in DSCP-494 (yet unpushed), so it will be alright eventually.

PR #404, opened 2019-02-26
[DSCP-502] TODOs hunting

pasqu4le:
### Description

This PR aims at solving some TODOs in the code that are not part of any issue.
It could also serve to start new issues, perhaps.

This is not a WIP anymore and the only commit that leaves me unhappy is the one about the product list class helper for `Serialise`.

Single TODOs are handled in single commits, in this order:
- proper Show instance for Address
  was easily solved because the module problem was not there anymore
- well-specified Serialize instances
  I made a specific instance for every declaration in`Dscp.Core.Foundation.Instances`, in particular following the rules that:
  - newtypes & data with 1 constructor got a basic conversion implementation
  - product types are encoded `List`s. Where possible all record fields name are specified so to have better errors in case the types are modified in the future
  - sum types are encoded `Word`s with a simple mapping
- `Buildable` instance for `MerkleProof`
  has required more `Buildable` instances and constraint, but they seem acceptable to me
- `BaseUrl` Json instances
  here I ended up just removing the comment because it is not true anymore that this is the only reason why `disciplina-core` needs `servant-client-core` (`Dscp.Util.Servant` needs it too). On top of this also `disciplina-multi-educator` need the JSON instance too and I don't think there is a better place then this.
- move generic auth stuff from Educator.Web.Auth
  I ended up moving a lot of it, because it seems to me that is all generic enough, but this ended up adding more dependencies to `disciplina-core`, so I don't know how to feel about it
- let `authTimeout` be defined in config
  this took quite some more changes than originally expected, but seems to have turned out well
- class to Serialise product types
  this has some problem (see Slack discussion) and I'm not satisfied with the result, but couldn't do any better
- helper function to Serialise Enums
  these seem simple and useful, in a comment it is explained why this is not simply an `instance` definition: it cannot coexist with the product type one of the previous commit
- about swagger docs
  this has been recognized as obsolete and so it has been removed
- about an implementation for sqlite
  this was not really solved but replaced with a `NOTE` that explains it better, because the module itself is not used anymore
- add reference to DSCP-505
  there was a pair of TODOs about removing the transaction list from `AccountInfo` and the `GET account` endpoint, a new issue has been made because this need some changes in the frontend as well. Also, more TODOs added
- `simpleBlkConfiguration` check for header's slot time
  following how `SlotId`s are made I build a simple check for what the TODO asked

### YT issue

https://issues.serokell.io/issue/DSCP-502

### Checklist

Stylistic (obligatory):
- [x] My code is documented (haddock) and these docs are decent (full enough and sp ellchecked).


core/src/Dscp/Core/Foundation/Instances.hs
```diff
@@ -18,51 +18,263 @@ import Dscp.Util
...
+            1 -> return Offline
+            _ -> fail "Unexpected tag"
 
 instance Serialise Day where
     encode = encode . toModifiedJulianDay
     decode = ModifiedJulianDay <$> decode
 
-instance Serialise EducationForm
-instance Serialise CertificateMeta
+-- consider that it is an Enum
```
pasqu4le: Just remembered that I left this comment behind, I'll remove it in the next commit

core/src/Dscp/Core/Foundation/Instances.hs
```diff
@@ -103,23 +313,210 @@ instance Serialise Timestamp where
...
+        ]
+    decode = do
+        decodeListLenOf 2
+        bHeader <- decode
+        bBody   <- decode
+        return $ Block {..}
+
+instance Serialise BlockBody where
+    encode = encode . bbTxs
+    decode = BlockBody <$> decode
```
Martoon-00: Thanks, that's a lot of work!

However, I personally think that we could go in a less boilerplate-y way. Except for long instances definition, now we have to write roundtrip tests.

Don't you mind thinking about how could we reduce manual work on defining these instances?
Out of the box idea: use `HList` or `Rec` as an intermediate representation for CBOR encoded list. `HList <-> Encoding` conversion can be applied automatically, while `type <-> HList` is still to be defined manually. Profit: you cannot mistake in specifying list length and it's pretty difficult to mistaken in arguments order. As a bonus, `vinyl` suggests a [function](http://hackage.haskell.org/package/vinyl-0.11.0/docs/Data-Vinyl-Curry.html#v:rcurry) which allows filling a constructor with values from `Rec`. I'm not sure that this approach lacks its downsides though.

There is a slight problem here, that no one yet understands what do we want from these instances and where flexibility is required :sweat_smile:. I hope you will find a solution )

pasqu4le: You are right, I started trying to see what the difficulties were and ended up making instances in one of the 3 ways described above in the description.
As you said there are a couple of problems with this approach (list length is very easy to mess up, sum-type "word" conversion is even more fragile), but also has a couple advantages (the decode side made with all the record fields spelled out will force one to update the definition when necessary).

I'll give a look at your suggestions, it may be possible to at least remove some of the boilerplate, this can only be a benefit, and pay attention at how easy can we modify them later.

Martoon-00 (commented):
Left a couple of minor comments about serialization :)

Martoon-00 (commented):
And a pack of more ones)
Looks good to me generally. Could you remind what unpleasant do you find in current serialization? (I lost track of related discussion in slack :/)

Martoon-00 (approved):
Btw, LGTM

flyingleafe (approved):
Looks great!

PR #405, opened 2019-02-27
[DSCP-500] Add QR-codes to certificates

flyingleafe:
### Description

The PR inserts a public link to PDF via QR code to certificate.
Per-commit review is recommended.

Tested on localhost and on development cluster.

### YT issue

https://issues.serokell.io/issue/DSCP-500

### Checklist

Stylistic (obligatory):


educator/src/Dscp/Educator/Config.hs
```diff
@@ -59,6 +63,11 @@ defaultEducatorConfig = upcast defaultWitnessConfig
     & sub #educator . sub #keys . sub #keyParams .~ defaultBaseKeyParams
     & sub #educator . sub #api . sub #botConfig . tree #params . selection ?~ "disabled"
     & sub #educator . sub #certificates . option #latex ?~ "xelatex"
+    & sub #educator . sub #certificates . option #downloadBaseUrl ?~ defBaseUrl
+    & sub #educator . sub #certificates . sub #issuer . option #id ?~ "Principal"
+  where
+    defBaseUrl = nothingToPanic "url is correct" $
+                 parseBaseUrl "https://localhost/api/certificates/v1/cert"
```
pasqu4le: Shouldn't it be `nothingToPanic "url is not correct"` or is this because you know it is correct already?
BTW this could be avoided entirely by building a `BaseUrl` instead of parsing one.

educator/tests/Test/Dscp/Educator/Mode.hs
```diff
@@ -86,6 +87,11 @@ testResourcePath = unsafePerformIO $ do
         Just path -> pure path
     return (Pdf.ResourcePath path)
 
+testDownloadBaseUrl :: Pdf.DownloadBaseUrl
+testDownloadBaseUrl = Pdf.DownloadBaseUrl $
+    nothingToPanic "url is correct" $
+    parseBaseUrl "https://educator.disciplina.io/api/certificates/v1/cert"
```
pasqu4le: same comment here as in Educator config

pasqu4le (approved):
Nothing bad to say and it seems to be working fine, just left a comment on an odd thing.

educator/src/Dscp/Educator/Config.hs
flyingleafe: Yeah, it means that I know that it is correct.
You're right about building the URL, but having it like this in the code (in a string literal) is easier to read, I guess.

pasqu4le: Good then, I figured that could be the motivation

